PR #10, opened 2017-07-07
Add a minimal hello_web example

Shnatsel:
Add a convenient starting point for a Rust web project with very basic crate usage.

Something like this might be useful in `cargo web`

`console.log` was chosen over `alert` because it also works in node.js; may or may not be a good idea.

Should I also add an HTML that calls the .js file? If so, how do I call the JS? The todomvc example looks for the code in js/app.js and that's not where `cargo build --target=asmjs-unknown-emscripten` puts it. I imagine there is some magic that `cargo web` does in this case, but I'd strongly favor a "batteries included" kind of hello world :smile_cat:

### DIFF
--- a/examples/hello_web/Cargo.toml
+++ b/examples/hello_web/Cargo.toml
@@ -1,0 +1,7 @@
+[package]
+name = "hello_web"
+version = "0.1.0"
+authors = ["Sergey \"Shnatsel\" Davidoff <shnatsel@gmail.com>"]
+
+[dependencies]
+stdweb = "*"
--- a/examples/hello_web/src/main.rs
+++ b/examples/hello_web/src/main.rs
@@ -1,0 +1,8 @@
+#[macro_use]
+extern crate stdweb;
+
+fn main() {
+    stdweb::initialize();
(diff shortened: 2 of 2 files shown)


koute:
Sounds like a good idea to me!

Although I'd make two changes:
1. Add a semicolon at the end of `console.log`. (: (It works without one, but semicolon-less JavaScript will not be supported by the `js!` macro in the future.)
2. Change the name. As much as I like the `hello_web` name I think `hello_world` would be more appropriate, considering this would be an example for absolute beginners and basically every programmer knows what a "hello world" is.

> Should I also add an HTML that calls the .js file?

What do you mean by "calls"? Only includes? Or explicitly calls a function which is exported by the Rust code?

> I imagine there is some magic that cargo web does in this case

Yep. That's exactly what happens. It serves the crate as `js/app.js`, and if it doesn't find an index.html it will dynamically generate one and serve it (and the index.html which it generates includes `js/app.js`, which is why everything magically works by default).

> but I'd strongly favor a "batteries included" kind of hello world 😸

Do you mean a hello world without the use of cargo-web?

PR #17, opened 2017-08-28
Implement methods and properties for events

Herschel:
I'd be happy to help fill out some of the DOM APIs. Here's my first attempt at filling out the properties for the various Events that have been defined.

Some thoughts/questions:
1) I've mostly avoided adding any API marked as "obsolete/deprecated/non-standard."
2) What would be a good way to test these properties? Just a simple test that creates the object and gets+sets the properties?
3) Most properties follow the simple forwarding pattern of `js!( return @{self}.foo; ).try_into().unwrap()`. Would it be worthwhile to make a macro for getters/setters to reduce this boilerplate?


src/webapi/event.rs
```diff
@@ -10,26 +12,169 @@ use webcore::try_from::{TryFrom, TryInto};
...
+            @{self.as_ref()}.stopPropagation();
+        }
+    }
+
+
+    /// Returns a reference to the target to which this event was originally registered.
+    ///
+    /// [(JavaScript docs)](https://developer.mozilla.org/en-US/docs/Web/API/Event/target)
+    #[inline]
+    fn target( &self ) -> Option< Reference > {
```
koute: Hmm... it probably would make more sense to return an `EventTarget` instead of a `Reference` here. (Since you can only register events on `EventTarget`s.)

Looking at the [spec](https://dom.spec.whatwg.org/#concept-event) I can find the following:

```
An event has [...] target (null or an EventTarget object),
```

so if the spec is to be believed it looks like it indeed should be an `EventTarget`.

src/webapi/event.rs
```diff
@@ -121,6 +315,212 @@ reference_boilerplate! {
...
+            return @{self.as_ref()}.altKey;
+        ).try_into().unwrap()
+    }
+
+    /// Indicates the mouse button that fired this event.
+    ///
+    /// [(JavaScript docs)](https://developer.mozilla.org/en-US/docs/Web/API/MouseEvent/button)
+    fn button( &self ) -> MouseButton {
+        match js!(
+            return @{self.as_ref()},.button;
```
koute: Accidental `,` before the `.button`. (:

src/webapi/event.rs
```diff
@@ -121,6 +315,212 @@ reference_boilerplate! {
...
+    ///
+    /// [(JavaScript docs)](https://developer.mozilla.org/en-US/docs/Web/API/MouseEvent/clientX)
+    #[inline]
+    fn client_x( &self ) -> f64 {
+        js!(
+            return @{self.as_ref()}.clientX;
+        ).try_into().unwrap()
+    }
+
+    /// Returns the X position in the application's client area where this event occured.
```
koute: `X position` -> `Y position`

src/webapi/event.rs
```diff
@@ -453,3 +1085,55 @@ reference_boilerplate! {
...
+        return js!(
+            @{self.as_ref()}.filename;
+        ).try_into().unwrap()
+    }
+
+    /// Returns the line number of the script file where the error occurred.
+    ///
+    /// [(JavaScript docs)](https://developer.mozilla.org/en-US/docs/Web/API/ErrorEvent/lineNo)
+    #[inline]
+    pub fn line_no( &self ) -> i32 {
```
koute: As far as I can see [here](https://developer.mozilla.org/en-US/docs/Web/API/ErrorEvent) the property is named `lineno`, so to keep with the convention there shouldn't be an underscore in this function's name.

Ditto for `colno`.

src/webapi/event.rs
```diff
@@ -453,3 +1085,55 @@ reference_boilerplate! {
...
+        return js!(
+            @{self.as_ref()}.colno;
+        ).try_into().unwrap()
+    }
+
+    /// Returns the object that is concerned with the event.
+    ///
+    /// [(JavaScript docs)](https://developer.mozilla.org/en-US/docs/Web/API/ErrorEvent/error)
+    #[inline]
+    pub fn error( &self ) -> Reference {
```
koute: MDN tells me this is an experimental API.

I'm not against including experimental APIs, but I'd be nice to at least document that in the docs. (e.g. put `(experimental API)` or something like that on the first line so that the user immediately knows it's experimental)

src/webapi/event.rs
```diff
@@ -121,6 +315,212 @@ reference_boilerplate! {
...
+            3 => MouseButton::Button4,
+            4 => MouseButton::Button5,
+            _ => unreachable!("Unexpected MouseEvent.button value"),
+        }
+    }
+
+    /// Indicates which mouse buttons were down when this event was fired.
+    ///
+    /// [(JavaScript docs)](https://developer.mozilla.org/en-US/docs/Web/API/MouseEvent/buttons)
+    fn buttons( &self ) -> Vec<MouseButton> {
```
koute: Hmm... I'm not sure how I feel about making this return an vector. The original JS API returns a bitmask, and it would be nice to transfer this to Rust, but in a type safe matter.

How about creating a `MouseButtons` type which we would return from this function instead of a raw vector? Then we wouldn't force ourselves to always return a vector here, and we could still expose this functionality as `MouseButtons::to_vec` while we could postpone having to figure out how to best deal with those flags in an idiomatic manner.

koute:
Apart from a few minor issues this looks great, thanks a lot!

> I've mostly avoided adding any API marked as "obsolete/deprecated/non-standard."

Yep. I don't think such APIs should be added; if the user *really* wants to use a non-standard API he/she can do it manually with the `js!` macro.

> What would be a good way to test these properties? Just a simple test that creates the object and gets+sets the properties?

That's a very good (and still pretty much open) question! There are a few problems/issues concerning running tests in this crate:

 * Currently my Travis job runs the tests with Node.js, so it will blow up on any tests that use Web-specific APIs. (Basically anything which isn't implemented by Node.) So for now any tests for Web APIs would have to be temporarily annotated with something like `#[cfg(web_api_tests)]` to disable them by default.
 * What we really need is a way to run the tests in a real web browser.
 * `cargo-web` has *some* support for running tests under Chrome/Chromium, but it's *very* hacky, and requires a non-headless Chrome/Chromium. One could, however, run them locally for now, and as long as they would be `#[cfg]`'d away they wouldn't be a problem on Travis.
 * What we *really* need is to rework `cargo-web` to use headless Chrome/Chromium for running tests, and use a different (better) approach to communicate with it. (Currently `cargo-web` injects some JS code and uses AJAX to talk with itself; ideally we want to replace it with something like Chrome's debugger protocol.)
 * Currently it's not possible *at all* to test stuff like `setTimeout`, or anything else which is asynchronous for that matter. I haven't thought too deeply how we could fix that yet; it would probably require support from `cargo` and Rust to handle asynchronicity in tests.

> Most properties follow the simple forwarding pattern of js!( return @{self}.foo; ).try_into().unwrap(). Would it be worthwhile to make a macro for getters/setters to reduce this boilerplate?

Hmm... I don't see why not!

Herschel:
Thanks for the thorough + helpful review! I tried to fix the issues mentioned.

- Added tests. They are hidden behind a web_api_tests cfg.
- Fixed some bugs that the tests caught.
- Switched `MouseEvent::buttons()` to return an opaque type instead of a `Vec`.
- Decided to remove experimental web APIs for now.

Some more possible todo:
- Implement the `Event` constructors. Not sure how the design will look exactly for this (Builder pattern for each event subtype?)
- Now that associated constants are stable, use them for `static_event_type`.

koute:
Looks good to me; thanks!

> Now that associated constants are stable, use them for `static_event_type`

Yep!

>  Implement the `Event` constructors. Not sure how the design will look exactly for this (Builder pattern for each event subtype?)

Hmm... yeah, since we definitely want to do it in a forward-compatible way something like that would probably be the way to go, e.g. `fn new() -> KeyboardEventBuilder< KeypressEvent >` defined on `KeypressEvent`. (Having a generic builder would allow us to reuse it for similar events, like `keypress` and `keyrelease`.)

PR #18, opened 2017-09-17
Implement ImageElement

Herschel:
Implement `ImageElement`.
Also add `CrossOriginSetting` enum for the `crossOrigin` property, used by various HTML elements including `ImageElement`.

Done as part of the the San Diego Rust September Libz Blitzzz!
https://www.meetup.com/San-Diego-Rust/events/243150834/
With help from @chrishiestand and @jkleint!

### DIFF
--- a/src/webapi/html_elements/image.rs
+++ b/src/webapi/html_elements/image.rs
@@ -1,0 +1,293 @@
+use webcore::value::{Value, Reference};
+use webcore::try_from::TryInto;
+use webapi::cross_origin_setting::CrossOriginSetting;
+use webapi::event_target::{IEventTarget, EventTarget};
+use webapi::node::{INode, Node};
+use webapi::element::{IElement, Element};
+use webapi::html_element::{IHtmlElement, HtmlElement};
+
+/// The HTML image element is used to manipulate the layout and presentation of
+/// `<img>` elements.
+///
+/// [(JavaScript docs)](https://developer.mozilla.org/en-US/docs/Web/API/HTMLImageElement)
+pub struct ImageElement( Reference );
+
+impl IEventTarget for ImageElement {}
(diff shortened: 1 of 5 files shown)


koute:
This looks great! Thanks a lot!

src/webapi/html_elements/image.rs
```diff
@@ -0,0 +1,293 @@
...
+        js! { @(no_return)
+            @{self}.isMap = @{value};
+        }
+    }
+
+    /// Returns the intrinsic height of the image in CSS pixels, if it is available;
+    /// else, it returns 0.
+    ///
+    /// [(JavaScript docs)](https://developer.mozilla.org/en-US/docs/Web/API/HTMLImageElement/naturalHeight)
+    pub fn natural_height( &self ) -> u32 {
```
koute: It would probably be better to return an `Option< u32 >` here, and in case of a zero return `None`. (As far I see from the documentation `0` means the lack of a value, which screams `Option` to me)

src/webapi/html_elements/image.rs
```diff
@@ -0,0 +1,293 @@
...
+        js!(
+            return @{self}.naturalHeight;
+        ).try_into().unwrap()
+    }
+
+    /// Returns the intrinsic width of the image in CSS pixels, if it is available;
+    /// else, it returns 0.
+    ///
+    /// [(JavaScript docs)](https://developer.mozilla.org/en-US/docs/Web/API/HTMLImageElement/naturalWidth)
+    pub fn natural_width( &self ) -> u32 {
```
koute: Ditto here.

koute:
Would be nice to fix the `natural_height` and `natural_width` to return an `Option`, but considering that's a minor change it can go in a separate PR and I'll merge this as is.

PR #19, opened 2017-09-21
Fix unused_mut warning

Herschel:
Squelches some "variable does not need to be mutable" warnings that would occur in serialization.rs on the nightly compiler.


koute:
Thanks!

PR #22, opened 2017-10-06
ConcreteEvent: use associated constants for `event_type`s

chpio:
requires rust [v1.20](https://github.com/rust-lang/rust/blob/master/RELEASES.md#version-1200-2017-08-31)


koute:
Thanks!

PR #23, opened 2017-10-08
Support passing symbols

Pajn:
Symbols can not be used as keys in a WeakMap, therfore ref_to_id_map.set throws when a symbol is passed from JS to Rust.
I don't know if this is the correct way to solve it but it seems to work with my limited testing.


koute:
Hmm... we might want to actually support symbols natively in `Value`; I'll have to think about it.

In the meantime, thanks! For now this looks good to me.

PR #24, opened 2017-10-08
Export Value in private module

Pajn:
I need to create pass around a custom JS type without serializing it.
I guess there is a good reason that you have not exported the Value type so I thought exporting it from the private module could be an ok solution for now.


Pajn:
Oops, I'm blind. Value is already exported.

PR #27, opened 2017-10-15
Add Node methods and tests

Herschel:
Add missing properties and method for Node interface.
Add tests for all methods.


koute:
This looks great; thanks!

PR #37, opened 2017-12-09
implement date bindings

bbyler:
First, a big thank you to the contributors/maintainers of this library: this is great! I'd like to find a way to contribute to stdweb and help build the foundation for Rust --> (W)ASM --> web

Opening a PR for date bindings to help fill in the browser API, bit by bit. Tests are working, albeit they are mostly "happy path" tests. 

**Edit**: tests breaking in CI build...looking into it
**Update**: these are all tests that pass locally and work in the browser...that said, I've been running against WASM. Travis CI is using asmjs-unknown-emscripten...could this be the difference?


bbyler:
"Meta" comments: A couple things I can do if need be:

1. squash commit into one
2. create labels "WIP", "Needs Review", "Needs Tests"

src/webapi/date.rs
```diff
@@ -1,22 +1,903 @@
...
+    /// NOTE: for the following functions that take arguments in addition to self, the arguments
+    /// are allowed to be empty. From the user's perspective, the first argument is mandatory, with
+    /// the rest optional. However, browsers do not need all arguments, even the first, but calling
+    /// with empty arguments returns NaN.
+
+    ///
+    ///
+    /// [(JavaScript docs)](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/setDate)
+    #[inline]
+    fn set_date( &self, _day_value: f64 ) -> f64 {
```
koute: Remove the leading `_` from `_day_value`; also does taking and returning `f64` make sense here? (Is passing a fractional value here sensible? Will it ever return a fractional value back?)

src/webapi/date.rs
```diff
@@ -1,22 +1,903 @@
...
+        js!( return @{self.as_ref()}.setDate(@{_day_value}); ).try_into().unwrap()
+    }
+
+    ///
+    ///
+    /// [(JavaScript docs)](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/setFullYear)
+    #[inline]
+    fn set_full_year( &self, args: &[i32] ) -> f64 {
```
koute: It would be better to just take `year: i32, month: Option< i32 >, day: Option< i32 >` here.

src/webapi/date.rs
```diff
@@ -1,22 +1,903 @@
...
+            },
+            _ => js!( return @{self.as_ref()}.setFullYear(); ).try_into().unwrap()
+        }
+    }
+
+    ///
+    ///
+    /// [(JavaScript docs)](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/setHours)
+    #[inline]
+    fn set_hours( &self, args: &[i32] ) -> f64 {
```
koute: Same here; taking `Options` for the optional arguments would be better.

src/webapi/date.rs
```diff
@@ -1,22 +1,903 @@
...
+            },
+            _ => js!( return @{self.as_ref()}.setHours(); ).try_into().unwrap()
+        }
+    }
+
+    ///
+    ///
+    /// [(JavaScript docs)](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/setMilliseconds)
+    #[inline]
+    fn set_milliseconds( &self, _milliseconds_value: f64 ) -> f64 {
```
koute: Remove leading `_`; the same question regarding `f64` as with `set_date`

src/webapi/date.rs
```diff
@@ -1,22 +1,903 @@
...
+
+    /// DEPRECATED
+    /// fn set_year() -> f64 { }
+
+    ///
+    ///
+    /// [(JavaScript docs)](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/toDateString)
+    #[inline]
+    fn to_date_string( &self ) -> String {
+        js! ( return @{self.as_ref()}.toDateString; ).try_into().unwrap()
```
koute: Missing `()` to actually call the `toDateString`.

src/webapi/date.rs
```diff
@@ -1,22 +1,903 @@
...
+    }
+
+    /// DEPRECATED
+    /// fn to_gmt_string( &self ) -> f64 { }
```
koute: You can get rid of this comment. (:

src/webapi/date.rs
```diff
@@ -1,22 +1,903 @@
...
+    #[inline]
+    fn to_json( &self ) -> String {
+        js! ( return @{self.as_ref()}.toJSON(); ).try_into().unwrap()
+    }
+
+    /// to_locale_date_string
+    ///
+    /// [(JavaScript docs)](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/toLocaleDateString)
+    #[inline]
+    fn to_locale_date_string( &self, _locales: Option< &str >, _options: Option< &HashMap< String, String > >) -> String {
```
koute: Uh, this function (on the JS side) looks super complex to reasonably bind in Rust. I definitely wouldn't want to have a stringly-typed prototype like this one.

My suggestion would be to just remove this one altogether for now. (If someone *really* wants to use it he/she can always use the `js!` macro.)

src/webapi/date.rs
```diff
@@ -1,22 +1,903 @@
...
+    }
+
+    /// NON-STANDARD
+    /// fn to_locale_format( &self ) -> f64 { }
+
+    /// to_locale_string
+    ///
+    /// [(JavaScript docs)](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/toLocaleString)
+    #[inline]
+    fn to_locale_string( &self, _locales: Option< &str >, _options: Option< &HashMap< String, String > >) -> String {
```
koute: Same as `to_locale_date_string`.

src/webapi/date.rs
```diff
@@ -1,22 +1,903 @@
...
+
+            None => return js! ( return @{self.as_ref()}.toLocaleString(); ).try_into().unwrap()
+        }
+    }
+
+    /// to_locale_time_string
+    ///
+    /// [(JavaScript docs)](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/toLocaleTimeString)
+    #[inline]
+    fn to_locale_time_string( &self, _locales: Option< &str >, _options: Option< &HashMap< String, String > > ) -> String {
```
koute: Same as `to_locale_date_string`.

src/webapi/date.rs
```diff
@@ -1,22 +1,903 @@
...
+                None => return js! (
+                    return @{self.as_ref()}.toLocaleTimeString(@{_locales});
+                ).try_into().unwrap()
+            },
+
+            None => return js! ( return @{self.as_ref()}.toLocaleTimeString(); ).try_into().unwrap()
+        }
+    }
+
+    /// NON-STANDARD
```
koute: You can remove this comment. (:

src/webapi/date.rs
```diff
@@ -1,22 +1,903 @@
...
+    /// fn to_source( &self ) -> f64 { }
+
+    ///
+    ///
+    /// [(JavaScript docs)](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/toString)
+    #[inline]
+    fn toString( &self ) -> String {
```
koute: It would probably be more Rusty to do an `impl From< &Date > for String` or something rather than this.

src/webapi/date.rs
```diff
@@ -1,22 +1,903 @@
...
+        js! ( return @{self.as_ref()}.toString(); ).try_into().unwrap()
+    }
+
+    ///
+    ///
+    /// [(JavaScript docs)](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/toTimeString)
+    #[inline]
+    fn to_time_string( &self ) -> String {
+        js! ( return @{self.as_ref()}.toTimeString; ).try_into().unwrap()
```
koute: Missing `()`

src/webapi/date.rs
```diff
@@ -1,22 +1,903 @@
...
+    }
+
+    ///
+    ///
+    /// [(JavaScript docs)](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/toUTCString)
+    #[inline]
+    fn to_utc_string( &self ) -> String {
+        js! ( return @{self.as_ref()}.toUTCString; ).try_into().unwrap()
```
koute: Missing `()`

koute:
Thanks a lot for the PR!

Besides the comments I've already posted a few other remarks:
1. I'd definitely like to see some tests. (: A lot of the APIs are currently untestable due to the lack of tooling, however basic stuff like this **is** testable.
2. Since you've asked for commits I *usually* like them in a way where separate changes are in separate commits (e.g. the whitespace change is unrelated to your other changes, so it would be nice to have it in a separate commit), although I don't intend to be a nazi about this. (: 
3. In the commit message you've wrote that the tests are not working for you due to a linker error - how are you trying to run them? `cargo test` will probably not work; if you use my cargo-web (e.g. `cargo-web test --nodejs`) it should work.

examples/todomvc/src/main.rs
```diff
@@ -236,6 +239,23 @@ fn main() {
         update_dom( &state );
     }));
 
+    let d: Date = js!( return new Date(); ).try_into().unwrap();
+    let todays_date: HtmlElement = document().query_selector( "header h2" ).unwrap().try_into().unwrap();
+
+    let options: HashMap< String, String > = [
```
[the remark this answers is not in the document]
bbyler: HashMap< &str, &str > works well too - making the change!

src/webapi/date.rs
```diff
@@ -1,22 +1,903 @@
...
+pub trait IDate: AsRef< Reference > + TryFrom< Value >  {
     /// The Date.now() method returns the number of milliseconds elapsed since 1 January 1970 00:00:00 UTC.
     ///
     /// [(JavaScript docs)](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/now)
-    pub fn now() -> f64 {
-        em_asm_double!( "return Date.now();" )
+    #[inline]
+    fn now() -> f64 {
+        // em_asm_double!( "return Date.now();" )
+        js!( return Date.now(); ).try_into().unwrap()
```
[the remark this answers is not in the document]
bbyler: I'm with you on this one. I hesitated on em_asm_double only out of concern from not fully understanding how emscripten is really working behind the scenes and thinking this was triggering my "symbol not found"-like errors when compiling. This was not at all the issue though (I figured out how to get tests working a few hours after putting this together - found the cargo web call in a YAML file somewhere. Huge help for development!!)

Coming in to this repo, I was curious about overhead when making calls to js! vs e.g. em_asm_double...the conclusion (for now?) seems to be that em_asm_double and friends are preferred, wherever possible

src/webapi/date.rs
```diff
@@ -1,22 +1,903 @@
...
+    }
+
+    /// Date.utc
+    ///
+    /// [(JavaScript docs)](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/utc)
+    #[inline]
+    fn utc(args: &[i32] ) -> f64 {
```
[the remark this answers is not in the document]
bbyler: I appreciate the feedback/specs link - I wasn't aware of the default args (though I had a sneaking suspicion), coming from the MDN docs at least. The difference in semantics between Rust and JS makes things challenging. In fact, I dislike the fact that sending no args doesn't throw an exception; rather, a lot of these functions return something like NaN. Or, for args that don't make sense (negative numbers, fractions, numbers outside the typical month/day range), something like "Jan 1 1969" is returned. I can see why it's this way, but it'd be nice to change this behavior from Rust. I'm with you on your proposed function signature

src/webapi/date.rs
```diff
@@ -1,22 +1,903 @@
...
+        }
+    }
+
+    /// The Date.parse method parses a string representation of a date, and returns the number of
+    /// milliseconds since 1 January 1970 00:00:00 UTC, or NaN if the string is unrecognize or, in
+    /// some cases, contains illegal data values (e.g. 2015-02-31)
+    ///
+    /// [(JavaScript docs)](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/parse)
+    #[inline]
+    fn parse( _date_string: &str ) -> f64 {
```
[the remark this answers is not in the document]
bbyler: I did this to escape linter warnings...I'll see what happens when I remove "_"s isA

src/webapi/date.rs
```diff
@@ -1,22 +1,903 @@
...
+        em_asm_double!( "return Date.parse(@{_date_string}" )
```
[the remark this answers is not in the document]
bbyler: Admittedly, it took me a while to get comfortable with js! and em_asm_double! / understand their usage

src/webapi/date.rs
```diff
@@ -1,22 +1,903 @@
...
+    }
+
+    ///
+    ///
+    /// [(JavaScript docs)](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/getDate)
+    #[inline]
+    fn get_date( &self ) -> i32 {
+        js!( return @{self.as_ref()}.getDate() ).try_into().unwrap()
```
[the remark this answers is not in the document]
bbyler: Thanks!

src/webapi/date.rs
```diff
@@ -1,22 +1,903 @@
...
+    #[inline]
+    fn get_day( &self ) -> i32 {
+        js!( return @{self.as_ref()}.getDay(); ).try_into().unwrap()
+    }
+
+    ///
+    ///
+    /// [(JavaScript docs)](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/getFullYear)
+    #[inline]
+    fn get_full_year( &self ) -> f64 {
```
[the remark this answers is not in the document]
bbyler: I'm in the process of fixing a ton of types that weren't really gelling with JS land and were just inconsistent

src/webapi/date.rs
bbyler: Thanks!

src/webapi/date.rs
bbyler: Thanks!

src/webapi/date.rs
bbyler: I have mixed feelings on this one. Overall I don't mind building a typed hash map, but to your point I could see it getting repetitive/verbose. On the flip side, I like being able to utilize compile-time checks to make sure function calls/signatures make sense

src/webapi/date.rs
bbyler: Thanks!

src/webapi/date.rs
bbyler: Good catch!!

src/webapi/date.rs
bbyler: Good catch!!

src/webapi/date.rs
bbyler: Totally agree. In fact, I'm in the process of removing any return type from setter methods. Hopefully it'll be a little more flexible that way

koute:
@bbyler Taking a quick glance this already looks better! That said I can still see a few things I'd like changed. I'll take a more thorough look at this again (it's a lot of code after all!) in a day or two once I have some more free time. Sorry for the delay! (:

bbyler:
@koute  Thanks! 

Totally off topic, but question for you: 

Background: I noticed the src/ecosystem mods there, with serde and serde_json already present. Great idea btw! 

I'm particularly curious about ecosystem/serde_json. It doesn't look like it's really used anywhere, but it has a critical ```TryFrom``` implementation. Question 0: as it stands, does this work? (Looks good to me...)

If it does, then here's the situation: say I have a block of serde JSON, e.g.

```rust
let s =  r#"{"greeting_string": "hello"}"#;
let json_object: Value = serde_json::from_str(s)?; <-- serde_json Value, not stdweb Value...
// pretty cool that this works, without strong typing
println!(json_object["greeting_string"]); // hello
```

Nothing new here. 

Say we want to do something with it in stdweb, e.g.

```rust
use serde_json::value::Value as JsonValue;
use webcore::value::Value;

fn say_hello_to_world(json_value: &JsonValue) {
    // assuming a json_value object has a key and value for "greeting_string" and so on
    js!( @(no_return)
        console.log(@{json_value["greeting_string"]} + " world!");
    );
}
```

This example will break, since a serde JsonValue can't be marshaled into a JS object *without that ecosystem tie-in* (or a similar tie-in or implementation). How can that be done?

I've tried some stuff on my own and it blew up in my face...

I'm purposely not opening an issue just to try to keep the issue count/backlog to a minimum for now. It's also more of a question...and with an answer I can see about documenting it in the README

Thanks!

src/ecosystem/m/querystring/build.rs
```diff
@@ -0,0 +1,104 @@
+use std::borrow::Cow;
```
koute: This looks like new functionality?

For now I'd suggest you should put stuff like this in a separate repository/crate. (:

In general I don't want to put too much original stuff in `stdweb`; I just want the very basics here, and anything extra should most likely live as a separate crate. (You might ask then - what is `serde` integration doing here? And the answer to that is simple - `serde` is extremely popular in the Rust community, and you **can't** integrate it with `stdweb` outside of `stdweb` due to Rust's trait implementation coherence rules.)

src/webapi/date.rs
```diff
@@ -1,25 +1,1166 @@
...
-#[derive(Debug)]
-pub struct Date {
-    dummy: PhantomData< () >
-}
-
-impl Date {
-    /// The Date.now() method returns the number of milliseconds elapsed since 1 January 1970 00:00:00 UTC.
+/// A date object
+///
+/// [(JavaScript docs)](https://developer.mozilla.org/en-US/docs/Web/API/Date)
```
koute: Broken link.

src/webapi/date.rs
```diff
@@ -1,25 +1,1166 @@
...
+pub trait IDate: AsRef< Reference > + TryFrom< Value >  {
```
koute: Does anything actually inherit from Date? (I don't think so, but I might be missing something.) If that's the case then I don't see a point in making an interface out of it; just put all of the methods in an `impl Date`.

src/webapi/date.rs
```diff
@@ -1,25 +1,1166 @@
...
+        js!( return Date.now(); ).try_into().unwrap()
+    }
+
+    /// The Date::utc() method accepts the same parameters as the longest form of the constructor,
+    /// and returns the number of milliseconds in a Date object since January 1, 1970, 00:00:00,
+    /// universal time
+    ///
+    /// [(JavaScript docs)](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/utc)
+    #[inline]
+    fn utc( year: i32, _month: Option< i32 >, _day: Option< i32 >, _hour: Option< i32 >, _minutes: Option< i32 >, _seconds: Option< i32 >, _milliseconds: Option< i32 >) -> i64 {
```
koute: Yep, this is better!

However I still think the `Option`s are unnecessary; taking everything as `i32` would be simpler with very little downsides.

src/webapi/date.rs
```diff
@@ -1,25 +1,1166 @@
...
+
+    /// The to_s() method returns a string representing the specified Date object.
+    ///
+    /// [(JavaScript docs)](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/toString)
+    #[inline]
+    // naming this to to_s 1. to avoid collision with standard to_string(), and 2. to avoid linting
+    // error on toString()...
+    fn to_s( &self ) -> String {
+        js! ( return @{self.as_ref()}.toString(); ).try_into().unwrap()
+    }
```
koute: Again, I don't think this method should be here. (: As a Rubyist seeing `to_s` warms my heart, but if we want to expose it then we should do it as `impl From` to be idiomatic Rust (or not do it at all for now)

src/webapi/date.rs
```diff
@@ -1,25 +1,1166 @@
...
+
+                 None => return js! (
+                     return @{self.as_ref()}.toLocaleDateString(@{_locales});
+                 ).try_into().unwrap()
+            },
+
+            None => return js! ( return @{self.as_ref()}.toLocaleDateString(); ).try_into().unwrap()
+        }
+
+    }
```
koute: Get rid for now. (:

src/webapi/date.rs
```diff
@@ -1,25 +1,1166 @@
...
+                ).try_into().unwrap(),
+
+                None => return js! (
+                    return @{self.as_ref()}.toLocaleString(@{_locales});
+                ).try_into().unwrap()
+            },
+
+            None => return js! ( return @{self.as_ref()}.toLocaleString(); ).try_into().unwrap()
+        }
+    }
```
koute: Get rid for now. (:

src/webapi/date.rs
```diff
@@ -1,25 +1,1166 @@
...
+                ).try_into().unwrap(),
+
+                None => return js! (
+                    return @{self.as_ref()}.toLocaleTimeString(@{_locales});
+                ).try_into().unwrap()
+            },
+
+            None => return js! ( return @{self.as_ref()}.toLocaleTimeString(); ).try_into().unwrap()
+        }
+    }
```
koute: Get rid for now. (:

src/webapi/date.rs
```diff
@@ -1,25 +1,1166 @@
...
+                    @{self.as_ref()}.setUTCFullYear(@{year});
+                };
+            }
+        }
+    }
+
+    /// The set_utc_hours() method sets the hour for a specified date according to universal time,
+    /// and returns the number of milliseconds since 1 January 1970 00:00:00 UTC until the time
+    /// represented by the updated Date instance.
+    ///
```
koute: Please get rid of the method name from all of these doc comments, e.g.:

`The set_utc_hours() method sets the hour` -> `Sets the hour`

src/webapi/date.rs
```diff
@@ -1,25 +1,1166 @@
...
+                js! { @(no_return)
+                    @{self.as_ref()}.setSeconds(@{seconds}, @{_ms});
+                };
+            },
+            None => {
+                js! { @(no_return)
+                    @{self.as_ref()}.setSeconds(@{seconds});
+                };
+            }
+        }
```
koute: Hmmmm... I'm really conflicted on how we should handle methods like this. I **really** don't like the optional arguments, and after I took a better look at the APIs they seem to be mostly unnecessary. For example, if you want to set the milliseconds you can either use `setSeconds(..., ms)` or `setMilliseconds(ms)` - two methods which do the same thing, it's just that `setSeconds` also sets the seconds at the same time.

So how about we have these methods do only one thing? (e.g. `setSeconds` only sets the seconds, etc.)

src/webapi/date.rs
bbyler: sorry about that, thanks for the catch! Same with the (highly experimental) ecosystem stuff - accidentally polluted this PR

src/webapi/date.rs
bbyler: 🎉  for the Ruby shoutout  ! On thinking over it a little more, I have to agree. For one, it doesn't seem like this would really be that useful in practice, and two, we would need to put a little more thought into a true "to_string" method from the Rust side, in case we wanted to show what Rust thinks of the object as a whole (to be used for debugging/inspection etc.)

src/webapi/date.rs
bbyler: definitely had the same thought while implementing this: setters should set what they say they're setting...but followed the documentation through. I'd be happy to make these single arg functions

examples/todomvc/src/main.rs
```diff
@@ -236,6 +239,33 @@ fn main() {
...
+    let d: Date = js! (
+        return new Date(2000, 0, 1, 12, 5, 7, 10);
+    ).try_into().unwrap();
+
+    let d_utc = d.get_time();
+
+    js!(
+        console.log(@{d_utc});
+    );
+
```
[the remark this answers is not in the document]
bbyler: I'd be happy to undo the todomvc stuff - it's a little tricky finding a place for new features when the example is really just meant to be re-used over and over as is (with different implementations). It was helpful as a testbed while I couldn't get tests up and running

src/webapi/date.rs
bbyler: I can see ```toLocaleDateString```, ```toLocaleString```, and ```toLocaleTimeString``` being useful without specifying the locale and options explicitly. If someone is OK with the defaults, then these calls may still be beneficial for simple date formatting/presentation

src/webapi/date.rs
bbyler: Sounds fair enough - if we're not shooting for exact parity with the browser's UTC usage, then I can make the change. in progress

src/webapi/date.rs
bbyler: Correct me if I'm wrong, but does Date need to be a trait in order to coerce a ```js!``` statement into a date "object"? For example, if Date was an ```impl```, would we still be able to do this?

```rust
let d: Date = js!(
    return new Date();
)
```

koute: It doesn't need to be a trait. Please take a look at, for example, `array_buffer.rs` for a minimal example on how to do it.

src/webapi/date.rs
koute: We're definitely not aiming for absolute 100% coverage of the APIs, **especially** for APIs which would be super awkward like this one and aren't really essential. (I know this reasoning is kind-of wishy-washy, but it's hard to define an absolute rule to which we can adhere to besides using common sense.)

src/webapi/date.rs
```diff
@@ -1,25 +1,903 @@
...
+
+    /// Parses a string representation of a date, and returns the number of
+    /// milliseconds since 1 January 1970 00:00:00 UTC, or NaN if the string is unrecognize or, in
+    /// some cases, contains illegal data values (e.g. 2015-02-31)
+    ///
+    /// [(JavaScript docs)](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/parse)
+    #[inline]
+    fn parse( date_string: &str ) -> i64 {
+        js!( return Date.parse( @{date_string} ); ).try_into().unwrap()
+    }
```
koute: Looking at the [specs](https://tc39.github.io/ecma262/#sec-date.parse) this function can return `NaN`, which would trigger a panic. We don't want panics, so it'd be better to check for `NaN` and return an `Option`. Also modify the doc comment to reflect this and write a test.

src/webapi/date.rs
```diff
@@ -1,25 +1,903 @@
...
+        js!( return Date.now(); ).try_into().unwrap()
+    }
+
+    /// Accepts the same parameters as the longest form of the constructor,
+    /// and returns the number of milliseconds in a Date object since January 1, 1970, 00:00:00,
+    /// universal time
+    ///
+    /// [(JavaScript docs)](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/utc)
+    #[inline]
+    fn utc( year: i32, month: u8, day: u8, hour: u8, minutes: u16, seconds: u16, milliseconds: u32) -> i64 {
```
koute: 1. These arguments should probably all be `i32`.
   * To be consistent with what we return from other methods (e.g. the result of `get_year` should be directly pluggable into `utc` without casting)
   * Putting negative values in here, while kind-of silly, seems to be not disallowed by the [specs](https://tc39.github.io/ecma262/#sec-date.utc), and looks like it's working in both Firefox and Chromium.

2. And the return value should be consistent with `Date::now`, since it's returning the same thing. (So I guess we can make it as `f64` here?)

src/webapi/date.rs
```diff
@@ -1,25 +1,903 @@
...
+    /// returns the number of milliseconds since 1 January 1970 00:00:00 UTC until the time
+    /// represented by the updated Date instance.
+    ///
+    /// [(JavaScript docs)](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/setHours)
+    #[inline]
+    fn set_hours( &self, hour: i32 ) {
+        js! { @(no_return)
+            @{self.as_ref()}.setHours( @{hour} );
+        }
+    }
```
koute: The doc comments say that the method returns something, and yet it doesn't. (:

Since the original JavaScript API returns a value here we should *probably* also return a value here. Ditto for other methods like this. (Right now it's not going to be super efficient, but I do have plans to make JS interop cheaper in the future.)

src/webapi/date.rs
koute: Hmm... okay, that sounds fine to me, especially since the original JS API can be called that way. We can add an ~~overload~~ another method later which will take the arguments if we wish so.

src/webapi/date.rs
```diff
@@ -1,25 +1,903 @@
...
+
+reference_boilerplate! {
+    Date,
+    instanceof Date
+}
+
+/////////////////////////////////////////////////////////////////////
+
+#[cfg(test)]
+mod tests {
```
koute: Your tests are failing; please fix them. (:

PR #39, opened 2017-12-18
added requestAnimationFrame to window

keeslinp:
I added the https://developer.mozilla.org/en-US/docs/Web/API/window/requestAnimationFrame to window. I ended up copying the funcall_adapter function over temporarily to verify that it actually works, but I'm assuming there should be some sort of shared module that it can go in that would be better?


keeslinp:
I did a little more reading and realized that I left out the delta value. Is there a better way to use the callback so that I can pass a value into it? funcall_adapter doesn't seem to support that. I saw a trait in serialize that seems to have an arguments value but I'm not sure how to use it.

koute:
For now you could *probably* use the `js!` macro. It will be slower than using a direct `__js_raw_asm!` call, but we can speed it up later. (In general I would prefer to make the `js!` macro have less overhead than to mess around with doing interop by hand everywhere.)

Also, please create a `RequestAnimationFrameHandle` structure and return it from this function, and add `fn cancel( self )` to it which will call `cancelAnimationFrame`; we have something similar for event listeners in `EventListenerHandle`.

keeslinp:
Just updated to include your request. Sorry it took a while, it's been a crazy week.

src/webapi/window.rs
```diff
@@ -86,4 +88,17 @@ impl Window {
             ).into_reference_unchecked()
         }
     }
+
+    /// Sets a callback in requestAnimationFrame to call on next frame
+    /// Callback accepts an f64 as an argument
+    ///
+    /// [(JavaScript docs)](https://developer.mozilla.org/en-US/docs/Web/API/window/requestAnimationFrame)
+    pub fn request_animation_frame< F: FnMut(f64) + 'static>( &self, callback: F) -> RequestAnimationFrameHandle {
```
koute: Correct me if I'm wrong, but shouldn't this be `FnOnce` instead of `FnMut`, right? If that's the case then please change it. To pass an `FnOnce` you need to wrap it with `webcore::serialization::Once`.

src/webapi/request_animation_frame_handle.rs
```diff
@@ -0,0 +1,23 @@
...
+/// Use webapi::window::request_animation_frame to create this
+#[derive(Debug)]
+pub struct RequestAnimationFrameHandle {
+    id: u32,
+}
+
+impl RequestAnimationFrameHandle {
+    /// This id comes from here:
+    /// [(Javascript docs)](https://developer.mozilla.org/en-US/docs/Web/API/window/requestAnimationFrame)
+    pub fn new(id: u32) -> RequestAnimationFrameHandle {
```
koute: Please delete this function; I don't really see a point in exposing this.

src/webapi/window.rs
```diff
@@ -86,4 +88,17 @@ impl Window {
...
+        let id : u32 = js!{
+            var callback = @{callback};
+            var request = requestAnimationFrame(callback);
+            return request;
+        }.try_into().unwrap_or(0);
```
koute: Why the `unwrap_or`? Shouldn't it be simply `unwrap`?

src/webapi/request_animation_frame_handle.rs
```diff
@@ -0,0 +1,23 @@
...
+        RequestAnimationFrameHandle {
+            id,
+        }
+    }
+    /// This cancels an animation frame request
+    /// [(Javascript docs)](https://developer.mozilla.org/en-US/docs/Web/API/Window/cancelAnimationFrame)
+    pub fn cancel(&self) {
```
koute: `&self` -> `self`

src/webapi/window.rs
```diff
@@ -86,4 +88,17 @@ impl Window {
...
     }
+
+    /// Sets a callback in requestAnimationFrame to call on next frame
+    /// Callback accepts an f64 as an argument
+    ///
+    /// [(JavaScript docs)](https://developer.mozilla.org/en-US/docs/Web/API/window/requestAnimationFrame)
+    pub fn request_animation_frame< F: FnMut(f64) + 'static>( &self, callback: F) -> RequestAnimationFrameHandle {
+        let id : u32 = js!{
+            var callback = @{callback};
+            var request = requestAnimationFrame(callback);
```
koute: I'm not sure if this has any practical consequences, but this probably should be `@{self}.requestAnimationFrame`. (Since `self` is a particular window instance.)

src/webapi/window.rs
```diff
@@ -86,4 +88,17 @@ impl Window {
             ).into_reference_unchecked()
         }
     }
+
+    /// Sets a callback in requestAnimationFrame to call on next frame
+    /// Callback accepts an f64 as an argument
+    ///
+    /// [(JavaScript docs)](https://developer.mozilla.org/en-US/docs/Web/API/window/requestAnimationFrame)
+    pub fn request_animation_frame< F: FnMut(f64) + 'static>( &self, callback: F) -> RequestAnimationFrameHandle {
+        let id : u32 = js!{
```
koute: Why `u32`? This should probably be `i32`, which is the "native" integer type in the JS world.

src/webapi/request_animation_frame_handle.rs
```diff
@@ -0,0 +1,23 @@
...
+        js!{
+            cancelAnimationFrame(@{self.id});
```
koute: This is not enough; if someone cancelled this *before* it was triggered then the callback needs to be dropped too, or it will leak memory.

src/webapi/window.rs
```diff
@@ -86,4 +88,17 @@ impl Window {
             ).into_reference_unchecked()
         }
     }
+
+    /// Sets a callback in requestAnimationFrame to call on next frame
+    /// Callback accepts an f64 as an argument
```
koute: Please copy the docs from two first paragraphs from MDN and lightly edit it. (:

src/webapi/request_animation_frame_handle.rs
```diff
@@ -0,0 +1,23 @@
+/// Struct representing an animation frame request
+/// Use webapi::window::request_animation_frame to create this
```
koute: `Struct representing an animation frame request` -> `A handle to a pending animation frame request.`, or something like that.

If you want to point to the function which creates it then you can link to it properly (the docs are markdown; look at other doc comments in other files for examples on how to do it)

src/webapi/mod.rs
```diff
@@ -22,3 +22,4 @@ pub mod file_list;
 pub mod file_reader;
 pub mod array_buffer;
 pub mod typed_array;
+pub mod request_animation_frame_handle;
```
koute: No need for this to be in a separate file; just put it with `Window` (:

src/webapi/request_animation_frame_handle.rs
```diff
@@ -0,0 +1,23 @@
...
+
+impl RequestAnimationFrameHandle {
+    /// This id comes from here:
+    /// [(Javascript docs)](https://developer.mozilla.org/en-US/docs/Web/API/window/requestAnimationFrame)
+    pub fn new(id: u32) -> RequestAnimationFrameHandle {
+        RequestAnimationFrameHandle {
+            id,
+        }
+    }
+    /// This cancels an animation frame request
```
koute: `This cancels an animation frame request` -> `Cancels an animation frame request.` (:

Also please put an extra newline before `[(Javascript docs)]`

src/webapi/window.rs
keeslinp: I wasn't sure how to take care of this part. I couldn't find an example of how to wrap a FnOnce in a Once. In order to do it directly I had to make the field public. If there's another way to do this, I'd be willing to do that :).

src/webapi/window.rs
keeslinp: I mostly just did unwrap_or out of habit, changed it.

src/webapi/window.rs
keeslinp: I'm pretty sure they do the same thing, but I added it for consistency.

src/webapi/window.rs
keeslinp: I think the id is always a positive integer, but that might be totally imagined by me. To make things line up easier I changed it to `i32`

src/webapi/request_animation_frame_handle.rs
keeslinp: In order to manage this, I had to do some stuff that might be a bit weird. I basically stored the whole handle in javascript land and just used it in inlined javascript.

src/webcore/serialization.rs
```diff
@@ -891,7 +891,7 @@ impl JsSerializable for Value {
 __js_serializable_boilerplate!( Value );
 
 // A wrapper for passing FnOnce callbacks into the `js!` macro.
-pub struct Once< T >( T );
+pub struct Once< T >( pub T );
```
koute: Oh right, yes. This should have been public. (:

src/webapi/window.rs
```diff
@@ -86,4 +106,25 @@ impl Window {
...
+    /// each receive the same timestamp even though time has passed during the computation of every previous callback's workload.
+    /// This timestamp is a decimal number, in milliseconds, but with a minimal precision of 1ms (1000 µs).
+    ///
+    /// [(JavaScript docs)](https://developer.mozilla.org/en-US/docs/Web/API/window/requestAnimationFrame)
+    pub fn request_animation_frame< F: FnOnce(f64) + 'static>( &self, callback: F) -> RequestAnimationFrameHandle {
+        let values: Value = js!{
+            var callback = @{Once(callback)};
+            var request = @{self}.requestAnimationFrame(callback);
+            return { request, callback };
+        }.try_into().unwrap();
```
koute: The `js!` macro returns a `Value` by default, so `try_into().unwrap()` is unnecessary.

src/webapi/window.rs
```diff
@@ -86,4 +106,25 @@ impl Window {
...
+    /// callbacks queued by requestAnimationFrame() begin to fire. Multiple callbacks in a single frame, therefore,
+    /// each receive the same timestamp even though time has passed during the computation of every previous callback's workload.
+    /// This timestamp is a decimal number, in milliseconds, but with a minimal precision of 1ms (1000 µs).
+    ///
+    /// [(JavaScript docs)](https://developer.mozilla.org/en-US/docs/Web/API/window/requestAnimationFrame)
+    pub fn request_animation_frame< F: FnOnce(f64) + 'static>( &self, callback: F) -> RequestAnimationFrameHandle {
+        let values: Value = js!{
+            var callback = @{Once(callback)};
+            var request = @{self}.requestAnimationFrame(callback);
+            return { request, callback };
```
koute: Returning an array here would probably be very slightly more efficient instead of an object. Maybe? (Honestly I don't know!) Either use an array, or an object initialized the old way. (This syntax is ES6+ only.)

src/webapi/window.rs
```diff
@@ -3,6 +3,26 @@ use webapi::event_target::{IEventTarget, EventTarget};
...
+pub struct RequestAnimationFrameHandle(Value);
+
+impl RequestAnimationFrameHandle {
+    /// Cancels an animation frame request.
+    ///
+    /// [(Javascript docs)](https://developer.mozilla.org/en-US/docs/Web/API/Window/cancelAnimationFrame)
+    pub fn cancel(self) {
+        js!{
+            let val = @{self.0};
+            cancelAnimationFrame(val.request);
```
koute: This should probably also be called on the same `Window` instance which was used in `request_animation_frame` instead of on the globally implicit `window`. (Again, I don't know if this *really* matters or will ever matter in practice; we can always change it later. Since you're passing a `Value` around putting a reference to the `window` in there is going to be essentially free anyway.)

src/webapi/window.rs
```diff
@@ -3,6 +3,26 @@ use webapi::event_target::{IEventTarget, EventTarget};
...
+#[derive(Debug)]
+pub struct RequestAnimationFrameHandle(Value);
+
+impl RequestAnimationFrameHandle {
+    /// Cancels an animation frame request.
+    ///
+    /// [(Javascript docs)](https://developer.mozilla.org/en-US/docs/Web/API/Window/cancelAnimationFrame)
+    pub fn cancel(self) {
+        js!{
+            let val = @{self.0};
```
koute: Use `var` instead of `let`. (We support Emscripten, which doesn't like ES6+ features.)

src/webapi/window.rs
keeslinp: objects and array are just syntactic sugar a lot of times in javascript. They end up getting stored the same a lot of the time. I just did it this way to make it clearer when I used the values later. I'll change it to the es5 way

keeslinp:
Changes made :). Thanks for all your help! I tried squashing, but I'm not super good at editing git history. I think I mostly got it right, but for some reason, conflicts showed up and I had to fix those so it added to the history? I'm not 100% sure how to get that fixed. If you know how I'd love to learn :).

koute:
> I'm not 100% sure how to get that fixed. If you know how I'd love to learn :).

I'm not an expert on git myself, but you could do something like this:

    $ git remote add upstream https://github.com/koute/stdweb.git
    $ git fetch --all
    $ git checkout request-animation
    $ git rebase upstream/master
    # Resolve the conflict with your favorite editor (grep for "HEAD"):
    $ vim src/webcore/serialization.rs
    $ git add src/webcore/serialization.rs
    $ git rebase --continue
    $ git push --force

Basically this will grab the newest `master` and reapply your changes on top of it.

In general I'm a fan of [a rebase workflow](http://kensheedlo.com/essays/why-you-should-use-a-rebase-workflow/), but it's not a *huge* deal.

> Thanks for all your help!

Thanks for your contribution. (:

RangerMauve:
You can do squashing through github if you go into the [project settings](https://github.com/koute/stdweb/settings#merge_types_squash)

koute:
@RangerMauve Oh, that's nice to know! I totally didn't notice that. Are there any docs on how it handles merge commits like these two here?

RangerMauve:
@koute When you squash it puts all the commits from the PR into a single commit that references the PR number, so if you did a merge commit it'd just be part of it. I haven't done rebasing yet, though.

I found [this](https://help.github.com/articles/about-pull-request-merges/#squash-and-merge-your-pull-request-commits) for docs, but I'm not sure if there's anything else.

keeslinp:
So does that mean we're going to let github handle it?

koute:
Hmmm... looks like it worked!

@keeslinp Thanks for the contribution! @RangerMauve Thanks for telling me that Github can do it automatically!

chpio:
`RequestAnimationFrameHandle` isn't public/exported

keeslinp:
https://docs.rs/stdweb/*/stdweb/web/struct.Window.html#method.request_animation_frame

Does this not work? I had it working for me. I just did
`window().requestAnimationFrame(...)` and it worked fine from another
project.

chpio:
Yeah, it does work, but you can't store `RequestAnimationFrameHandle` as it's not accessible.

keeslinp:
Ah, now I see what you mean. @koute, what mod would you like to be in? It's in the window.rs file so I could see it being in that file but that might be a little unintuitive to find.

koute:
@keeslinp You mean in what namespace to export it? You can put it in `stdweb::web` where `EventListenerHandle` is too.

keeslinp:
I fixed the problem, but I'm not sure how to reopen the pull request? I pushed the changes to my branch and it doesn't appear to have reflected on this PR. Do I need to open another one?

koute:
> Do I need to open another one?

Yes.

PR #40, opened 2017-12-18
Add `InputEvent`

mrdziuban:
The `input` event is often useful as an alternative to `change` since "Unlike the input event, the change event is not necessarily fired for each change to an element's value."


koute:
LGTM; thanks!

Issue #9, opened 2017-06-20
Compiling in release mode fails


FliegendeWurst:
I'm trying to run the todo example.

    $ cargo web start --port 8008 --release
    error: linking with `emcc` failed: exit code: 1
      |
[...]

PR #42, opened 2017-12-21
Speeding CI builds

lilianmoraru:
`cargo`'s incremental compilation is reliable when not using `build.rs`.
This makes Travis take advantage of that.
`cargo-web` linking will still take a long time.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,4 +1,5 @@
 language: rust
+cache:    cargo
 
 dist: trusty
 sudo: false
--- a/ci.sh
+++ b/ci.sh
@@ -21,7 +21,10 @@ if [ "$IS_NIGHTLY" = "0" ]; then
     fi
 fi
 
-cargo install cargo-web -f
+cargo_web_dest="cargo-web-install"
+rm -rf "${cargo_web_dest}"
+git clone https://github.com/koute/cargo-web.git -b 0.4.0 --depth 1 "${cargo_web_dest}"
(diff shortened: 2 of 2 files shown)


lilianmoraru:
Uncached: https://travis-ci.org/koute/stdweb/builds/319606978
Cached: https://travis-ci.org/lilianmoraru/stdweb/builds/319721056

koute:
This looks great, thanks!

However, I'd like to keep the `--release` flag as-is, mainly because of issues like https://github.com/koute/stdweb/issues/9 where the problem only appeared when building in release mode. (And I *think* we're spending the majority of the compile time in macro expansion due to the `js!` macro, so release vs debug shouldn't be *that* different?)

lilianmoraru:
@koute What do you think about testing also in release mode?
I'm not sure that this would be better but the idea would be:
- Building in release might find other issues because of the optimized output?
- `cargo web build` won't have to rebuild the dependencies it has on `stdweb`(currently `stdweb` is compiled in `debug` mode, so `cargo web build --release` has to build the `release` version of its dependencies)
- The release `target` will also be cached, so this should not affect compile times much(maybe linking optimizations and `stdweb` itself).

I'll underscore again that I am not sure that this would be better, just guesses(reasonable expectations?).

koute:
@lilianmoraru  Hmmm... I guess testing in the release mode would be fine, provided `debug-assertions` will be still set to true. (So I'd guess we'd have to add `[profile.test]` to `Cargo.toml` and set `opt-level = 3`?)

PR #46, opened 2017-12-22
Publish TextNode struct

DenisKolodin:
[`Document::create_text_node`](https://docs.rs/stdweb/*/stdweb/web/struct.Document.html#method.create_text_node) returns `TextNode` struct, but actually it's private and cannot be imported:

```rust
error[E0432]: unresolved import `stdweb::web::TextNode`
  |
9 | use stdweb::web::{INode, IElement, Node, Element, TextNode, document};
  |                                                   ^^^^^^^^ no `TextNode` in `web`
```

This PR fixes it.


koute:
Thanks!

PR #49, opened 2017-12-24
Implement bindings to the History API

tomhoule:
They work in the small example included in the PR. There are however a few questions left open:

- I don't think the unit test for `PopStateEvent` is run, it never fails.
- Is the signature for `History::state` the preferred way to extract typed values from js?
- Should the example stay there? It might gain from being a bit less trivial in that case.
- Is the code formatted correctly?

Looking forward to your feedback :)

### DIFF
--- a/src/webapi/history.rs
+++ b/src/webapi/history.rs
@@ -1,0 +1,90 @@
+use webcore::value::Reference;
+use webcore::try_from::TryInto;
+use webcore::serialization::JsSerializable;
+
+/// [(JavaScript docs)](https://developer.mozilla.org/en-US/docs/Web/API/History)
+pub struct History(Reference);
+
+reference_boilerplate! {
+    History,
+    instanceof History
+}
+
+impl History {
+    /// Adds a new entry to history.
+    ///
+    /// pushState() takes three parameters: a state object, a title (which is currently ignored),
+    /// and (optionally) a URL. Let's examine each of these three parameters in more detail:
+    ///
+    /// - state object — The state object is a JavaScript object which is associated with the new
+    /// history entry created by pushState(). Whenever the user navigates to the new state, a
+    /// popstate event is fired, and the state property of the event contains a copy of the history
+    /// entry's state object.
+    ///
+    /// - title — Firefox currently ignores this parameter, although it may use it in the future.
+    /// Passing the empty string here should be safe against future changes to the method.
+    /// Alternatively, you could pass a short title for the state to which you're moving.
+    ///
+    /// - URL — The new history entry's URL is given by this parameter. Note that the browser won't
+    /// attempt to load this URL after a call to pushState(), but it might attempt to load the URL
+    /// later, for instance after the user restarts the browser. The new URL does not need to be
+    /// absolute; if it's relative, it's resolved relative to the current URL. The new URL must be
+    /// of the same origin as the current URL; otherwise, pushState() will throw an exception.
+    /// This parameter is optional; if it isn't specified, it's set to the document's current URL.
+    ///
+    /// [(JavaScript docs)](https://developer.mozilla.org/en-US/docs/Web/API/History_API#The_pushState%28%29_method)
+    pub fn push_state<T: JsSerializable>(&self, state: T, title: &str, url: Option<&str>) {
+        js!{ @(no_return)
+            @{self}.pushState(@{state}, @{title}, @{url});
+        };
+    }
+
+    /// Operates exactly like history.push_state() except that replace_state() modifies the current
+    /// history entry instead of creating a new one. Note that this doesn't prevent the creation of
+    /// a new entry in the global browser history.
+    ///
+    /// [(JavaScript docs)](https://developer.mozilla.org/en-US/docs/Web/API/History_API#The_replaceState%28%29_method)
+    pub fn replace_state<T: JsSerializable>(&self, state: T, title: &str, url: Option<&str>) {
+        js!{ @(no_return)
+            @{self}.replaceState(@{state}, @{title}, @{url});
+        };
+    }
+
+    /// You can use the go() method to load a specific page from session history, identified by its
+    /// relative position to the current page (with the current page being, of course, relative
+    /// index 0).
+    ///
+    /// [(JavaScript docs)](https://developer.mozilla.org/en-US/docs/Web/API/History_API#Traveling_through_history)
+    pub fn go(&self, offset: i32) {
+        js! { @(no_return)
+            @{self}.go(@{offset});
+        };
+    }
+
+    /// Move one step backward through history.
+    ///
+    /// [(JavaScript docs)](https://developer.mozilla.org/en-US/docs/Web/API/History_API#Traveling_through_history)
+    pub fn back(&self) {
+        js! { @(no_return)
+            @{self}.back();
+        };
+    }
+
+    /// Move one step forward through history.
+    ///
+    /// [(JavaScript docs)](https://developer.mozilla.org/en-US/docs/Web/API/History_API#Traveling_through_history)
+    pub fn forward(&self) {
+        js! { @(no_return)
+            @{self}.forward();
+        };
+    }
+
+    /// Returns the current number of history entries.
+    ///
+    /// [(JavaScript docs)](https://developer.mozilla.org/en-US/docs/Web/API/History)
+    pub fn len(&self) -> usize {
+        js!(
+            return @{self}.length;
+        ).try_into().unwrap()
+    }
+}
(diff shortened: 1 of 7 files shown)


src/webapi/history.rs
```diff
@@ -0,0 +1,64 @@
...
+reference_boilerplate! {
+    History,
+    instanceof History
+}
+
+impl History {
+    /// `History.pushState` allows you to add history entries.
+    ///
+    /// [(JavaScript docs)](https://developer.mozilla.org/en-US/docs/Web/API/History_API#The_pushState%28%29_method)
+    pub fn push_state<T: JsSerializable>(&self, state: T, title: Option<&str>, url: String) {
```
koute: In the [specs](https://html.spec.whatwg.org/multipage/history.html#the-history-interface) I can see that the `title` is marked as being non-optional (is not inside square brackets) while the `url` is, so shouldn't the prototype be `&self, state: T, title: &str, url: Option< &str >`?

src/webapi/history.rs
```diff
@@ -0,0 +1,64 @@
...
+/// [(JavaScript docs)](https://developer.mozilla.org/en-US/docs/Web/API/History)
+pub struct History(Reference);
+
+reference_boilerplate! {
+    History,
+    instanceof History
+}
+
+impl History {
+    /// `History.pushState` allows you to add history entries.
```
koute: In general we don't reference the original JS `camelCased` name in the docs, and there is no point in referencing the name of the function you're documenting anyway. (:

koute:
> I don't think the unit test for PopStateEvent is run, it never fails.

Yes, it's not run. You need to delete the `[cfg(...)]` there and run the test without passing `--nodejs` to `cargo-web` (it will then run the test in a web browser, which is kinda hack-ish and experimental for now, which is why it isn't yet done by default)

> Should the example stay there? It might gain from being a bit less trivial in that case.

I don't mind extra examples. (: 

> Is the code formatted correctly?

Don't worry about it; worrying about formatting is not something humans should waste time on. I'll have to run the whole project through `rustfmt` sooner or later.

tomhoule:
Just a quick message to say my laptop died so it will be a few days until I replace it and amend the PR :)

koute:
@tomhoule Sure; thanks for the heads up. (:

tomhoule:
I managed to free up some time to amend the PR. I think I addressed everything from the last review, managed to run the tests, but I haven't improved the example. It just does some light history manipulation and logs to the console.

Maybe rendering history state to the DOM would make it more convincing? I'll try to think of something better over the weekend :)

src/webapi/history.rs
```diff
@@ -0,0 +1,91 @@
...
+            @{self}.go(@{offset})
+        };
+    }
+
+    /// Move one step backward through history.
+    ///
+    /// [(JavaScript docs)](https://developer.mozilla.org/en-US/docs/Web/API/History_API#Traveling_through_history)
+    pub fn back(&self) {
+        js! { @(no_return)
+            @{self}.back()
```
koute: Missing a semicolon. (:

src/webapi/history.rs
```diff
@@ -0,0 +1,91 @@
...
+        };
+    }
+
+    /// Move one step forward through history.
+    ///
+    /// [(JavaScript docs)](https://developer.mozilla.org/en-US/docs/Web/API/History_API#Traveling_through_history)
+    pub fn forward(&self) {
+        js! { @(no_return)
+            @{self}.forward()
```
koute: Missing a semicolon here too.

src/webapi/history.rs
```diff
@@ -0,0 +1,91 @@
...
+    }
+
+    /// You can use the go() method to load a specific page from session history, identified by its
+    /// relative position to the current page (with the current page being, of course, relative
+    /// index 0).
+    ///
+    /// [(JavaScript docs)](https://developer.mozilla.org/en-US/docs/Web/API/History_API#Traveling_through_history)
+    pub fn go(&self, offset: i32) {
+        js! { @(no_return)
+            @{self}.go(@{offset})
```
koute: Missing semicolon here.

src/webapi/history.rs
```diff
@@ -0,0 +1,91 @@
+#![cfg(feature = "serde")]
```
koute: Looks like you forgot to remove this? (:

src/webapi/history.rs
```diff
@@ -0,0 +1,91 @@
...
+    pub fn forward(&self) {
+        js! { @(no_return)
+            @{self}.forward()
+        };
+    }
+
+    /// Returns the current number of history entries.
+    ///
+    /// [(JavaScript docs)](https://developer.mozilla.org/en-US/docs/Web/API/History)
+    pub fn len(&self) -> i32 {
```
koute: `i32` -> `usize` to keep with Rust`s convention

src/webapi/history.rs
tomhoule: My habit of writing JS without semicolons is showing :) Fixed!

src/webapi/history.rs
tomhoule: Oh, indeed, my bad.

tomhoule:
`usize` does not work as a return type for `history.len()`, there is apparently a missing trait implementation. When trying to compile the example I get the following error:

```
error[E0277]: the trait bound `stdweb::private::Newtype<_, usize>: stdweb::private::JsSerializableOwned` is not satisfied
  --> src/main.rs:53:9
   |
53 |         js!(console.log("number of history entries: ", @{history_length}));
   |         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ the trait `stdweb::private::JsSerializableOwned` is not implemented for `stdweb::private::Newtype<_, usize>`
   |
             <stdweb::private::Newtype<(stdweb::webcore::serialization::FunctionTag, (A4, A5, A6, A7, A8, A9, A10, A11, A12)), F> as stdweb::private::JsSerializableOwned>
             <stdweb::private::Newtype<(stdweb::webcore::serialization::FunctionTag, (A2, A3, A4, A5, A6, A7, A8, A9, A10, A11, A12)), F> as stdweb::private::JsSerializableOwned>
             <stdweb::private::Newtype<(stdweb::webcore::serialization::FunctionTag, (A1, A2, A3, A4, A5, A6, A7, A8, A9, A10, A11, A12)), std::option::Option<F>> as stdweb::private::JsSerializableOwned>
             <stdweb::private::Newtype<(stdweb::webcore::serialization::FunctionTag, (A10, A11, A12)), std::option::Option<F>> as stdweb::private::JsSerializableOwned>
           and 39 others
   = note: required by `stdweb::private::JsSerializableOwned::into_js_owned`
   = note: this error originates in a macro outside of the current crate (in Nightly builds, run with -Z external-macro-backtrace for more info)
```

It could probably be added in `webcore/serialization.rs` if my understanding is correct. I don't see implementations for `u64` there either. wasm is 32 bits only so far if I remember correctly.

Is it something that should be implemented in this PR?

koute:
> It could probably be added in webcore/serialization.rs if my understanding is correct. I don't see implementations for u64 there either.

Hmm... well, yes, the problem with passing 64-bit numbers into the JavaScript is that the maximum number you can safely pass is 2^53 - 1, which is why neither the `u64` conversion is implemented nor `usize` (as *technically* `usize` could be 64 bit in the future).

I guess for pragmatic reasons we *could* make `usize` serializable (since it's safe on 32-bit targets), and when the 64-bit WASM target is a thing (**if** it's going to be a thing at all) add a check when serializing whenever the number is less than 2^53.

Feel free to add it! Just please put it in a separate commit from the `History` changes.

Pauan:
`History.prototype.length` (and almost all lengths in JavaScript, including `Array.prototype.length`) returns `u32`:

https://html.spec.whatwg.org/multipage/history.html#the-history-interface

So using `usize` is wrong in any case, regardless of whether wasm supports 64-bit or not.

koute:
Well, yes, but, conversely any length that JavaScript will give us will be always convertible to an `usize`; the problem is only when going the other way.

If I think about this really hard I guess you're right - `usize` in Rust is explicitly defined as to be as big as a pointer, and those sizes have absolutely nothing to do with Rust's pointer size! Even though it's more convenient to have those as `usize`s since then you can interop with Rust's `Vec` and others without any casts.

I'll merge this as-is and later I'll fix it myself along with other instances where we currently return `usize`.

@tomhoule Thanks for the PR!

@Pauan Thanks for keeping an eye out!

Pauan:
@koute Yeah, needing to cast `u32 <=> usize` sucks, but it is technically correct.

If `usize` auto-casted to / from `u32` then we could use `usize`, but that feels wrong, since it would either wrap around (bad) or truncate (also bad).

Because casting causes a loss of information, I feel like it should be manual, not automatic, and therefore the correct type is `u32`

PR #50, opened 2017-12-25
Added XMLHttpRequest support

keeslinp:
At first, I just made a http_get function on the window that would make a get request and callback with the response, but I figured I might as well just give it a little more substance and create the whole object.

I think that it would be more cost-effective to couple these together so that there is less crossing of he wasm-js boundary and you'd only need to call one function. However, I saw somewhere that this library was trying to maintain low-level access so I figured this would be more flexible for a user. In comparison to the cost of a network request, the cost of crossing the boundary should be negligible.

### DIFF
--- a/src/webapi/event.rs
+++ b/src/webapi/event.rs
@@ -1104,6 +1104,23 @@ reference_boilerplate! {
     convertible to ProgressRelatedEvent
 }
 
+/// The readystatechange event is fired when the readyState attribute of a document has changed.
+///
+/// [(JavaScript docs)](https://developer.mozilla.org/en-US/docs/Web/Events/readystatechange)
+pub struct ReadyStateChange( Reference );
+
+reference_boilerplate! {
+    ReadyStateChange,
+    instanceof Event
+    convertible to Event
+}
+
+impl IEvent for ReadyStateChange {}
+
+impl ConcreteEvent for ReadyStateChange {
+    const EVENT_TYPE: &'static str = "readystatechange";
+}
+
 #[cfg(all(test, feature = "web_test"))]
 mod tests {
     use super::*;
@@ -1338,5 +1355,13 @@ mod tests {
             return new ProgressEvent( @{LoadEndEvent::EVENT_TYPE} );
         ).try_into().unwrap();
         assert_eq!( event.event_type(), LoadEndEvent::EVENT_TYPE );
+    }
+
+    #[test]
+    fn test_ready_state_change_event() {
+        let event: ReadyStateChange = js!(
+            return new Event( @{ReadyStateChange::EVENT_TYPE} );
+        ).try_into().unwrap();
+        assert_eq!( event.event_type(), ReadyStateChange::EVENT_TYPE);
     }
 }
--- a/src/webapi/xml_http_request.rs
+++ b/src/webapi/xml_http_request.rs
@@ -1,0 +1,132 @@
+use webapi::event_target::{IEventTarget, EventTarget};
+use webcore::unsafe_typed_array::UnsafeTypedArray;
+use webcore::value::{
+    Reference,
+    Value,
+};
+use webcore::try_from::TryInto;
+
+/// Use XMLHttpRequest (XHR) objects to interact with servers.
+/// You can retrieve data from a URL without having to do a full page refresh.
+/// This enables a Web page to update just part of a page without disrupting
+/// what the user is doing. XMLHttpRequest is used heavily in Ajax programming.
+/// 
+/// [(JavaScript docs)](https://developer.mozilla.org/en-US/docs/Web/API/XMLHttpRequest)
+pub struct XMLHttpRequest( Reference );
+
+reference_boilerplate! {
+    XMLHttpRequest,
+    instanceof XMLHttpRequest
+    convertible to EventTarget
+}
+
+/// An enum indicating the state of the `XMLHttpRequest`.
+///
+/// [(JavaScript docs)](https://developer.mozilla.org/en-US/docs/Web/API/XMLHttpRequest/readyState)
+#[derive(Copy, Clone, PartialEq, Eq, Debug)]
+pub enum ReadyState {
+    /// Client has been created. [open()](struct.XMLHttpRequest.html#method.open) not called yet.
+    Unsent,
+    /// [open()](struct.XMLHttpRequest.html#method.open) has been called.
+    Opened,
+    /// [send()](struct.XMLHttpRequest.html#method.send) has been called, and headers and [status()](struct.XMLHttpRequest.html#method.status) are available.
+    HeadersReceived,
+    /// Downloading; [reponse_text()](struct.XMLHttpRequest.html#method.reponse_text) holds partial data.
+    Loading,
+    /// The operation is complete.
+    Done,
+}
+
+impl IEventTarget for XMLHttpRequest {}
+
+
+impl XMLHttpRequest {
+    /// Creates new `XMLHttpRequest`.
+    pub fn new() -> XMLHttpRequest {
+        js!( return new XMLHttpRequest(); ).try_into().unwrap()
+    }
+
+    /// Returns the current state of the request as a [ReadyState](enum.ReadyState.html). 
+    ///
+    /// [(JavaScript docs)](https://developer.mozilla.org/en-US/docs/Web/API/XMLHttpRequest/readyState)
+    pub fn ready_state(&self) -> ReadyState {
+        use self::ReadyState::*;
+        let state: u16 = js!( return @{self}.readyState; ).try_into().unwrap();
+        match state {
+            0 => Unsent,
+            1 => Opened,
+            2 => HeadersReceived,
+            3 => Loading,
+            4 => Done,
+            _ => unreachable!( "Unexpected value of XMLHttpRequest::readyState: {}", state )
+        }
+    }
+
+    /// Returns a string that contains the response to the request as text, or None
+    /// if the request was unsuccessful or has not yet been sent.
+    ///
+    ///[(JavaScript docs)](https://developer.mozilla.org/en-US/docs/Web/API/XMLHttpRequest/responseText)
+    pub fn response_text(&self) -> Option<String> {
+        let response = js!(return @{self}.responseText;);
+        match response {
+            Value::Null => None,
+            Value::String(resp) => Some(resp),
+            _ => unreachable!(),
+        }
+    }
+
+    /// Returns an unsigned short with the status of the response of the request.
+    ///
+    /// [(JavaScript docs)](https://developer.mozilla.org/en-US/docs/Web/API/XMLHttpRequest/status)
+    pub fn status(&self) -> u16 {
+        js!(return @{self}.status;).try_into().unwrap()
+    }
+
+    /// Open connection with given method (ie GET or POST), and the url to hit
+    ///
+    /// [(JavaScript docs)](https://developer.mozilla.org/en-US/docs/Web/API/XMLHttpRequest/open)
+    pub fn open(&self, method: &str, url: &str) {
+        js! {
+            @{self}.open(@{method}, @{url}, true);
+        };
+    }
+
+    /// Send request on an open connection with no data
+    ///
+    /// [(JavaScript docs)](https://developer.mozilla.org/en-US/docs/Web/API/XMLHttpRequest/send)
+    pub fn send(&self) {
+        js! {
+            @{self}.send();
+        };
+    }
+
+    /// Send request on an open connection with string body
+    ///
+    /// [(JavaScript docs)](https://developer.mozilla.org/en-US/docs/Web/API/XMLHttpRequest/send)
+    pub fn send_with_string(&self, body: &str) {
+        js! {
+            @{self}.send(@{body});
+        };
+    }
+
+    /// Send request on an open connection with a byte array body
+    ///
+    /// [(JavaScript docs)](https://developer.mozilla.org/en-US/docs/Web/API/XMLHttpRequest/send)
+    pub fn send_with_bytes(&self, body: &[u8]) {
+        js! {
+            @{self}.send(@{UnsafeTypedArray(body)});
+        };
+    }
+
+    /// Aborts the request if it has already been sent.
+    /// When a request is aborted, its [ready_state](struct.XMLHttpRequest.html#method.ready_state) is changed to [Done](enum.ReadyState.html#variant.Done)
+    /// and the [status](struct.XMLHttpRequest.html#method.status) code is set to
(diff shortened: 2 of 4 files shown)


src/webapi/xml_http_request.rs
```diff
@@ -0,0 +1,69 @@
...
+        let reference = self.reference.as_ref();
+        js! {
+            @{reference}.open(@{method}, @{url}, true);
+        };
+    }
+
+    /// Send request on an open connection with the Value that is being sent
+    /// If using GET it would be Null
+    /// [(JavaScript docs)](https://developer.mozilla.org/en-US/docs/Web/API/XMLHttpRequest/send)
+    pub fn send(&self, value: Value) {
```
koute: Looking at the JS interface I think it would be better to provide multiple versions of this function instead of having it take a catch-all `Value`. Perhaps something like this?

```
fn send( &self );
fn send_with_string( &self, request_body: &str );
fn send_with_bytes( &self, request_body: &[u8] );
```

src/webapi/xml_http_request.rs
```diff
@@ -0,0 +1,69 @@
...
+    Value,
+};
+use webcore::try_from::TryInto;
+
+/// XMLHttpRequest object to make http requests
+/// [(JavaScript docs)](https://developer.mozilla.org/en-US/docs/Web/API/XMLHttpRequest)
+#[derive(Debug)]
+pub struct XMLHttpRequest {
+    reference: Reference,
+}
```
koute: Please make this the same as other `Reference` wrappers. (e.g. look at `src/webapi/file.rs` for a minimal example) In this case the `XMLHttpRequest` derives from `IEventTarget`.

src/webapi/xml_http_request.rs
```diff
@@ -0,0 +1,69 @@
...
+        }.try_into().unwrap();
+        XMLHttpRequest {
+            reference,
+        }
+    }
+    /// Sets the callback to call once a request completes
+    /// First argument is the status code, the second argument is the response text
+    /// Makes use of onreadystatechange:
+    /// [(JavaScript docs)](https://developer.mozilla.org/en-US/docs/Web/API/XMLHttpRequest/onreadystatechange)
+    pub fn set_callback<F:FnMut(u32, String) + 'static>(&self, callback: F) {
```
koute: Instead of ad hoc methods like this I'd prefer to do this through `IEventTarget` machinery we already have. (So instead of using this function the user would use `add_event_listener` from the `IEventTarget` interface.) In which case it would also be nice to have an usage example in the doc comments.

src/webapi/xml_http_request.rs
```diff
@@ -0,0 +1,69 @@
...
+            var xmlHttp = @{reference};
+            xmlHttp.onreadystatechange = function() {
+                if (xmlHttp.readyState == 4) {
+                    callback(xmlHttp.status, xmlHttp.responseText);
+                }
+            }
+        }
+    }
+
+    /// Open connection with given method (ie GET or POST), and the url to hit
```
koute: For every method you expose please take the docs from MDN and lightly edit them as needed, just as in your other PR. (:

src/webapi/xml_http_request.rs
```diff
@@ -0,0 +1,69 @@
...
+    /// First argument is the status code, the second argument is the response text
+    /// Makes use of onreadystatechange:
+    /// [(JavaScript docs)](https://developer.mozilla.org/en-US/docs/Web/API/XMLHttpRequest/onreadystatechange)
+    pub fn set_callback<F:FnMut(u32, String) + 'static>(&self, callback: F) {
+        let reference = self.reference.as_ref();
+        js! {
+            var callback = @{callback};
+            var xmlHttp = @{reference};
+            xmlHttp.onreadystatechange = function() {
+                if (xmlHttp.readyState == 4) {
```
koute: The `readyState` should be exposed just like in (for example) `file_reader.rs`.

gabisurita:
@keeslinp do you plan to address the changes pointed by @koute? I'm specially interested on this feature and would be happy to help if you're ok with that.

keeslinp:
Yeah, I'm planning on doing that tonight or tomorrow. My schedule is a
little weird at the moment cause of vacation. I've got some of it done but
not all of it yet

src/webapi/xml_http_request.rs
CryZe: Should this be FnMut? I'm not entirely sure, but if the callback can only resolve once, it should probably be FnOnce

keeslinp:
It can resolve more than once. The callback isn't directly connected with a
sent request and so you could send more than once with the same callback.
When I change it to use the event listener machinery I'll do it however it
is done there.

keeslinp:
Should be all done. It now much more reflects the javascript api. Honestly, I don't know if that's necessarily a good thing because it feels a little clunky but I can see how that's less surprising that having to learn a specific api for the callback.

src/webapi/event.rs
```diff
@@ -1066,6 +1066,27 @@ reference_boilerplate! {
     convertible to ProgressRelatedEvent
 }
 
+/// An EventHandler that is called whenever the readyState attribute changes.
+/// The callback is called from the user interface thread.
+/// The XMLHttpRequest.onreadystatechange property contains the event handler to be
+/// called when the readystatechange event is fired,
+/// that is every time the readyState property of the XMLHttpRequest changes.
+///
+/// [(JavaScript docs)](https://developer.mozilla.org/en-US/docs/Web/API/XMLHttpRequest/onreadystatechange)
```
koute: This should link to this and have the docs copied from this page:

https://developer.mozilla.org/en-US/docs/Web/Events/readystatechange

src/webapi/xml_http_request.rs
```diff
@@ -0,0 +1,129 @@
+use webapi::event_target::{IEventTarget, EventTarget};
+use webcore::value::{
+    Reference,
+    Value,
+};
+use webcore::try_from::TryInto;
+
+/// XMLHttpRequest object to make http requests
+/// [(JavaScript docs)](https://developer.mozilla.org/en-US/docs/Web/API/XMLHttpRequest)
```
koute: Please copy the first paragraph from MDN, and also put an extra newline before `[(JavaScript docs)]`

src/webapi/xml_http_request.rs
```diff
@@ -0,0 +1,129 @@
...
+pub struct XMLHttpRequest( Reference );
+
+reference_boilerplate! {
+    XMLHttpRequest,
+    instanceof XMLHttpRequest
+    convertible to EventTarget
+}
+
+/// A number indicating the state of the `XMLHttpRequest`.
```
koute: In Rust it's not a number. (:

src/webapi/xml_http_request.rs
```diff
@@ -0,0 +1,129 @@
...
+/// [(JavaScript docs)](https://developer.mozilla.org/en-US/docs/Web/API/XMLHttpRequest/readyState)
+#[derive(Copy, Clone, PartialEq, Eq, Debug)]
+pub enum ReadyState {
+    /// Client has been created. open() not called yet.
+    Unsent,
+    /// open() has been called.
+    Opened,
+    /// send() has been called, and headers and status are available.
+    HeadersReceived,
+    /// Downloading; responseText holds partial data.
```
koute: Use Rust names here. (`responseText` -> `response_text()`)

Also it would be nice to link to those methods you're referring to.

src/webapi/xml_http_request.rs
```diff
@@ -0,0 +1,129 @@
...
+    /// Returns a DOMString that contains the response to the request as text, or None
+    /// if the request was unsuccessful or has not yet been sent.
+    ///
+    ///[(JavaScript docs)](https://developer.mozilla.org/en-US/docs/Web/API/XMLHttpRequest/responseText)
+    pub fn response_text(&self) -> Option<String> {
+        let response = js!(return @{self}.responseText;);
+        match response {
+            Value::Null => None,
+            Value::String(resp) => Some(resp),
+            _ => None,
```
koute: This should be an `unreachable!`.

src/webapi/xml_http_request.rs
```diff
@@ -0,0 +1,129 @@
...
+            0 => Unsent,
+            1 => Opened,
+            2 => HeadersReceived,
+            3 => Loading,
+            4 => Done,
+            _ => unreachable!( "Unexpected value of XMLHttpRequest::readyState: {}", state )
+        }
+    }
+
+    /// Returns a DOMString that contains the response to the request as text, or None
```
koute: `DOMString` -> `string`

src/webapi/xml_http_request.rs
```diff
@@ -0,0 +1,129 @@
...
+    pub fn response_text(&self) -> Option<String> {
+        let response = js!(return @{self}.responseText;);
+        match response {
+            Value::Null => None,
+            Value::String(resp) => Some(resp),
+            _ => None,
+        }
+    }
+
+    /// Returns an unsigned short with the status of the response of the request.
```
koute: More detailed docs would be nice. (First paragraph from MDN looks nice, without the remark that it's an `unsigned short` since we can already see that from the prototype.)

src/webapi/xml_http_request.rs
```diff
@@ -0,0 +1,129 @@
...
+    ///
+    /// [(JavaScript docs)](https://developer.mozilla.org/en-US/docs/Web/API/XMLHttpRequest/status)
+    pub fn status(&self) -> u16 {
+        js!(return @{self}.status;).try_into().unwrap()
+    }
+
+    /// Open connection with given method (ie GET or POST), and the url to hit
+    /// Async is set for now because it's more ergonimic in rust and it is good form in javascript
```
koute: "Async is set for now because it's more ergonimic in rust and it is good form in javascript" -> this seems unnecessary. You can document that it's asynchronous, but it's not necessary to explain it like this.

src/webapi/xml_http_request.rs
```diff
@@ -0,0 +1,129 @@
...
+    /// Send request on an open connection with a byte array body
+    ///
+    /// [(JavaScript docs)](https://developer.mozilla.org/en-US/docs/Web/API/XMLHttpRequest/send)
+    pub fn send_with_bytes(&self, body: &[u8]) {
+        js! {
+            @{self}.send(@{body});
+        };
+    }
+
+    /// The abort() method aborts the request if it has already been sent.
```
koute: You don't have to repeat the method name, e.g. `The abort() method` since it should be already obvious to the reader. (:

src/webapi/xml_http_request.rs
```diff
@@ -0,0 +1,129 @@
...
+    /// When a request is aborted, its readyState is changed to Done
```
koute: `readyState` -> link to the ready_state method, `status code` -> link to the status method.

src/webapi/xml_http_request.rs
```diff
@@ -0,0 +1,129 @@
...
+    Loading,
+    /// The operation is complete.
+    Done,
+}
+
+impl IEventTarget for XMLHttpRequest {}
+
+
+impl XMLHttpRequest {
+    /// Create new XMLHttpRequest
```
koute: Minor grammar nitpick (this also applies to other doc comments): 'Create' -> 'Creates', and also a dot at the end and put identifiers (like `XMLHttpRequest`) inside backticks. (:

src/webapi/xml_http_request.rs
```diff
@@ -0,0 +1,129 @@
...
+            @{self}.open(@{method}, @{url}, true);
+        };
+    }
+
+    /// Send request on an open connection with no data
+    ///
+    /// [(JavaScript docs)](https://developer.mozilla.org/en-US/docs/Web/API/XMLHttpRequest/send)
+    pub fn send(&self) {
+        js! {
+            @{self}.send(null);
```
koute: No need to pass the `null` here. Just leave the parens empty.

src/webapi/xml_http_request.rs
```diff
@@ -0,0 +1,129 @@
...
+            @{self}.send(@{body});
+        };
+    }
+
+    /// Send request on an open connection with a byte array body
+    ///
+    /// [(JavaScript docs)](https://developer.mozilla.org/en-US/docs/Web/API/XMLHttpRequest/send)
+    pub fn send_with_bytes(&self, body: &[u8]) {
+        js! {
+            @{self}.send(@{body});
```
koute: Use `UnsafeTypedArray` here to make it zero cost.

keeslinp:
Haha docs are going to be the death of me :). It's a good learning
experience for me though. I'll make these updates either tonight or
tomorrow night depending how travel goes.

koute:
> Haha docs are going to be the death of me :)

Yeah, sorry for being so pedantic. (:

keeslinp:
I'm sure many future users who are trying to find their way through the
docs are going to thank you :)

keeslinp:
Should be fixed. Hopefully all the docs are in order now.

src/webapi/xml_http_request.rs
```diff
@@ -0,0 +1,134 @@
...
+    pub fn send(&self) {
+        js! {
+            @{self}.send();
+        };
+    }
+
+    /// Send request on an open connection with string body
+    ///
+    /// [(JavaScript docs)](https://developer.mozilla.org/en-US/docs/Web/API/XMLHttpRequest/send)
+    pub fn send_with_string(&self, body: String) {
```
koute: This should take `&str`; no real reason to take `String`.

src/webapi/xml_http_request.rs
```diff
@@ -0,0 +1,134 @@
...
+    Done,
+}
+
+impl IEventTarget for XMLHttpRequest {}
+
+
+impl XMLHttpRequest {
+    /// Creates new `XMLHttpRequest`.
+    pub fn new() -> XMLHttpRequest {
+        XMLHttpRequest(js!{
```
koute: Actually, you should be able to convert it like this:

```
js!( return new XMLHttpRequest(); ).try_into().unwrap()
```

keeslinp: I think I tried that originally and I couldn't get it to work. I wonder if I set something up wrong with the boilerplate macro. I'll try it again tonight, maybe I just made a silly mistake.

keeslinp: I must have messed up something silly cause it worked fine this time. Thanks.

keeslinp:
Should be all better now :). Thanks for the help!

koute:
> PS: Kudos on breaking a thousand stars, you've done a great job!

Thanks! (:

Anyhow, this mostly looks fine now so I'm going to merge it in. Thank you for the contribution!

Issue #59, opened 2017-12-30
Run web API tests on CI and provide instructions for running locally


Diggsey:
I have no idea how to run them locally and the CI script passes `--nodejs` which doesn't support them anyway!

How difficult would it be to support something like http://zombie.js.org/ or just headless chrome?

koute:
You just don't pass `--nodejs` and then `cargo-web` will try to run them in Chrome/Chromium. (You also need to temporairly comment the `#[cfg(...)]` which prevents them from running, since I haven't yet implemented support for switch on features in `cargo-web`.)

Right now it's very hacky which is why it's not done on the CI. In the future I would certainly want to switch to headless Chrome or Firefox, *then* we can start running them on CI.

As to how hard would it be to support headless Chrome? Well, quite frankly I'm not sure. We'd have to implement the [devtools protocol](https://chromedevtools.github.io/devtools-protocol/) to communicate with it, and we'd have to get precompiled binaries for it from *somewhere* - probably build them ourselves just as we do for Emscripten (this alone is a huge pain since compiling Chrome takes **ages**, and the build system is a nightmare to set up).

RangerMauve:
What about using an existing node.js headless chrome and creating WASM bindings for it?

koute:
All of the tests are being run on CI and I've added instructions on how to run them at the end of the README, so closing.

PR #64, opened 2018-01-01
Adding support for innerHTML

richardanaya:
Hi there Koute :) it's that dude from reddit the other day making web component stuff. I'm still actively working on that library right now and I came across this pretty important functionality I could use and probably others. I tried to adhere to your existing standards. Thank you again for your awesome project.


src/webapi/element.rs
```diff
@@ -42,6 +42,26 @@ pub trait IElement: IEventTarget {
             js!( return @{self.as_ref()}.querySelectorAll( @{selector} ); ).into_reference_unchecked().unwrap()
         }
     }
+    
+     /// A property which represents the inner html of a element and its descendants as a DOM string.
```
koute: There is a stray space before the `///` here. (:

src/webapi/element.rs
```diff
@@ -42,6 +42,26 @@ pub trait IElement: IEventTarget {
...
+    ///
+    /// [(JavaScript docs)](https://developer.mozilla.org/en-US/docs/Web/API/Element/innerHTML)
+    fn inner_html( &self ) -> Option< String > {
```
koute: Why the `Option`? Can it return `null` or `undefined`? If it can't then it should simply be a `String`.

src/webapi/element.rs
```diff
@@ -42,6 +42,26 @@ pub trait IElement: IEventTarget {
...
+            return @{self.as_ref()}.innerHTML;
+        ).into_string()
+    }
+
+    /// Sets the inner html of this element; calling this removes all
+    /// of node's children and replaces them with html elements
+    /// of the given DOM string.
+    ///
+    /// [(JavaScript docs)](https://developer.mozilla.org/en-US/docs/Web/API/Element/innerHTML)
+    fn set_inner_html( &self, text: &str ) {
```
koute: According to specs [this can return an exception](https://w3c.github.io/DOM-Parsing/#extensions-to-the-element-interface) when the text is not well formed HTML **and** the document is XML.

For example:

```
var doc = document.implementation.createDocument("", "", null);
var e = doc.createElement("div");
e.innerHTML = "<foo"
```

will throw:

```
SyntaxError: An invalid or illegal string was specified
XML Parsing Error: not well-formed
Location: 
Line Number 1, Column 10:
```

So we should catch the exception and turn it into a `Result`. (e.g. see `INode::remove_child`)

Please also add a test for this; you can use the JS snippet I've pasted here. (You can take a look at the issue #59 if you don't know how to run Web API tests.)

richardanaya:
Hmm, I tried taking this as far as I could, i'm getting a strange error.

 RUST_BACKTRACE=1 cargo +nightly web test --nodejs --target-webasm
warning: debug builds on the wasm-unknown-unknown are currently totally broken
         forcing a release build
    Finished release [optimized] target(s) in 0.0 secs
thread 'main' panicked at 'internal error: nothing changed so I don't know which artifact corresponds to this build; change something and try again!', /Users/richard/.cargo/registry/src/github.com-1ecc6299db9ec823/cargo-web-0.4.1/src/cmd_test.rs:203:16
stack backtrace:
   0: std::sys::imp::backtrace::tracing::imp::unwind_backtrace
   1: std::sys_common::backtrace::_print
   2: std::panicking::default_hook::{{closure}}
   3: std::panicking::default_hook
   4: std::panicking::rust_panic_with_hook
   5: std::panicking::begin_panic
   6: cargo_web::cmd_test::command_test
   7: cargo_web::main
   8: __rust_maybe_catch_panic
   9: std::rt::lang_start

koute:
If you grab the newest `cargo-web` from `master` you can now run all of the `stdweb`'s tests like this:

```
cargo web test --features web_test
```

richardanaya:
Thanks, this got me a bit further, it appears certain tests are failing because the javascript can't emit certain events:

```bash
cargo +nightly web test --features web_test --nodejs --target-webasm
warning: debug builds on the wasm-unknown-unknown are currently totally broken
         forcing a release build
   Compiling num-traits v0.1.41
   Compiling dtoa v0.4.2
   Compiling serde v1.0.27
   Compiling itoa v0.3.4
   Compiling serde_json v1.0.9
   Compiling stdweb v0.3.0 (file:///Users/richard/Source/stdweb)
    Finished release [optimized] target(s) in 83.24 secs
    Garbage collecting "stdweb-4311a7bb42cfc854.wasm"...
    Processing "stdweb-4311a7bb42cfc854.wasm"...
    Finished processing of "stdweb-4311a7bb42cfc854.wasm"!
/Users/richard/Source/stdweb/target/wasm32-unknown-unknown/release/stdweb-4311a7bb42cfc854.js:383
                $1 = Module.STDWEB.to_js($1);Module.STDWEB.from_js($0, (function(){return new FocusEvent (($1));})());
                                                                                   ^

ReferenceError: FocusEvent is not defined
    at /Users/richard/Source/stdweb/target/wasm32-unknown-unknown/release/stdweb-4311a7bb42cfc854.js:383:84
    at __extjs_eb63655e16b9202baa2d895a8c70bce89feb4db6 (/Users/richard/Source/stdweb/target/wasm32-unknown-unknown/release/stdweb-4311a7bb42cfc854.js:383:115)
    at stdweb::webapi::event::tests::test_blur_event::hf5695b5b95946c61 (wasm-function[857]:189)
    at test::__rust_begin_short_backtrace::h6a7537760bdc56cb (wasm-function[395]:3)
    at <F as alloc::boxed::FnBox<A>>::call_box::h4d842f29ff41d9f7 (wasm-function[394]:6)
    at test::run_test::run_test_inner::hd684a4e0d7dbedfc (wasm-function[386]:880)
    at test::run_test::h574e0b1f824f5c09 (wasm-function[379]:1592)
    at test::run_tests_console::h09c6fade20c14962 (wasm-function[375]:7319)
    at test::test_main_static::h0fb58fac19f38ab3 (wasm-function[401]:1518)
    at stdweb::__test::main::hfc341bbbf1e32531 (wasm-function[288]:8)


node -v
v9.3.0
```

koute:
@richardanaya That's expected. You can't run `web_test` tests under Node.js.

What you need to do is to run it like this:

```
cargo web test --features web_test
```

Basically, currently to run those tests you need to use an Emscripten-based backend and you need to run the tests in a web browser (which is the default if you don't specify `--nodejs`)

src/webapi/element.rs
```diff
@@ -42,6 +59,39 @@ pub trait IElement: IEventTarget {
...
+    /// innerHTML an not well formed XML, this will throw an exception.
+    ///
+    /// [(JavaScript docs)](https://developer.mozilla.org/en-US/docs/Web/API/Element/innerHTML)
+    fn set_inner_html( &self, text: &str ) -> Result< (), InvalidStateError > {
+        let status = js! {
+            try {
+                @{self.as_ref()}.innerHTML = @{text};
+                return true;
+            } catch( exception ) {
+                throw exception;
```
koute: You still need to check the type of the exception. (:

richardanaya: Thanks! Hmm, what do you think I should do if it's not the exception type i'm expecting?

koute: What are you expecting and what are you getting?

richardanaya: Oh! Actually, I see now what's happening in another bit of code:

This is from node.rs
if( exception instanceof NotFoundError ) {
                    return false;
} else {
        throw exception;
}

You return false and throw a correct Rust error if its what's expected, and throw a JS exception if its not what is expected (thus stopping the rust call back)

richardanaya:
After reading the https://developer.mozilla.org/en-US/docs/Web/API/DOMError page, it sounds like there's a newer https://developer.mozilla.org/en-US/docs/Web/API/DOMException error. I chose to standardize around this and put it in its own file error.rs as a place we can grow support for its various scenerios its meant to handle.

I also added more fine grain checking in the set_inner_html as per your suggestion.

richardanaya:
Making progress on the tests, I seem to be able to run the emscripten version, I'm getting this error:
```bash
error: you need to have either Chromium or Chrome installed and in your PATH to run the tests!
```
I  tried:

echo PATH=$PATH:/Applications/Google\ Chrome.app/Contents/MacOS/

koute:
Can you launch chrome/chromium from a terminal manually?

Issue #28, opened 2017-10-15
Establish a pattern for JS exception handling


Herschel:
Lots of web APIs throw exceptions, so we should establish a pattern for passing these exceptions back to Rust in a nice way.

For example, `[Node::remove_child](https://github.com/koute/stdweb/blob/1a1f8434456cf1a9650daceb8f3791c8be0f11a6/src/webapi/node.rs#L64-L84)` uses a pattern of returning a bool when an exception happens, which is then converted into a Rust `NotFoundErr` as appropriate.

I could picture this pattern being wrapped into some kind of `js_try!` macro. Could the exception itself be serialized to Rust if we implemented the JS [Error](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Error) type? That way, instead of a bool, the JS could return the value itself or the exception, which could be converted into the proper Rust type.

What should the error type be? For example, `Node.removeChild` can return [other exception types](https://www.w3.org/TR/DOM-Level-3-Core/core.html#ID-1734834066). Should we make a more global `webapi::Error` enum that represents every possible exception that JS can throw?

koute:
My gut feeling is that having functions return an unified `Error` type isn't a great idea; a lot of the errors thrown from the JS world are unrelated to each other, so if we unify every error under a single enum we'll lose the very nice property of being able to tell which exact errors a given function can return based only on its return value.

What we could do is to have them return specific types (or single use enums just for that function?), and have an unifying enum which has `Into` and `TryFrom` implemented for every concrete error type. Then if the user wants to unify them he easily can.

Another issue (which was iterated in https://github.com/koute/stdweb/issues/21) is that we *probably* would like the errors to be a little more that a simple struct/enum. JS errors have stack traces, and I would love to fully support that on the Rust side. (This, however, would require some more investigation and work, e.g. for example we'd definitely need some way to decode backtraces to be human readable.)

Anyway, so maybe something like this?

```
// Corresponds to the JS' Error interface.
trait IError: FromReference {
}

// Corresponds to an instance of a JS object implementing the Error interface.
struct Error( Reference );
    
impl IError for Error {}
            
reference_boilerplate! {
    Error,
    instanceof Error
}

// Actual error types.
struct TypeError { backtrace: Backtrace, ... };

// An unifying type.
enum JsError {
    // All of the standard errors.
    TypeError( TypeError ),
    // ....
    // An error that we don't know of, so we store the original reference.
    OtherError( Error )
}
```

[...]

PR #66, opened 2018-01-03
Implement set and get headers for XMLHttpRequest

gabisurita:
Adds support for adding and checking HTTP request/response headers with XMLHTTPRequest.

Related to https://github.com/koute/stdweb/pull/50

### DIFF
--- a/src/webapi/xml_http_request.rs
+++ b/src/webapi/xml_http_request.rs
@@ -86,7 +86,7 @@ impl XMLHttpRequest {
     ///
     /// [(JavaScript docs)](https://developer.mozilla.org/en-US/docs/Web/API/XMLHttpRequest/open)
     pub fn open(&self, method: &str, url: &str) {
-        js! {
+        js! { @(no_return)
             @{self}.open(@{method}, @{url}, true);
         };
     }
@@ -113,7 +113,7 @@ impl XMLHttpRequest {
     pub fn set_request_header(&self, header: &str, value: &str) {
         // TODO: Handle InvalidStateError and SyntaxError exceptions.
         // https://xhr.spec.whatwg.org/#the-setrequestheader()-method
-        js! {
+        js! { @(no_return)
             @{self}.setRequestHeader(@{header}, @{value});
         };
     }
@@ -122,7 +122,7 @@ impl XMLHttpRequest {
     ///
     /// [(JavaScript docs)](https://developer.mozilla.org/en-US/docs/Web/API/XMLHttpRequest/send)
     pub fn send(&self) {
-        js! {
+        js! { @(no_return)
             @{self}.send();
         };
     }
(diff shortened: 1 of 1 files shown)


src/webapi/xml_http_request.rs
```diff
@@ -91,6 +91,31 @@ impl XMLHttpRequest {
         };
     }
 
+    /// Returns the string containing the text of the specified header. If there
+    /// are multiple response headers with the same name, then their values are
+    /// returned as a single concatenated string.
+    ///
+    /// [(JavaScript docs)](https://developer.mozilla.org/en-US/docs/Web/API/XMLHttpRequest/getRequestHeader)
```
koute: Hmm... that link doesn't lead to anything, and I'm not seeing a method called `getRequestHeader` on [MDN](https://developer.mozilla.org/en-US/docs/Web/API/XMLHttpRequest) nor [in the specs](https://xhr.spec.whatwg.org/).

Are you sure this works? If so where does it come from? O_O

src/webapi/xml_http_request.rs
```diff
@@ -91,6 +91,31 @@ impl XMLHttpRequest {
...
+            _ => unreachable!(),
+        }
+    }
+
+    /// Sets the value of an HTTP request header. Must be called after `open()`,
+    /// but before `send()`. If this method is called several times with the same
+    /// header, the values are merged into one single request header.
+    ///
+    /// [(JavaScript docs)](https://developer.mozilla.org/en-US/docs/Web/API/XMLHttpRequest/setRequestHeader)
+    pub fn set_request_header(&self, header: &str, value: &str) {
```
koute: Looking at the [specs](https://xhr.spec.whatwg.org/#the-setrequestheader()-method) this method can throw.

I guess we can leave it now as-is; one of the exceptions it can throw is the `InvalidStateError`, which coincidentially is being currently added in https://github.com/koute/stdweb/pull/64. (I guess we should do something with https://github.com/koute/stdweb/issues/28 rather sooner than later and have a more principled way of handling throwing methods.)

Please leave a `TODO` that this function throws and that we need to properly handle that.

src/webapi/xml_http_request.rs
```diff
@@ -91,6 +91,31 @@ impl XMLHttpRequest {
...
+        js! {
```
koute: I forgot this in the last review, but if you have a `js!` macro invocation from which you don't need to return anything then please add `@(no_return)` at the start. It will be marginally more efficient and generate less code.

src/webapi/xml_http_request.rs
gabisurita: Ops. Typo there. Should be response header there.

koute:
This looks fine now! Thanks a lot!

PR #68, opened 2018-01-03
Fix µtypo on the examples docs

gabisurita:


koute:
Ah right, I noticed it myself this morning and then... completely forgot about it. -_-

Thanks!

PR #69, opened 2018-01-05
Add TextAreaElement with value getter/setter

DenisKolodin:
I want to use textareas like inputs, but now I have to use `js!` workaround only to get or set value to a textarea element.
Also this PR helps me to implement https://github.com/DenisKolodin/yew/pull/82

### DIFF
--- a/src/webapi/html_elements/textarea.rs
+++ b/src/webapi/html_elements/textarea.rs
@@ -1,0 +1,43 @@
+use webcore::value::Reference;
+use webcore::try_from::TryInto;
+use webapi::event_target::{IEventTarget, EventTarget};
+use webapi::node::{INode, Node};
+use webapi::element::{IElement, Element};
+use webapi::html_element::{IHtmlElement, HtmlElement};
+
+/// The HTML <textarea> element represents a multi-line plain-text editing control.
+///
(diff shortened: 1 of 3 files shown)


src/webapi/html_elements/textarea.rs
```diff
@@ -0,0 +1,42 @@
...
+    convertible to EventTarget
+    convertible to Node
+    convertible to Element
+    convertible to HtmlElement
+}
+
+impl TextAreaElement {
+    /// The value of the control.
+    #[inline]
+    pub fn value( &self ) -> Value {
```
koute: According to the [specs](https://html.spec.whatwg.org/multipage/form-elements.html#the-textarea-element) this should return a `String`.

src/webapi/html_elements/textarea.rs
```diff
@@ -0,0 +1,42 @@
...
+        js! (
+            return @{self}.value;
+        )
+    }
+
+    /// Sets the value of the control.
+    #[inline]
+    pub fn set_value< T: Into< Value > >( &self, value: T ) {
```
koute: Since the field is a string this should also take an `&str`.

DenisKolodin:
@koute You're absolutely right, it really must be a string. I've fixed it.

koute:
LGTM, thanks!

PR #74, opened 2018-01-12
Added mouseup, mousedown, and mousemove events.

iirelu:
This *should* work, but I haven't tested it myself yet. It's definitely something I need, though.


koute:
LGTM! Thanks!

Note: you still need to export it to be able to use it. (:

PR #79, opened 2018-01-14
Make `insert_before` reference node optional

### LINKED ISSUE #77: `Node::insert_before` argument should be optional
The `referenceNode` is allowed to be `null`, so the rust API should take an `Option<&Node>`. Might be worth checking other related methods too.

Diggsey:
Fixes #77


DenisKolodin:
It seems this PR will help me to simplify VDom here: https://github.com/DenisKolodin/yew/blob/master/src/virtual_dom/vtext.rs#L84

Pauan:
@DenisKolodin Yup, but there's an alternative proposal: https://github.com/koute/stdweb/issues/77#issuecomment-357535191

We haven't decided yet which proposal we want to do.

DenisKolodin:
@Pauan I expect it will work like [`Node.insertBefore()`](https://developer.mozilla.org/en-US/docs/Web/API/Node/insertBefore):

> The `Node.insertBefore()` method inserts a node before the reference node as a child of a specified parent node. If the reference node is null, the specified node is added to the end of the list of children of the specified parent node.

Other alternatives are suitable too. I need a method to insert a node before another or append it to the tail.

Pauan:
@DenisKolodin Yes, we know this is a pain point, we just haven't decided on *which* alternative to use for the Rust API.

Diggsey:
Closing - no activity

PR #80, opened 2018-01-14
Add Canvas API

Diggsey:


src/webapi/event.rs
```diff
@@ -332,6 +332,24 @@ reference_boilerplate! {
     convertible to UiEvent
 }
 
+/// The resize event is fired when the document view has been resized.
+///
+/// MDN incorrectly documents this as a UIEvent, but in browsers it is actually
+/// just an Event.
+/// [(JavaScript docs)](https://developer.mozilla.org/en-US/docs/Web/Events/resize)
```
koute: Did you check what the specs say?

src/webapi/html_elements/canvas.rs
```diff
@@ -0,0 +1,108 @@
...
+    }
+
+    /// Returns a drawing context on the canvas, or None if the context identifier is not supported.
+    ///
+    /// [(JavaScript docs)](https://developer.mozilla.org/en-US/docs/Web/API/HTMLCanvasElement/getContext)
+    pub fn get_context<T: RenderingContext>( &self ) -> Option<T> {
+        js! (
+            return @{self}.getContext(@{T::CONTEXT_TYPE});
+        ).try_into().ok().and_then(|r| T::from_reference(r))
+    }
```
koute: I think this might not be flexible enough. E.g. for WebGL at some point in time in various browsers you could use one of the following names for what effectively was the same kind of a context: `"webgl", "experimental-webgl", "webkit-3d", "moz-webgl"`, so this mechanism should support specifying a list. (Or perhaps we should just have the `RenderingContext` export a `get_context` method and just delegate calling `getContext` to it completely?)

src/webapi/html_elements/canvas.rs
```diff
@@ -0,0 +1,108 @@
...
+        js! (
+            return @{self}.toDataUrl(@{mime_type}, @{quality});
+        ).try_into().unwrap()
+    }
+
+    /// Creates a Blob object representing the image contained in the canvas; this file may be
+    /// cached on the disk or stored in memory at the discretion of the user agent.
+    ///
+    /// [(JavaScript docs)](https://developer.mozilla.org/en-US/docs/Web/API/HTMLCanvasElement/toBlob)
+    pub fn to_blob<F: FnMut(Blob) + 'static>( &self, f: F, mime_type: Option<&str>, quality: Option<f64> ) {
```
koute: `FnOnce` (There is a wrapper called `Once` which allows passing `FnOnce` functions to `js!`)

koute: Also, semi related question - do you know why the original JS API takes a callback here instead of returning a `Blob`? Is it because the `Blob` is done asynchronously?

src/webcore/macros.rs
```diff
@@ -654,6 +661,7 @@ macro_rules! reference_boilerplate {
     };
 }
 
+#[macro_export]
```
koute: Hmm... I assume you're exporting these to allow the context to be implemented externally? Hmmm.... Could we keep those private? I would actually be fine having the `RenderingContext`s implementations inside `stdweb` for now. (They're JS APIs after all!)

src/webapi/html_elements/canvas.rs
```diff
@@ -0,0 +1,108 @@
+use webcore::value::{Reference, FromReference};
+use webcore::try_from::TryInto;
+use webapi::event_target::{IEventTarget, EventTarget};
+use webapi::node::{INode, Node};
+use webapi::element::{IElement, Element};
+use webapi::html_element::{IHtmlElement, HtmlElement};
+use webapi::blob::Blob;
+
+/// The HTML image element is used to manipulate the layout and presentation of
+/// `<img>` elements.
```
koute: This doc comment is wrong. (:

src/webapi/event.rs
Diggsey: It was changed in the specs recently from a UIEvent to an Event to match expectations/browser implementations. Either way, this errors in all browsers if it's specified as a UIEvent rather than an Event.

src/webapi/html_elements/canvas.rs
Diggsey: OK, I will give `RenderingContext` a `from_canvas` method or something...

src/webapi/html_elements/canvas.rs
Diggsey: Yes.

src/webcore/macros.rs
Diggsey: Hm, OK, but I think these macros should be exposed anyway - it's not uncommon to want to wrap new types of references.

koute: I'm not against exposing something like this macro in the future, but I'd rather keep it private for now. Unlike the `js!` macro this macro wasn't designed to be used outside of `stdweb`, so I will likely want to export something *slightly* different to the outside.

Diggsey:
@koute is this good to go?

koute:
Right, sorry!

It's almost ready to go! Let me just put a few more comments.

examples/canvas/src/main.rs
```diff
@@ -0,0 +1,70 @@
...
+    ( ($( $x:ident ),*) $y:expr ) => {
+        {
+            $(let $x = $x.clone();)*
+            $y
+        }
+    };
+}
+
+#[derive(Clone, Debug)]
+struct CanvasRenderingContext2D(Reference);
```
koute: As I wrote in one of the comments, I would be fine with having the `RenderingContext`s implementations inside of `stdweb`, and I think I'd actually prefer that right now, so please move this from the example to inside the `stdweb`.

Also, please rename the `2D` to `2d` to keep with the `PascalCase` naming convention we've decided the other day to stick to.

src/lib.rs
```diff
@@ -188,12 +190,16 @@ pub mod web {
             KeyupEvent,
             ClickEvent,
             DoubleClickEvent,
+            MouseDownEvent,
+            MouseUpEvent,
+            MouseMoveEvent,
```
koute: These were already exported in another PR, so please get rid of them.

src/webapi/html_elements/canvas.rs
```diff
@@ -0,0 +1,107 @@
...
+    convertible to EventTarget
+    convertible to Node
+    convertible to Element
+    convertible to HtmlElement
+}
+
+/// Trait implemented by rendering contexts which can be obtained from a canvas.
+pub trait RenderingContext {
+    /// Name which identifies this kind of rendering context.
+    fn from_canvas(canvas: &CanvasElement) -> Option<Self> where Self: Sized;
```
koute: I think it would make more sense for this to return `Result<Self, Self::Error>`.

examples/canvas/src/main.rs
Diggsey: Hm, OK I will move it in - I'm not planning on expanding on the set of implemented methods though.

I do think these would work better as a separate crate: have you thought about splitting up stdweb?

eg.
- stdweb-core (contains just webcore)
- stdweb-common (contains bindings to core js APIs, not specific to environment)
- stdweb-browser (browser-specific bindings)
- stdweb-node (node-specific bindings)

Adding *everything* into `stdweb` will increase compile times quite a bit, especially if you add the whole of canvas 2d, webgl and webgl 2

src/webapi/html_elements/canvas.rs
Diggsey: Yeah

src/lib.rs
Diggsey: Yeah

examples/canvas/src/main.rs
koute: Yeah, it's fine not adding every method there is right now.

Yes, I did think about splitting it up in the future! The first thing I'd like to split up is `webapi` and `webcore` directories, and then we'll go from there and see if/what would make sense to further split up. However, I don't want to do that just yet.

As far as the compile times go there is probably some room for improvement too without splitting it up. (E.g. once we make the `js!` macro into a procedural macro)

Diggsey:
@koute updated.

koute:
Looks good to me except you forgot to change `2D` into `2d`. (:

But to not delay merging it I'll just search & replace it myself.

Thanks a lot for the PR!

Diggsey:
Oops, the comment got lost somehow. Thanks!

Boscop:
It only support `fill_rect` right now, right?

Diggsey:
@Boscop yeah, it was just laying the ground-work, no reason not to add more though!

Boscop:
Can you please add `clearRect` and `fillText`, too? Then I can port my current polymer app to Rust :)

Issue #62, opened 2017-12-31
Media events & related stuff


emk:
Wow, `stdweb` is looking really amazing! I'm impressed at how quickly this has developed, and I've been enjoying experimenting with *.wasm code for "headless" computational tasks. But I'd like to try and port over one of my HTML-related media projects, possibly using a framework like `yew`.

To do this, I would need the [HTML media events](https://developer.mozilla.org/en-US/docs/Web/Guide/Events/Media_events), and probably some related stuff, such as methods and properties on `<video>` and `<audio>`. I'd like to try my hand at it.

Would you be interested in a PR? Any advice for anybody who wants to just dive in?

koute:
> Would you be interested in a PR?

Of course!

> Any advice for anybody who wants to just dive in?

Hmm... well, I guess look at how our already existing bindings are made and make yours similar? If you have any questions feel free to ask!

PR #81, opened 2018-01-14
WIP: HTML media events and elements

emk:
Not yet ready to merge! I'm working on implementing #62. However, I would be very interested in feedback about how I can make this code fit nicely into stdweb without causing any headaches for the maintainers.

Progress:

- [x] HTML media events. These are pretty complete at this point (except for some proprietary Mozilla extensions). There are a large number of these events, and they all have nearly-identical implementations, so I decided to implement them using a macro.
- [x] A basic `IHTMLMediaElement` trait. This is shared between `<audio>` and `<video>`. I've implemented a handful of the most important attributes and methods, but there are plenty more that could be added. Right now, I put this in the `html_elements` namespace, along with a supporting `ReadyState` type, but maybe this should live elsewhere?
- [x] `AudioElement` and `VideoElement` structs. These are just implementations of `IHTMLMediaElement`, and they shoudn't need any logic of their own, as far as I know.
- [ ] Tests. I haven't quite figured out `stdweb` testing yet, but I need to look at it before I go further.
- [ ] A real-world app. I want to actually try to use this from Rust.
- [ ] More `IHTMLMediaElement` attributes and methods?

Anyway, thank you for your encouragement, and please let me know if you have any suggestions!

### DIFF
--- a/src/webapi/html_elements/media.rs
+++ b/src/webapi/html_elements/media.rs
@@ -1,0 +1,183 @@
+use webcore::value::Reference;
+use webcore::try_from::TryInto;
+use webapi::event_target::{IEventTarget, EventTarget};
+use webapi::node::{INode, Node};
+use webapi::element::{IElement, Element};
+use webapi::html_element::{IHtmlElement, HtmlElement};
+
+/// The readiness state of a media element.
+///
+/// [(JavaScript docs)](https://developer.mozilla.org/en-US/docs/Web/API/HTMLMediaElement/readyState)
+#[derive(Copy, Clone, Debug, PartialEq, Eq, PartialOrd, Ord)]
+pub enum ReadyState {
+    /// No information is available about this media resource yet.
+    HaveNothing,
+    /// Metadata for the media has been loaded, and seeking is possible.
+    HaveMetadata,
+    /// Data is available for the current playback position, but not for more
+    /// than one frame.
+    HaveCurrentData,
+    /// Data for several frames after the current position is available, but
+    /// not enough to allow consistent playback.
+    HaveFutureData,
+    /// Enough data is available to permit uninterrupted playback at the current
+    /// download rate.
+    HaveEnoughData,
+}
+
+/// The `IHtmlElement` interface represents an `<audio>` or `<video>` element.
+///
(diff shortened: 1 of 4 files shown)


src/webapi/html_elements/media.rs
```diff
@@ -0,0 +1,183 @@
...
+use webapi::event_target::{IEventTarget, EventTarget};
+use webapi::node::{INode, Node};
+use webapi::element::{IElement, Element};
+use webapi::html_element::{IHtmlElement, HtmlElement};
+
+/// The readiness state of a media element.
+///
+/// [(JavaScript docs)](https://developer.mozilla.org/en-US/docs/Web/API/HTMLMediaElement/readyState)
+#[derive(Copy, Clone, Debug, PartialEq, Eq, PartialOrd, Ord)]
+pub enum ReadyState {
```
koute: `ReadyState` -> `MediaReadyState` maybe? (We have way to many `ReadyState`s, so I'd like them to all have unique names to reduce confusion.)

emk: Great idea! I'll rename it.

emk:
> One extra thing I'd like to request - in places you can please add links to the specs (as normal comments, not doc comments).

I've tried to add doc comments to MDN wherever MDN has actual docs. (I omitted some of the individual attribute links for `IHtmlMediaElement`, because the MDN docs were only available for the interface as a whole.)

Do you have any examples of the kind of comment that you'd like me to add? Do you have any preferences for a particular version of the HTML spec that I should link to?

I _think_ I can provide at least partial test coverage without async tests.

koute:
> I've tried to add doc comments to MDN wherever MDN has actual docs.

Yep! Those are mostly for the users.

> Do you have any examples of the kind of comment that you'd like me to add? Do you have any preferences for a particular version of the HTML spec that I should link to?

You can just put a link to the specs as a normal `//` comment right before the thing the specs is for.

These are only meant to be for us so it doesn't really matter how exactly it'll look like; it's just nice for it to be there so that when reading the code one can check how it's supposed to behave without having to search for the specs manually.

You can link to the newest version as long as it's official.

PR #82, opened 2018-01-15
Actually export mouse{down,up,move} events

iirelu:
I didn't realise in my last commit that the internal structure of the module wasn't directly exposed in the library. This commit properly "exports" the new events.


CryZe:
I've seen that this is a common problem with stdweb. Would it make sense to change the structure of the code to match the actual exposed API to avoid problems like these in the future? @koute

Diggsey:
I fixed this as part of my canvas PR too.

@CryZe I think it is useful having a different structure internally to what is exposed - to avoid problems like these I think it would be good to add examples whenever adding new features. Examples are very easy to write and serve a dual purpose of documentation and testing.

PR #86, opened 2018-01-20
Adding in MutationObserver

### LINKED ISSUE #85: Add in MutationObserver
I've been working on adding in [MutationObserver](https://developer.mozilla.org/en-US/docs/Web/API/MutationObserver), but I have some questions:

* When calling the `observe` method, you can pass in an `attributeFilter` attribute, which in JavaScript is an Array of strings. What should the type be in Rust? `Vec<String>`?

   If the correct type is `Vec<String>`, how do I convert it into a JavaScript Array so I can pass it to the `observe` method?

* `MutationObserver` accepts a callback. I took a look at how event listeners are implemented, and the callbacks have a `'static` trait bound.

   I assume that exists to guarantee that the callback will live indefinitely, because Rust doesn't have a garbage collector. Is that correct?

* I want to provide a nice Rust-y API for `MutationRecord`, something like this:

   ```rust
   pub enum MutationRecord {
       Attribute {
           target: Node,
           attribute_name: Option< String >,
           attribute_namespace: Option< String >,
           old_value: Option< String >,
       },

       CharacterData {
           target: Node,
           old_value: Option< String >,
       },

       ChildList {
           target: Node,
           added_nodes: NodeList,
           removed_nodes: NodeList,
           previous_sibling: Option< Node >,
           next_sibling: Option< Node >,
       },
   }
   ```

   I tried to add in an `impl TryFrom< Value > for MutationRecord` implementation, so I could convert from JS objects into `MutationRecord`, but I'm getting this error:
   ```
   error[E0277]: the trait bound `std::option::Option<webapi::node::Node>: webcore::try_from::TryFrom<webcore::value::Value>` is not satisfied
      --> src/webapi/mutation_observer.rs:124:71
       |
   124 |                         previous_sibling: js!( @{r}.previousSibling ).try_into()?,
       |                                                                       ^^^^^^^^ the trait `webcore::try_from::TryFrom<webcore::value::Value>` is not implemented for `std::option::Option<webapi::node::Node>`
       |
                 <std::option::Option<bool> as webcore::try_from::TryFrom<webcore::value::Value>>
                 <std::option::Option<u8> as webcore::try_from::TryFrom<webcore::value::Value>>
                 <std::option::Option<u16> as webcore::try_from::TryFrom<webcore::value::Value>>
                 <std::option::Option<u32> as webcore::try_from::TryFrom<webcore::value::Value>>
               and 14 others
       = note: required because of the requirements on the impl of `webcore::try_from::TryInto<std::option::Option<webapi::node::Node>>` for `webcore::value::Value`
   ```

   Basically, I'm trying to convert from a JavaScript value (which is either `null` or a DOM `Node`) into an `Option< Node >` in Rust, but it's not working. How do I fix this?

You can see my (unfinished) work [here](https://github.com/Pauan/stdweb/blob/f77682bf4e4abcbe6b37bd341468815fffbf59db/src/webapi/mutation_observer.rs).

Pauan:
Fixes #85 

@koute Requesting a review on this, since I'm not 100% sure how all these abstractions work yet.

Also, I haven't tested this yet, and it's missing unit tests. I can't really add in unit tests, because the unit tests need to be async.

Also, I tried to keep the API the same as the web, but I did significantly change `MutationRecord`, because that makes the API **much** safer and more Rust-y.

### DIFF
--- a/src/webapi/mutation_observer.rs
+++ b/src/webapi/mutation_observer.rs
@@ -230,19 +230,17 @@ mod tests {
 
     #[test]
     fn test_observe() {
-        let observer = MutationObserver::new( |records| {
-
-        } );
+        let observer = MutationObserver::new( |_| {} );
 
         // TODO replace with document.body
-        observer.observe( &document(),  MutationObserverInit {
+        observer.observe( &document(),  &MutationObserverInit {
             child_list: true,
             attributes: false,
             character_data: true,
             subtree: true,
             attribute_old_value: false,
             character_data_old_value: false,
-            attribute_filter: &[ "foo", "bar", "qux" ],
+            attribute_filter: Some( &[ "foo", "bar", "qux" ] ),
         } );
     }
 }
--- a/src/webapi/mutation_observer.rs
+++ b/src/webapi/mutation_observer.rs
@@ -179,7 +179,7 @@ impl Drop for MutationObserverHandle {
         self.disconnect();
 
         js! { @(no_return)
-            @{self.callback_reference}.drop();
+            @{&self.callback_reference}.drop();
         }
     }
 }
--- a/src/webapi/mutation_observer.rs
+++ b/src/webapi/mutation_observer.rs
@@ -99,7 +99,7 @@ impl MutationObserver {
     /// [`character_data`](struct.MutationObserverInit.html#structfield.character_data) must be `true`.
     ///
     /// [(JavaScript docs)](https://developer.mozilla.org/en-US/docs/Web/API/MutationObserver#observe())
-    pub fn observe< T: INode >( &self, target: &T, options: &MutationObserverInit ) {
+    pub fn observe< T: INode >( &self, target: &T, options: MutationObserverInit ) {
         let attribute_filter: Value = match options.attribute_filter {
             Some( a ) => a.into(),
             // This must compile to JavaScript `undefined`, NOT `null`
@@ -244,7 +244,7 @@ mod tests {
         let observer = MutationObserver::new( |_, _| {} );
 
         // TODO replace with document.body
-        observer.observe( &document(),  &MutationObserverInit {
+        observer.observe( &document(),  MutationObserverInit {
             child_list: true,
             attributes: true,
             character_data: true,
--- a/src/webapi/mutation_observer.rs
+++ b/src/webapi/mutation_observer.rs
@@ -325,7 +325,5 @@ mod tests {
             character_data_old_value: true,
             attribute_filter: Some( &[ "foo", "bar", "qux" ] ),
         } );
-
-        observer.disconnect();
     }
 }
--- a/src/webapi/mutation_observer.rs
+++ b/src/webapi/mutation_observer.rs
@@ -241,7 +241,7 @@ mod tests {
 
     #[ test ]
     fn test_observe() {
-        let observer = MutationObserver::new( |_| {} );
+        let observer = MutationObserver::new( |_, _| {} );
 
         // TODO replace with document.body
         observer.observe( &document(),  &MutationObserverInit {
--- a/src/webapi/mutation_observer.rs
+++ b/src/webapi/mutation_observer.rs
@@ -20,7 +20,7 @@ reference_boilerplate! {
 /// Only used with the [`MutationObserver::observe`](struct.MutationObserver.html#method.observe) method.
 ///
 /// [(JavaScript docs)](https://developer.mozilla.org/en-US/docs/Web/API/MutationObserver#MutationObserverInit)
-#[derive( Debug, Clone )]
+#[ derive( Debug, Clone ) ]
 pub struct MutationObserverInit< 'a > {
     /// If `true` it will observe all inserts and removals of the target's children (including text nodes).
     ///
@@ -146,7 +146,7 @@ impl MutationObserver {
 /// It is passed to [`MutationObserver`](struct.MutationObserver.html)'s callback.
 ///
 /// [(JavaScript docs)](https://developer.mozilla.org/en-US/docs/Web/API/MutationRecord)
-#[derive( Debug, Clone )]
+#[ derive( Debug, Clone ) ]
 pub enum MutationRecord {
     /// One of the target's attributes was changed.
     Attribute {
@@ -236,7 +236,7 @@ mod tests {
     use super::*;
     use webapi::document::document;
 
-    #[test]
+    #[ test ]
     fn test_observe() {
         let observer = MutationObserver::new( |_| {} );
 
--- a/src/webapi/mutation_observer.rs
+++ b/src/webapi/mutation_observer.rs
@@ -4,7 +4,7 @@ use webcore::try_from::{TryFrom, TryInto};
 use webapi::node::{INode, Node};
 
 
-/// `MutationObserver` provides developers with a way to react to changes in a DOM.
+/// `MutationObserver` provides a way to receive notifications about changes in the DOM.
 ///
 /// [(JavaScript docs)](https://developer.mozilla.org/en-US/docs/Web/API/MutationObserver)
 pub struct MutationObserver( Reference );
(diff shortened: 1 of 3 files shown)


Pauan:
I did some light testing, it appears to work correctly, but I need to do some more thorough tests.

However, I have a question. Consider this code:

```rust
let observer1 = MutationObserver::new(|records, observer2| {});
```

The `observer1` and `observer2` variables are exactly the same object in JavaScript. In other words, `observer1 === observer2`

What's a good way of describing that with Rust types? Should `observer2` have the type `MutationObserver`, or `&MutationObserver`, or `&mut MutationObserver`, or something else entirely?

src/webapi/mutation_observer.rs
```diff
@@ -0,0 +1,259 @@
...
+    ///
+    /// 1. Vector of changes to the observed DOM nodes.
+    ///
+    /// 2. The `MutationObserver`.
+    ///
+    /// [(JavaScript docs)](https://developer.mozilla.org/en-US/docs/Web/API/MutationObserver#Constructor)
+    pub fn new< F >( callback: F ) -> Self
+        where F: FnMut( Vec< MutationRecord >, Self ) + 'static {
+        js! (
+            return new MutationObserver( @{callback} );
```
koute: Rust's `Fn` and `FnMut` callbacks when passed to the `js!` macro must to be deallocated manually by calling `.drop()`. Right now any callback passed into `MutationObserver::new` will permanently leak memory, which is not good.

So, we need to figure out what to do with this. As far as I can see there are two cases we have to deal with:

1) We receive a `MutationObserver` created on the JavaScript side with a JavaScript callback. We treat this like any other `Reference` value.
2) We create our own `MutationObserver` with `new()`. We then know that we have a closure on Rust's side that needs to be deallocated.

One random idea that I have to deal with (2) would be modifying the `new` to return something like `OwnedMutationObserver` instead of `MutationObserver`. The `OwnedMutationObserver` would implement `Drop` where it would disconnect the observer and deallocate the callback. That way when the `OwnedMutationObserver` goes out of scope our callback would be automatically deallocated.

src/webapi/mutation_observer.rs
```diff
@@ -0,0 +1,259 @@
...
+    /// [`attributes`](struct.MutationObserverInit.html#structfield.attributes) must be `true`.
+    ///
+    /// * If [`attribute_filter`](struct.MutationObserverInit.html#structfield.attribute_filter) is `Some`, then
+    /// [`attributes`](struct.MutationObserverInit.html#structfield.attributes) must be `true`.
+    ///
+    /// * If [`character_data_old_value`](struct.MutationObserverInit.html#structfield.character_data_old_value) is `true`, then
+    /// [`character_data`](struct.MutationObserverInit.html#structfield.character_data) must be `true`.
+    ///
+    /// [(JavaScript docs)](https://developer.mozilla.org/en-US/docs/Web/API/MutationObserver#observe())
+    pub fn observe< T: INode >( &self, target: &T, options: &MutationObserverInit ) {
```
koute: In this case you want to take `MutationObserverInit` instead of `&MutationObserverInit`.

As I've previously explained - yes, we usually want to take a borrow, *but* for a single-use helper structure like this which is used to emulate named parameters there isn't really a point.

src/webapi/mutation_observer.rs
```diff
@@ -0,0 +1,259 @@
...
+        let attribute_filter: Value = match options.attribute_filter {
```
koute: This would be slightly nicer than a `match` I think:

```rust
options.attribute_filter.map( |val| val.into() ).unwrap_or( Value::Undefined )
```

src/webapi/mutation_observer.rs
```diff
@@ -0,0 +1,259 @@
...
+}
+
+
+impl TryFrom< Value > for MutationRecord {
+    type Error = ConversionError;
+
+    fn try_from( v: Value ) -> Result< Self, Self::Error > {
+        match v {
+            Value::Reference( ref r ) => {
+                let _type: String = js!( return @{r}.type ).try_into()?;
```
koute: We usually use the name `kind`. The `_` prefix in Rust has a special meaning - it makes the compiler effectively ignore the variable when generating warnings, and we don't want that here.

src/webapi/mutation_observer.rs
```diff
@@ -0,0 +1,259 @@
...
+                let target: Node = js!( return @{r}.target ).try_into()?;
+
+                match _type.as_str() {
+                    "attributes" => Ok( MutationRecord::Attribute {
+                        target: target,
+                        name: js!( return @{r}.attributeName ).try_into()?,
```
koute: All of these `js!` snippets are missing a `;` inside of them at the end.

koute:
Yeah, the lack of support for async tests is kind of a bummer. I have some ideas on how to make them possible, but that's going to take some time.

For now as a substitute you could add an example in the `examples` directory and we can eyeball whenever it works.

src/webapi/mutation_observer.rs
Pauan: I'm still a bit fuzzy on when to borrow and when to own, and your comment isn't making that any clearer. 😛 But I'll change it.

src/webapi/mutation_observer.rs
Pauan: Ah, good catch! It seems I'm still too used to JavaScript, where using `_` prefix is the norm.

src/webapi/mutation_observer.rs
Pauan: If we have `new` return an `OwnedMutationObserver` which cleans things up when it goes out of scope, does the `FnMut` still need a `'static` lifetime?

Pauan:
I fixed all the suggestions, except:

* I still need to figure out the memory leak situation with the callback.

* I will create an example in a different pull request, because it requires some changes to non-MutationObserver things.

src/webapi/mutation_observer.rs
koute: Yes, it still needs the `'static` lifetime; even if we'll make it automatically drop the callback when the `OwnedMutationObserver` goes out of scope the actual `OwnedMutationObserver` still could be, for example, assigned to some global variable and kept alive forever.

koute:
Oh also, I've missed one of your questions.

> The `observer1` and `observer2` variables are exactly the same object in JavaScript. In other words, `observer1 === observer2`
> What's a good way of describing that with Rust types? Should `observer2` have the type `MutationObserver`, or `&MutationObserver`, or `&mut MutationObserver`, or something else entirely?

I'm not really sure what you're asking for here? You're already using `Self` (which roughly means "type for which the current `impl` block is being defined for") in the prototype of `new`. Besides that you can't really assert in the type system that two objects have the same *value*.

src/webapi/mutation_observer.rs
Pauan: But memory leaks are always possible in Rust, and that doesn't normally warrant `'static` lifetimes, so what's special about this situation? Since the reference will be held indefinitely in both JS and Rust, the memory should always be valid, right?

Pauan:
> I'm not really sure what you're asking for here?

Yes it's currently using `Self`, but in Rust there can only be a single owner, by making the callback take `Self` that allows for the same object to be owned at two different places (`observer1` and `observer2`), so shouldn't the callback take `&mut Self` instead, to ensure that the object always has a single owner?

Similarly, shouldn't the "mutable" methods like `observe`, `disconnect`, and `take_records` take in `&mut self` rather than `&self`? Logically speaking they are mutating the object, and using `&mut` ensures that we get the protection of Rust's static type system.

src/webapi/mutation_observer.rs
koute: As far as the `'static` lifetime is concerned it's not about memory leaks, it's about safety, and it's not really about the closure *itself*, but about what it's allowed to close over. In a nutshell the `'static` lifetime tells the compiler that any references passed into the closure by whoever creates it can only point to stuff with a lifetime of the whole program, so in particular they can *not* reference anything on the stack for example.

Technically we *could* make the closure have an arbitrary lifetime `'a` and then make the return type `OwnedMutationObserver< 'a >`, but it's probably not worth it.

koute:
Ah yes, right. In this case I don't think it matters though, since almost all of our types in `webapi` are essentially reference types. (So even in situations when they're passed by-value they still act like a reference, or in other words - just because you have an instance of `Foobar` doesn't mean you're the owner.)

Semantically using `&mut` doesn't have much point here as the JavaScript code itself can hold a reference to it. So we're using a concept called [interior mutability](https://doc.rust-lang.org/book/second-edition/ch15-05-interior-mutability.html) throughout the whole `stdweb`, hence everything takes `&self`, and not `&mut self`.

src/webapi/mutation_observer.rs
Pauan: Ahh, right, the closure could hold references to a variable on the stack, which then won't exist anymore when the callback is called. I'll experiment with using `'a`, but you may be right that `'static` is necessary.

Pauan:
> In this case I don't think it matters though, since almost all of our types in webapi are essentially reference types.

Understood. Sorry about all these obvious questions: I understand JavaScript, and I *mostly* understand Rust, it's the interaction between Rust and JS (and all of the `stdweb` abstractions) that I'm not fully understanding.

> Semantically using &mut doesn't have much point here as the JavaScript code itself can hold a reference to it.

You're completely right: JavaScript can mutate the object even if it's marked as immutable on the Rust side. So using `&mut` is indeed superfluous.

Pauan:
@koute Requesting another (hopefully final) review of this. I didn't make very many changes, but I did add in `IMutationObserver` and `MutationObserverHandle`.

The only parts that I'm unsure about are `MutationObserver::new` and `MutationObserverHandle::drop`

----

Using `'a` rather than `'static` ended up being trickier than I thought, so I'll add that in a different pull request. Therefore, this pull request is ready to merge (assuming you approve the changes).

src/webapi/mutation_observer.rs
```diff
@@ -4,7 +4,60 @@ use webcore::try_from::{TryFrom, TryInto};
 use webapi::node::{INode, Node};
 
 
-/// `MutationObserver` provides a way to receive notifications about changes in the DOM.
+/// The `IMutationObserver` interface represents objects which can receive notifications about changes to the DOM.
+///
+/// [(JavaScript docs)](https://developer.mozilla.org/en-US/docs/Web/API/MutationObserver)
+pub trait IMutationObserver {
```
koute: Hmm... the `IMutationObserver` is probably unnecessary. We can just implement `Deref` for `MutationObserverHandle` and have it return `MutationObserver`. (The `MutationObserverHandle` is, after all, only here to handle the lifetime of the observer; it's not a *different* kind of an object implementing the same interface.)

Pauan: I think including `IMutationObserver` makes sense because:

* It's an interface in the DOM spec, which means that theoretically multiple objects can implement it.

   Right now there's only one object implementing it, but making it a `trait` is forwards-compatible.

* It's consistent with other interfaces like `IElement` and `INode`.

But I don't have strong opinions about this, so I don't mind removing `IMutationObserver`

Implementing `Deref` sounds like a good idea, I'll do that.

koute: It is indeed an interface, but if only a single object implements it then there isn't really much point - it just adds extra complexity for not much gain. It's not like the interfaces are actually anything concrete in JS either - they're just contracts to which a given type must adhere.

While it's common in some object oriented languages to have useless interfaces which look good on paper I'd like to think that in Rust we tend to not do that if there isn't actually a good practical reason for it.

Pauan: My experience is not with OOP languages, it is with functional languages like Haskell, F#, etc. which use interfaces / typeclasses copiously (and for good reason!).

I can understand why you would think that adding in traits with a single implementation is useless, but there are some practical advantages to it:

* It is easier to write unit tests, because you can simply create a mock object which implements the trait.

* Users can create their own custom implementations, outside of `stdweb`.

   For example, I can imagine React Native (or similar) creating an implementation of `IMutationObserver` for iOS / Android.

   Or a virtual DOM library could implement `IMutationObserver` for virtual DOM nodes.

* If a user writes a function with this type signature...

   ```rust
   fn foo<A: IMutationObserver>(observer: &A)
   ```

   ...then it is forward-compatible and will work with all present and future implementations, but if the user writes this type signature...

   ```rust
   fn foo(observer: &MutationObserver)
   ```

   ...then it can *only* be used with `MutationObserver`, nothing else. This hurts code reuse.

* Traits let you precisely specify your intentions in the code, so the code uses exactly the methods that it needs to use, and no more. Using a concrete type allows it to use *any* method on the concrete type, which makes it harder to understand what the code is doing.

   As an example, if I see `fn foo<A: IMutationObserver>(observer: &A)` then I know *exactly* what methods it can use (`observe`, `disconnect`, and `take_records`), but if I see `fn foo(observer: &MutationObserver)` then it's harder to know which methods it's using or not (without looking at the implementation of `foo` and all the functions that `foo` calls).

* Trait objects let you combine multiple different types together into the same data structure (e.g. a vector), whereas you can't combine concrete types like `Element` and `Node` together (without casting them first).

As such, I don't think using traits is useless, even if they only have a single implementation. In my opinion the complexity of using traits is only slightly more than the complexity of not using traits, and you gain great flexibility by using traits.

I would even go so far as to say that we should actively avoid using concrete types like `Node` and `Element` and should instead prefer trait objects like `&INode`. But that's going pretty far off topic...

In any case, as I said I don't have strong opinions about this, so I'll remove `IMutationObserver`

Pauan:
@koute I removed `IMutationObserver`, added in `Deref` for `MutationObserverHandle`, and also removed the `mutation_observer` method from `MutationObserverHandle` since it's now redundant.

This should be ready to merge now.

koute:
Thanks!

Issue #21, opened 2017-10-01
Use error-chain


sphinxc0re:
I recognized that a most parts of the api either `panic!` or return an `Option<T>`. `Option<T>` should only be used when there legitimately might be nothing to return. In any other case I would sugges to use the `Result<T>` that the `error-chain` crate offers.

I wouldn't use the actual chaining feature of the crate in tight loops as this can lead to severe performance issues.

I would like to implement this myself if it is a wanted feature. The `error-chain` crate also is very well maintained and already found a good adoption rate in the Rust community.

koute:
Could you perhaps give a few examples of places where we return an `Option<T>` when we should return a `Result<T, E>` or there's a `panic!` that shouldn't be there? (There are a few stray panics, but they're mostly there because the error handling in that particular place hasn't been properly implemented yet.)

sphinxc0re:
https://github.com/koute/stdweb/blob/5926b666004ecaeb7bc27e23b3c9fa40dbcff7b4/src/webapi/document.rs#L72
https://github.com/koute/stdweb/blob/5926b666004ecaeb7bc27e23b3c9fa40dbcff7b4/src/webapi/document.rs#L81
These are lines where you'd just call `unwrap()` on the result. This is usually considered "bad style".

https://github.com/koute/stdweb/blob/5926b666004ecaeb7bc27e23b3c9fa40dbcff7b4/src/webapi/document.rs#L37-L41
You even wrote a comment for that one. My suggestion is to simply wrap every call that could probably fail with an error message to produce a `Result<T>`

koute:
The first two examples you've given **should** call `unwrap`. The `createElement` function (according to the documentation) will **always** return a valid reference to an element. Ditto for the `createTextNode`.

You need to remember that `unwrap` is effectively the same as an `assert!`, telling the compiler that "yes, we'll always have a value here, trust me"; using `unwrap` **is** bad style, but **only** when you're unwrapping possible errors - here there are no possible errors, so it would actually be a bad style to return an `Result<T, E>` even though you know that it will **never** contain `E`.

In the case of the third example - yes, it should return a `Result`.

CryZe:
The documentation is wrong then. createElement can throw exceptions if the tag name is not valid.

koute:
The MDN explicitly says that:

```
In an HTML document, the Document.createElement() method creates the HTML element specified by tagName, or an HTMLUnknownElement if tagName isn't recognized.
```

[...]

PR #89, opened 2018-01-23
Implement Error and DOMException. Add patterns for error handling.

Herschel:
I've had this floating around on my hard drive for a bit -- there is room for improvement, feedback requested!

Addresses #28. Methods that throw an error were using an ad-hoc `js!` block to catch the error and transfer it back. However, this gets difficult when a method returns multiple error types. This PR adds macros to make error handling more consistent and easy.

- Implement Error and DOMException types.
- Implement a `js_try!` macro that works like `js!`, but also catches errors and convert them to a `Result` type.
  - This macro relies on `TryFrom<Value>` being implemented for both the value and error type.
  - Specifically for `DOMException`, these have the same type in JS, but different names. For the concrete exception types in Rust, `FromReference` and `TryFrom` is implemented to check that the name matches.
  - The implementation of `js_try!` isn't optimal, with lots of calls back and forth to JS. I expect this could be done much better, perhaps by adding serialization support for a Result type. I was more concerned with establishing the pattern of how the errors should be caught on the Rust side; the macro can be improved.
  - Unlike the `js!` macro, `js_try!` does a `try_into` internally.
- For calls that may return several different errors, an enum can be used, and the `error_enum_boilerplate!` macro is used to implement `TryFrom<Value>` for the enum.
  - This will try to convert the value to each variant in sequence.
  - My macro-fu is a little weak, so maybe the syntax for this macro should be different. Even better, I think this could be done as a custom derive if we're okay with that.
- As an example, the `js_try!` macro is used to return errors from these APIs:
  - `Node::replace_child`, `Node::insert_before`
  - `WebSocket::new` and `WebSocket::close_with_status` (thanks @Diggsey!)

Things left for a future PR:
- Backtraces (could be placed in the `IError` trait?).
- Using a crate such as failure.
  - It didn't seem like there was a consensus on this yet.
  - #21 specified that error-chain was too heavy, but failure might be more appropriate. It would allow us to upcast into a unifying Error type. Maybe it's more ergonomic for failing methods to return `Result<_, failure::Error>` instead of defining an enum-per-throwing-function. But I think it's nice to know explicitly what errors a function might throw.

### DIFF
--- a/src/webapi/error.rs
+++ b/src/webapi/error.rs
@@ -1,0 +1,67 @@
+use webcore::value::{Value, Reference};
+use webcore::try_from::{TryFrom, TryInto};
+
+/// Represents the JavaScript `Error` interface. An `Error` is thrown whenever a run-time error
+/// occurs.
+///
+/// [(JavaScript docs)](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Error)
+pub trait IError: AsRef< Reference > + TryFrom< Value > {
+    /// Returns a human-readable description of the error.
+    ///
+    /// [(JavaScript docs)](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Error/message)
+    #[inline]
+    fn message( &self ) -> String {
+        js!(
+            return @{self.as_ref()}.message;
+        ).try_into().unwrap()
+    }
+
+    /// Returns a name specifiying the type of error.
+    ///
+    /// [(JavaScript docs)](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Error/name)
+    #[inline]
+    fn name( &self ) -> String {
+        js!(
+            return @{self.as_ref()}.name;
+        ).try_into().unwrap()
+    }
+}
+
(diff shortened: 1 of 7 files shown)


Diggsey:
This looks great!

A few comments:
- I wonder if `js_try!` should include the `unwrap` - given the way it is used, it's unlikely the caller will ever *not* want to unwrap the conversion, and if the macro performs it, then it can give nicer behaviour in case the unwrap fails.

- DOMExceptions should expose their type as an enum rather than a string.

- DOMExceptions should be constructible by type.

src/webapi/error.rs
```diff
@@ -0,0 +1,67 @@
...
+
+        let error: Error = js!(
+           return new Error("foo");
+        ).try_into().unwrap();
+
+        assert_eq!(error.name(), "Error");
+        assert_eq!(error.message(), "foo");
+        
+        let mut text = String::new();
+        write!(&mut text, "{}", error).unwrap();
```
koute: Can't you just do `let text = format!("{}", error);` here? (:

koute:
Since the comments I've made are pretty minor and this in general looks pretty solid (and I want to release a new `stdweb` with this soon) I'm going to merge it in for now. We can further tweak it as we go.

Thanks a lot @Herschel!

PR #90, opened 2018-01-24
IElement should inherit from INode

Pauan:
This follows the [specification](https://developer.mozilla.org/en-US/docs/Web/API/element).


koute:
Good catch! Thanks!

PR #91, opened 2018-01-24
Squelch some unused_parens warnings on nightly

Herschel:
Nightly currently produces a ton of warnings about unused parentheses. Most seem to be false positives from the macro syntax. Added `#[allow(unused_parens)]` to these cases.


koute:
Thanks!

PR #94, opened 2018-01-24
Expand CanvasRenderingContext API

teovoinea:
Added the [fillText](https://developer.mozilla.org/en-US/docs/Web/API/CanvasRenderingContext2D/fillText) function from the CanvasRenderingContext2D API

### DIFF
--- a/src/webapi/rendering_context.rs
+++ b/src/webapi/rendering_context.rs
@@ -39,4 +39,23 @@ impl CanvasRenderingContext2d {
             @{&self.0}.fillRect(@{x}, @{y}, @{width}, @{height});
         }
     }
+
+    /// Draws a text string at the specified coordinates, filling the string's characters 
+    /// with the current foreground color. An optional parameter allows specifying a maximum 
+    /// width for the rendered text, which the user agent will achieve by condensing the 
+    /// text or by using a lower font size.
+    /// 
+    /// [(JavaScript docs)](https://developer.mozilla.org/en-US/docs/Web/API/CanvasRenderingContext2D/fillText)
+    pub fn fill_text(&self, text: &str, x: f64, y: f64, max_width: Option<f64>) {
+        if let Some(max_width) = max_width {
+            js! { @(no_return)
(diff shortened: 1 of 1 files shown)


src/webapi/rendering_context.rs
```diff
@@ -39,4 +39,23 @@ impl CanvasRenderingContext2d {
...
+    /// Draws a text string at the specified coordinates, filling the string's characters 
+    /// with the current foreground color. An optional parameter allows specifying a maximum 
+    /// width for the rendered text, which the user agent will achieve by condensing the 
+    /// text or by using a lower font size.
+    /// 
+    /// [(JavaScript docs)](https://developer.mozilla.org/en-US/docs/Web/API/CanvasRenderingContext2D/fillText)
+    pub fn fill_text(&self, text: &str, x: f64, y: f64, max_width: Option<f64>) {
+        if let Some(max_width) = max_width {
+            js! { @(no_return)
+                @{&self.0}.fillText(@{text}, @{x}, @{y}, @{max_width});
```
koute: Is the `if` necessary here? What happens if you pass `null` as the last argument? What if you'd pass `undefined`? I looked [at the specs](https://html.spec.whatwg.org/multipage/canvas.html#text-preparation-algorithm) for the text preparation algorithm this method uses, but to me they seem... ambiguous. It says that `"If maxWidth was provided but is less than or equal to zero or equal to NaN, then return an empty array."`, so depending on how you interpret "provided" (does the arg count matter here or the value itself?) and "less than or equal" (is it a strict comparison, or weak one?) it can have different behavior.

teovoinea: `Undefined` works fine with something like `@{&self.0}.fillText(@{text}, @{x}, @{y}, @{Undefined});`, the issue is you would still need the if statement for unwrapping `max_width` because `unwrap_or` on an `f64` has a type mismatch with `Undefined`.

koute: Hmm.. yeah, since `None` converts to `null` and not `undefined` I guess it needs to stay this way for now.

koute:
Thanks!

PR #96, opened 2018-01-27
Complete CanvasRenderingContext2D API

teovoinea:
I plan on adding all of the methods and properties of [CanvasRenderingContext2D](https://developer.mozilla.org/en-US/docs/Web/API/CanvasRenderingContext2D). Currently a work in progress but I've opened this PR to get some feedback on the implementation of certain functions.

Starting with [`clip`](https://github.com/koute/stdweb/pull/96/commits/4032862cec0b8fb9861811d876ed8c232aa33732#diff-f6c540db6098f68dcb244aeeb396e8f8R103)  and [`create_pattern_image`](https://github.com/koute/stdweb/pull/96/commits/4a6fdde6925227272b4b8fa6200d5f8285ec3826#diff-f6c540db6098f68dcb244aeeb396e8f8R161) where I had to create Enums.

As well as in [createPattern](https://developer.mozilla.org/en-US/docs/Web/API/CanvasRenderingContext2D/createPattern) which takes a [CanvasImageSource](https://developer.mozilla.org/en-US/docs/Web/API/CanvasImageSource) but I've currently only implemented the `ImageElement` as `create_pattern_image`.


src/webapi/rendering_context.rs
```diff
@@ -15,6 +17,20 @@ pub trait RenderingContext {
 /// [(JavaScript docs)](https://developer.mozilla.org/en-US/docs/Web/API/CanvasRenderingContext2D)
 pub struct CanvasRenderingContext2d(Reference);
 
+#[derive(Debug)]
```
koute: For enums like these we also would want `Copy`, `Clone`, `PartialEq`, `Eq` derived.

koute: Also missing docs. (:

src/webapi/rendering_context.rs
```diff
@@ -30,6 +46,201 @@ impl RenderingContext for CanvasRenderingContext2d {
...
+    pub fn close_path(&self) {
+        js! { @(no_return)
+            @{&self.0}.closePath();
+        }
+    }
+
+    /// Creates a gradient along the line given by the coordinates represented by the parameters.
+    /// 
+    /// [(JavaScript docs)](https://developer.mozilla.org/en-US/docs/Web/API/CanvasRenderingContext2D/createLinearGradient)
+    pub fn create_linear_gradient(&self, x0: f64, y0: f64, x1: f64, y1: f64) {
```
koute: You don't have to implement every type, but in case a function returns something (like a `CanvasGradient` in this case) please create a dummy private type and return it. (Just so that we won't have to change the prototype in the future.)

src/webapi/rendering_context.rs
```diff
@@ -30,6 +46,201 @@ impl RenderingContext for CanvasRenderingContext2d {
...
+                }
+
+                Repitition::NoRepeat => {
+                    repitition_string = "no-repeat";  
+                }
+            }
+        }
+        else {
+            repitition_string = "repeat";
+        }
```
koute: You can do this with a single `match` statement. (In this and other methods.)

```
let repetition = match repitition {
    Some( Repetition::Repeat ) | None => "repeat",
    Some( Repetition::RepeatX ) => "repeat-x",
};
```

Also, a typo - `repitition` -> `repetition`. (:

src/webapi/rendering_context.rs
```diff
@@ -30,6 +46,201 @@ impl RenderingContext for CanvasRenderingContext2d {
...
+        //TODO: returns radial gradient
+        js! { @(no_return)
+            @{&self.0}.createRadialGradient(@{x0}, @{y0}, @{r0}, @{x1}, @{y1}, @{r1});
+        }
+    }
+
+    /// Draws a focus ring around the current path or given path, If a given element is focused.
+    /// 
+    /// [(JavaScript docs)](https://developer.mozilla.org/en-US/docs/Web/API/CanvasRenderingContext2D/drawFocusIfNeeded)
+    pub fn draw_focus_if_needed(&self, element: HtmlElement) {
```
koute: This should probably be generic over `< T: IHtmlElement >` and take `element; &T` to be consistent with the rest of `stdweb`.

src/webapi/rendering_context.rs
teovoinea: Thanks! That's a lot cleaner :)

teovoinea:
On my list of things to do for this PR:

- [ ] Some of the functions like [getImageData](https://developer.mozilla.org/en-US/docs/Web/API/CanvasRenderingContext2D/getImageData) throw errors, I'm going to make those return `Result<_, _>`
- [ ] Implement the `properties` of `CanvasRenderingContext2D`
- [ ] Implement the return object of some of the functions (eg: `CanvasGradient` and `CanvasPattern`)

teovoinea:
@koute Recently the builds have been [complaining](https://travis-ci.org/koute/stdweb/jobs/337116308#L501) about the most recent `cargo-web` version. My branch is up to date with master.

koute:
That's one massive PR, thanks a lot!

Yes, Travis was complaining but it's fixed now.

Ok, so, I'm sure this is going to be a huge pain in the rear for you, but could you please add links to the specs to every method you've added? (Look at the `fill_rect` and `fill-text` methods for how I'd like it to be done.) Basically from now on I'd like to have those everywhere as it's then a lot easier to review that we're actually binding to what's in the specs.

teovoinea:
So I was looking at the links provided for `fill_rect` and `fill_text` and they seem to be [broken](https://html.spec.whatwg.org/#2dcontext:#dom-context-2d-fillrect)/not lead to where you want them to. Whereas [CanvasRenderingContext2D](https://html.spec.whatwg.org/#canvasrenderingcontext2d) **does** link properly. Did you mean to have `fill_rect`, for example, link go to [this](https://html.spec.whatwg.org/#dom-context-2d-fillrect) url?

koute:
No, those links are the ones I've meant to put, however you're right that they don't work. I've copy-pasted them incorrectly - if you remove the second `#` they will work. Thanks for noticing!

src/webapi/rendering_context.rs
```diff
@@ -16,11 +18,187 @@ pub trait RenderingContext {
...
+
+error_enum_boilerplate! {
+    TypeMismatchError,
+    SyntaxError
+}
+
+error_enum_boilerplate! {
+    NsErrorNotAvailable,
+    SecurityError
+}
```
koute: 1. This doesn't seem to be used currently; do you intend to use it somewhere?
2. Isn't this Firefox-specific? (Due to the `Ns` prefix?)

teovoinea: Yup! You're right, I missed that

teovoinea:
This PR is 99% ready, only thing that I'm missing and having a bit of trouble with is figuring out how to catch the errors being thrown [here](https://github.com/koute/stdweb/pull/96/files#diff-f6c540db6098f68dcb244aeeb396e8f8R925). I haven't been able to reproduce `INDEX_SIZE_ERROR` even in regular js, and I was able to get `INVALID_STATE_ERR` thrown but can't figure out how to capture it in stdweb. Any suggestions?

teovoinea:
I've done a js example testing all of the possible errors that can be thrown from `draw_image`, only `INVALID_STATE_ERR` is ever thrown. You can verify with this [jsfiddle](https://jsfiddle.net/9ort3eLk/). I've left TODOs for the other errors in the function documentation and removed their definitions. I'm marking this as ready to review/merge.

teovoinea:
@koute I updated the PR with the new reference system.

ie:
```
#[derive(Clone, Debug, ReferenceType)]
#[reference(instance_of = "CanvasGradient")]
```

Are there any more comments for review? I've added all the features I was planning from my end.

koute:
@teovoinea Sorry for the delay! I'll look through this more thoroughly very soon and let you know. Thanks!

src/webapi/rendering_context.rs
```diff
@@ -18,6 +20,160 @@ pub trait RenderingContext {
 #[reference(instance_of = "CanvasRenderingContext2D")]
 pub struct CanvasRenderingContext2d(Reference);
 
+/// The CanvasGradient interface represents an opaque object describing a gradient. 
+/// It is returned by the methods CanvasRenderingContext2D.createLinearGradient() or 
+/// CanvasRenderingContext2D.createRadialGradient().
```
koute: Very minor nitpick - these docs could use some light editing. (e.g. the MDN docs do say "interface", but in Rust this is somewhat confusing where by "interface" you'd usually think of traits, and these are concrete structures.)

src/webapi/rendering_context.rs
```diff
@@ -18,6 +20,160 @@ pub trait RenderingContext {
...
+/// 
+/// [(JavaScript docs)](https://developer.mozilla.org/en-US/docs/Web/API/CanvasRenderingContext2D/lineJoin)
+#[derive(Debug, Copy, Clone, PartialEq, Eq)]
+pub enum LineJoin {
+    Bevel,
+    Round,
+    Miter
+}
+
+/// A DOMString indicating how to repeat the image.
```
koute: Minor nitpick: `DOMString` part is irrelevant, since in Rust this is an enum. (:

src/webapi/rendering_context.rs
```diff
@@ -27,7 +183,771 @@ impl RenderingContext for CanvasRenderingContext2d {
     }
 }
 
+impl CanvasGradient {
+
+    /// Adds a new stop, defined by an offset and a color, to the gradient. If the offset is 
+    /// not between 0 and 1, an INDEX_SIZE_ERR is raised, if the color can't be parsed as a 
+    /// CSS <color>, a SYNTAX_ERR is raised.
+    ///
```
koute: There are no exceptions in Rust so talking about raising exceptions doesn't really have much point. (:

src/webapi/rendering_context.rs
```diff
@@ -27,7 +183,771 @@ impl RenderingContext for CanvasRenderingContext2d {
...
 
+impl CanvasGradient {
+
+    /// Adds a new stop, defined by an offset and a color, to the gradient. If the offset is 
+    /// not between 0 and 1, an INDEX_SIZE_ERR is raised, if the color can't be parsed as a 
+    /// CSS <color>, a SYNTAX_ERR is raised.
+    /// 
+    /// [(JavaScript docs)](https://developer.mozilla.org/en-US/docs/Web/API/CanvasGradient/addColorStop)
+    // https://html.spec.whatwg.org/#2dcontext:dom-canvasgradient-addcolorstop
+    pub fn add_color_stop(&self, offset: f64, color: &str) -> Result<(), SyntaxError> {
```
koute: This also can throw `IndexSizeError` and yet it's marked as only returning a `SyntaxError`.

You should probably do something like this:

```rust
error_enum_boilerplate! {
    AddColorStopError,
    SyntaxError, IndexSizeError
}
```

where `SyntaxError` should be imported from `webapi::dom_exception` and `IndexSizeError` should be added to `webapi::dom_exception`.

src/webapi/rendering_context.rs
```diff
@@ -18,6 +20,160 @@ pub trait RenderingContext {
...
+
+error_enum_boilerplate! {
+    NotSupportedError,
+    SyntaxError
+}
+
+error_enum_boilerplate! {
+    TypeMismatchError,
+    SyntaxError
+}
```
koute: These aren't correct from what I can see. (: 

The idea of the `error_enum_boilerplate!` is to define a localized error enum which encompasses all of the errors which can be thrown by the original API.

```
error_enum_boilerplace! {
    <NameOfTheLocalErrorEnum>,
    <Error1>, <Error2>, <Error3>, ...
```

src/webapi/rendering_context.rs
```diff
@@ -27,7 +183,771 @@ impl RenderingContext for CanvasRenderingContext2d {
...
+    }
+}
+
+impl ImageData {
+
+    /// Returns a Uint8ClampedArray representing a one-dimensional array containing the data in the RGBA order, 
+    /// with integer values between 0 and 255 (included).
+    /// 
+    /// [(JavaScript docs)](https://developer.mozilla.org/en-US/docs/Web/API/ImageData/data)
+    pub fn get_data(&self) -> Vec<u8> {
```
koute: This should return `Uint8ClampedArray`, not `Vec<u8>`.

Right now we don't have any way of expressing `Uint8ClampedArray`. We have `TypedArray< u8 >`, but that's a non-clapmed variant. So until we have that you can temporairly comment this function out, leave a TODO and we can fix it later after we merge this PR.

src/webapi/rendering_context.rs
```diff
@@ -27,7 +183,771 @@ impl RenderingContext for CanvasRenderingContext2d {
...
+        js! (
+            return @{&self.0}.data;
+        ).try_into().unwrap()
+    }
+
+    /// Returns the number of rows in the image data object.
+    /// 
+    /// [(JavaScript docs)](https://developer.mozilla.org/en-US/docs/Web/API/ImageData/height)
+    pub fn get_height(&self) -> f64 {
```
koute: 1. These are missing links to the specs, e.g. for this method specifically `https://html.spec.whatwg.org/#2dcontext:dom-imagedata-width`.

2. In the specs this has type of `unsigned long`, so it should return `u32` instead of `f64`. Ditto for other methods. (This is why a link to the specs it useful - you can quickly check whenever the types are correct.)

src/webapi/rendering_context.rs
```diff
@@ -27,7 +183,771 @@ impl RenderingContext for CanvasRenderingContext2d {
...
+        js! { @(no_return)
+            @{&self.0}.fillStyle = @{color};
+        }
+    }
+
+    /// The CanvasRenderingContext2D.fillStyle property of the Canvas 2D API specifies the color or style to use inside shapes. 
+    /// The default is #000 (black).
+    /// 
+    /// [(JavaScript docs)](https://developer.mozilla.org/en-US/docs/Web/API/CanvasRenderingContext2D/fillStyle)
+    pub fn set_fill_style_gradient(&self, gradient: CanvasGradient){
```
koute: In cases like this it'd be better to take `&CanvasGradient` instead of `CanvasGradient`. (Otherwise the user will have to `clone()` the argument every time they call this function.)

src/webapi/rendering_context.rs
```diff
@@ -27,7 +183,771 @@ impl RenderingContext for CanvasRenderingContext2d {
...
+            "color-burn" => CompositeOperation::ColorBurn,
+            "hard-light" => CompositeOperation::HardLight,
+            "soft-light" => CompositeOperation::SoftLight,
+            "difference" => CompositeOperation::Difference,
+            "exclusion" => CompositeOperation::Exclusion,
+            "hue" => CompositeOperation::Hue,
+            "saturation" => CompositeOperation::Saturation,
+            "color" => CompositeOperation::Color,
+            "luminosity" => CompositeOperation::Luminosity,
+            _ => panic!("Unexpected globalCompositeOperation value"),
```
koute: It'd be nice to print the value in the `panic!` message so *if* this happens we'll know what to add.

src/webapi/rendering_context.rs
```diff
@@ -27,7 +183,771 @@ impl RenderingContext for CanvasRenderingContext2d {
...
+    /// [(JavaScript docs)](https://developer.mozilla.org/en-US/docs/Web/API/CanvasRenderingContext2D/drawFocusIfNeeded)
+    // https://html.spec.whatwg.org/#2dcontext:dom-context-2d-drawfocusifneeded
+    pub fn draw_focus_if_needed< T: IHtmlElement >(&self, element: &T) {
+        js! { @(no_return)
+            @{&self.0}.drawFocusIfNeeded(@{element.as_ref()});
+        }
+    }
+
+    /// Provides different ways to draw an image onto the canvas.
+    /// TODO: Potentially throw more than just TypeMismatchError
```
koute: `TODO`s should probably be left in normal `//` comments; everything which starts with `///` will end up in the user-visible documentation. (:

src/webapi/rendering_context.rs
```diff
@@ -58,4 +978,288 @@ impl CanvasRenderingContext2d {
...
+        js! { @(no_return)
+            @{&self.0}.lineTo(@{x}, @{y});
+        }
+    }
+
+    /// Returns a TextMetrics object that contains information about the measured text (such as its width for example).
+    /// 
+    /// [(JavaScript docs)](https://developer.mozilla.org/en-US/docs/Web/API/CanvasRenderingContext2D/measureText)
+    // https://html.spec.whatwg.org/#2dcontext:dom-context-2d-measuretext
+    pub fn measure_text(&self, text: &str) -> TextMetrics {
```
koute: This can throw a `SecurityError` according to the specs.

koute:
This is starting to look really nice!

A few general remarks based on what I've found:

1. Please add links to the specs for everything you've added. I realize that is super tedious (sorry!), but it makes it a lot easier to review (and will enable us to automatically check our bindings for completeness in the future.)
2. Please go through every method and double check that the types are correct, e.g. if the specs says that a method takes or returns an `unsigned long` then it should have a type of `u32`, not `f64`. If a given method's type cannot be currently expressed (e.g. the clamped typed array case) then it's better to not bind the function at all instead of doing it with incorrect type.
3. Please make sure that every method which can throw will return a `Result`. Not every case in documented on the MDN, so it's best to just see what the specs says.
4. Some tests would be nice if you're feeling up to it (e.g. a few tests based on the jsfiddle you posted where you'd test that a given API really returns an error when it's fed an invalid input), but I'm willing to accept this without tests as those can be added later.

src/webapi/rendering_context.rs
teovoinea: OHHHHHHHHHH I totally misunderstood that macro! I even tried (unknowingly) to re-implement it myself! Makes a lot more sense now :)

teovoinea:
@koute Thanks! Sorry for the massive PR 😂 So I've added links to the specs for both the methods and the properties, while I was doing that I also double checked the types and made sure that all methods that can return an error have a `Result` type. Also, I've added a couple of quick tests mostly just for sanity checking.

src/webapi/rendering_context.rs
```diff
@@ -58,4 +1023,319 @@ impl CanvasRenderingContext2d {
...
+            @{&self.0}.setTransform(@{a}, @{b}, @{c}, @{d}, @{e}, @{f});
+        }
+    }
+
+    /// Strokes the current or given path with the current stroke style using the non-zero winding rule.
+    /// 
+    /// ctx.stroke(path) is not supported because [(Path2D)](https://developer.mozilla.org/en-US/docs/Web/API/Path2D) is still experimental
+    /// 
+    /// [(JavaScript docs)](https://developer.mozilla.org/en-US/docs/Web/API/CanvasRenderingContext2D/stroke)
+    // https://html.spec.whatwg.org/#2dcontext:dom-context-2d-stroke
```
koute: Wrong link.

src/webapi/rendering_context.rs
```diff
@@ -58,4 +1023,319 @@ impl CanvasRenderingContext2d {
...
+    /// 
+    /// [(JavaScript docs)](https://developer.mozilla.org/en-US/docs/Web/API/CanvasRenderingContext2D/translate)
+    // https://html.spec.whatwg.org/#2dcontext:dom-context-2d-translate
+    pub fn translate(&self, x: f64, y: f64) {
+        js! { @(no_return)
+            @{&self.0}.translate(@{x}, @{y});
+        }
+    }
+
+    fn fill_rule_to_str(&self, fill_rule: FillRule) -> &'static str {
```
koute: This should be a free function.

src/webapi/rendering_context.rs
```diff
@@ -58,4 +1023,319 @@ impl CanvasRenderingContext2d {
...
+#[cfg(all(test, feature = "web_test"))]
+mod test {
+    use super::*;
+    use webcore::try_from::TryInto;
+
+    fn new_canvas() -> CanvasRenderingContext2d {
+        js!(
+            var canvas = document.createElement("canvas");
+            return canvas.getContext("2d");
+        ).try_into().unwrap()
```
koute: This should use `document().create_element` and `CanvasRenderingContext2d::from_canvas`. (No need to use `js!` macro here.)

src/webapi/rendering_context.rs
```diff
@@ -1,7 +1,10 @@
 use webcore::value::{Reference, ConversionError};
-use webcore::try_from::TryInto;
-use webapi::html_elements::CanvasElement;
-
+use webcore::try_from::{TryFrom, TryInto};
+use webcore::value::Undefined;
+use webapi::html_elements::{CanvasElement, ImageElement};
+use webapi::html_element::IHtmlElement;
+//use webapi::typed_array::TypedArray;
```
koute: You can remove this commented-out line.

src/webapi/rendering_context.rs
```diff
@@ -27,7 +190,809 @@ impl RenderingContext for CanvasRenderingContext2d {
...
+        ).unwrap()
+    }
+}
+
+impl ImageData {
+
+    /// Returns a Uint8ClampedArray representing a one-dimensional array containing the data in the RGBA order, 
+    /// with integer values between 0 and 255 (included).
+    /// 
+    /// [(JavaScript docs)](https://developer.mozilla.org/en-US/docs/Web/API/ImageData/data)
```
koute: The doc comments should also be inside the `/* */`

src/webapi/rendering_context.rs
```diff
@@ -27,7 +190,809 @@ impl RenderingContext for CanvasRenderingContext2d {
...
+
 impl CanvasRenderingContext2d {
+    /// # Properties
+    
+    /// The CanvasRenderingContext2D.canvas property is a read-only reference to the HTMLCanvasElement 
+    /// object that is associated with the context. It might be null if there is no association with an <canvas> element.
+    /// 
+    /// [(JavaScript docs)](https://developer.mozilla.org/en-US/docs/Web/API/CanvasRenderingContext2D/canvas)
+    // https://html.spec.whatwg.org/#2dcontext:dom-context-2d-canvas
+    pub fn get_canvas(&self) -> Option<CanvasElement> {
```
koute: I know the MDN says it can return `null`, but in the specs we don't have a `?` after the type, so, are you sure it can return null? (If yes then adding a simple `#[test]` which demonstrates that would be nice.)

src/webapi/rendering_context.rs
```diff
@@ -27,7 +190,809 @@ impl RenderingContext for CanvasRenderingContext2d {
...
+            return @{&self.0}.createImageData(@{image_data});
+        ).try_into().unwrap()
+    }
+
+    /// Creates a pattern using the specified image (a CanvasImageSource). It repeats the source in 
+    /// the directions specified by the repetition argument. This method returns a CanvasPattern.
+    /// 
+    /// [(JavaScript docs)](https://developer.mozilla.org/en-US/docs/Web/API/CanvasRenderingContext2D/createPattern)
+    // https://html.spec.whatwg.org/#2dcontext:dom-context-2d-createpattern
+    pub fn create_pattern_image(&self, image: ImageElement, repetition: Option<Repetition>) -> Result<CanvasPattern, SyntaxError> {
```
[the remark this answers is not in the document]
koute: Also, since the `Repetition` has a well defined default value perhaps it would be a good idea to implement `Default` for it and make this method take `Repetition` instead of `Option<Repetition>`? (Then you would call it `create_pattern_image(image, Repetition::default())`.)

Ditto for methods which take a `FillRule` (where fill rule is always "nonzero" by default from what I can see.)

koute:
Thanks again for the great work!

Oh and, you might want to rebase your PR as there are currently some merge conflicts. (You might have to squash all of your commits before you do that.)

src/webapi/rendering_context.rs
```diff
@@ -27,7 +190,809 @@ impl RenderingContext for CanvasRenderingContext2d {
...
+            return @{&self.0}.createImageData(@{width}, @{height});
+        ).unwrap()
+    }
+
+    /// Creates a new, blank ImageData object with the specified dimensions. 
+    /// All of the pixels in the new object are transparent black.
+    /// 
+    /// [(JavaScript docs)](https://developer.mozilla.org/en-US/docs/Web/API/CanvasRenderingContext2D/createImageData)
+    // https://html.spec.whatwg.org/#2dcontext:dom-context-2d-createimagedata
+    pub fn create_image_data_id(&self, image_data: ImageData) -> ImageData {
```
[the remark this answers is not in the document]
koute: Hmm... I guess I like `create_image_data_size_of` more.

src/webapi/rendering_context.rs
```diff
@@ -27,7 +190,809 @@ impl RenderingContext for CanvasRenderingContext2d {
...
+            @{&self.0}.drawImage(@{image}, @{sx}, @{sy}, @{s_width}, @{s_height}, @{dx}, @{dy}, @{d_width}, @{d_height});
+        ).unwrap()
+    }
+
+    /// Fills the current or given path with the current fill style using the non-zero or even-odd winding rule.
+    /// 
+    /// ctx.fill(path, fillRule) is not supported because [(Path2D)](https://developer.mozilla.org/en-US/docs/Web/API/Path2D) is still experimental
+    /// 
+    /// [(JavaScript docs)](https://developer.mozilla.org/en-US/docs/Web/API/CanvasRenderingContext2D/fill)
+    // https://html.spec.whatwg.org/#2dcontext:dom-context-2d-fill
```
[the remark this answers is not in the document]
teovoinea: Are you sure? This function and the stroke function seem to be linking correctly for me

src/webapi/rendering_context.rs
teovoinea: I wasn't able to produce a null value from accessing the `canvas` property.

teovoinea:
Implemented those last few fixes and up to date with the latest commits!

Also, I think there's an option to squash on github when you accept the merge request :)

teovoinea:
@koute Did you have any more comments? :)

koute:
I haven't had the chance to go through it again, but will very soon! Sorry for the delay!

Hmm.... I'm not really sure how the squash through Github will work in presence of merge conflicts? I can try doing it when I'll be merging this it, though it would *probably* be better to rebase it manually (even without squashing).

src/webapi/rendering_context.rs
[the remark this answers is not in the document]
koute: Ah, you're right. Sorry! Firefox autocompleted the `...fill` into `...fillstyle` and I didn't notice... -_-

koute:
Although something might have slipped through cracks in a PR as big as this one it looks good to me now!

Thank you!

teovoinea:
Thanks a lot! I was thinking of making a sort of guideline for implementing web APIs since I’m seeing a few PRs where it could save a lot of duplicate comments

koute:
@teovoinea If you feel like it I would certainly love to see such a thing! (:

Issue #92, opened 2018-01-24
Some WebGL objects cannot put in the weakmap in firefox


edwin0cheng:
I am trying to using WebGL in rust by stdweb, however, i found that
some WebGL related objects (WebGLBuffer, WebGLUniformLocation, etcs) could not put in the WeakMap in firefox (Chrome do not have this problem).
 
Do you think there are any alternative solutions instead of current WeakMap implementation ?

koute:
That's... interesting. Hmm... possibly.

Have you tried to see how Emscripten's OpenGL emulation layer handles this?

edwin0cheng:
Just checked, basically, emscripten WebGL Client use a big object table to store  all objects :

https://github.com/kripken/emscripten/blob/7f89560101843198787530731f40a65288f6f15f/src/webGLClient.js

You can search "getUniformLocation" and "uniform4fv" to understand how it works...
Maybe i would need to do something like these by myself.  T T

edwin0cheng:
Oh seem like you mis-understand what i say.

When using js! macro to return an object to rust side, stdweb are using WeakMap to store the object.
However, firefox is not allowed to store these objects in WeakMap, so stdweb should fail at that point.

Diggsey:
AIUI, the only reason we need a weak map is to ensure that if the same object is exposed to rust multiple times, it gets the same ID each time - could we just drop that guarantee? (Use a strong map and remove items when their refcount goes to zero)

edwin0cheng:
As i understand, use a strong map should have the same ID each time too.  
According to [MDN](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Map), Object as keys are considered equal according to the semantics of the === operator which is only equal compare to itself.

And Maybe we do not need a WeakMap here as we have keep track the refcount itself? Do i misunderstand something?

Diggsey:
@edwin0cheng yes, the point of using the weak map is so that items stay in the *weak* map even after their reference count reaches zero, as long as the object is still in use by javascript.

If we switch to a strong map, then to avoid memory leaks we would have to remove it form the strong map when the reference count reaches zero: that means that if the object is kept alive on the javascript side, and then passed into rust again at a later time, the later use will receive a different ID.

edwin0cheng:
Oh... you are right, seem like i totally misunderstand that part.

[...]

PR #97, opened 2018-01-28
Add webgl example

Diggsey:
image

This is blocked on https://github.com/brendanzab/gl-rs/pull/447, which is blocked on publishing a new version of stdweb.


koute:
This looks really cool, thanks!

I'll try to get a new `stdweb` published in the next few days once I'll finish what I'm doing now.

Diggsey:
Live example here: http://brave-bell-e62dc4.bitballoon.com/

koute:
I just took a look at the code produced by your generator; could we make so that it doesn't use anything from `stdweb::private` nor `__js_raw_asm!`?

From a cursory look at it I'd guess you need a way to define reference types outside of `stdweb`. Anything else?

Diggsey:
Yeah I use quite a lot of internal stuff right now.

From `stdweb::private`:
```
JsSerializable
FromReferenceUnchecked
FromReference
PreallocatedArena
SerializedValue
```

From `stdweb::unstable`:
```
TryFrom
TryInto
Void
```

Everything from private except `JsSerializable` is used for defining reference types (and making them serialisable). `JsSerializable` is also used as a trait bound when accepting an `any` WebIDL type as a parameter.

koute:
@Diggsey I've added an `stdweb-derive` crate which should make it possible to define reference-like types externally outside of the `stdweb`. Can you take it for a spin and tell me if it works fine for you/whenever there is anything missing?

Diggsey:
Will do - what about `JsSerializable`?

koute:
It's still private. You can maybe try to use `ReferenceType` in the bounds, and then use `as_ref` to convert it to a `Reference` and serialize that? (If not then we'll figure something else out.)

Diggsey:
It's not necessarily a reference type - this is for `any` so it should accept any type that can be used in the `js!` macro.

koute:
Ok, so I've exposed a `JsSerialize` trait which you can use in the trait bounds. (But everything inside of it is still private!)

Diggsey:
Cool, I've updated both PRs with those changes.

See here for the generated bindings (they're ~4000 lines shorter now 😄 ): https://gist.github.com/Diggsey/c8a4cb93c7d5643d20203401ead178a7

The only internal thing I am using now is `__js_raw_asm`.

koute:
Nice!

Hmmm... could you switch to the `js!` macro for now? That will be very slightly slightly slower, but we can make it as fast as `__js_raw_asm` later. (Either by making `js!` itself faster [wouldn't be hard if only procedural macros were stable, eh...], or by adding an extra attribute similar to `#[reference(instance_of = "...")]` which would allow you to compare an arbitrary field to an arbitrary value.)

koute:
@Diggsey Ping. Any progress? (:

Diggsey:
Still waiting on the upstream PR :/

Diggsey:
@koute one possibility to progress this without depending on `webgl_generator` getting merged is to simply commit the generated code - we can always swap it out later. What do you think?

Diggsey:
@koute the upstream PR was merged, the new crate was published and I've updated this PR to use that crate. It should be ready to go now!

kuviman:
@Diggsey this is affected by #92, is there a workaround?

Diggsey:
No, #92 just needs to be fixed.

anlumo:
In my own (JavaScript) code, I worked around that Firefox bug by just wrapping the WebGL objects in JavaScript objects (like `{texture: texture}`). That might be a bit harder to do here, though.

Diggsey:
@koute can this be merged?

koute:
Can you add this new example to `ci/run_tests.sh` so that it's going to be built on Travis? Otherwise LGTM!

PR #100, opened 2018-01-31
Add rustup link to README.md

jdanford:
Also fixed a typo 🙂


koute:
Unfortunately I'm just about to remove that section from the README since `cargo-web` installs these targets automatically now. (:

But I'll take the typo fix! Thanks!

PR #103, opened 2018-02-03
Adding in Promises

### LINKED ISSUE #72: Promise support
^ Title

Pauan:
Fixes #72

@koute @CryZe This is untested and is definitely **not** ready to merge.

But I wanted to at least get started on adding in Promise support to stdweb. This pull request adds in the ability to convert any JS Promise into a Rustified Futurified version of that Promise. This conversion happens automatically with the `TryFrom<Value>` trait.

Please look this over and let me know if you see anything weird, or anything that can be improved.

### DIFF
--- a/src/webapi/error.rs
+++ b/src/webapi/error.rs
@@ -37,7 +37,9 @@ pub trait IError: ReferenceType {
 pub struct Error( Reference );
 
 impl Error {
+    /// Creates a new `Error` with the specified `description`.
     ///
+    /// [(JavaScript docs)](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Error)
     #[inline]
     pub fn new( description: &str ) -> Self {
         js!( return new Error( @{description} ); ).try_into().unwrap()
--- a/src/webapi/error.rs
+++ b/src/webapi/error.rs
@@ -44,6 +44,16 @@ impl Error {
     pub fn new( description: &str ) -> Self {
         js!( return new Error( @{description} ); ).try_into().unwrap()
     }
+
+    /// Prints the `Error` to the console (this prints the error's description and also its stack trace).
+    ///
+    /// [(JavaScript docs)](https://developer.mozilla.org/en-US/docs/Web/API/Console/error)
+    #[inline]
+    pub fn print( &self ) {
+        js! { @(no_return)
+            console.error( @{self} );
+        }
+    }
 }
 
 // Error specification:
--- a/src/webapi/error.rs
+++ b/src/webapi/error.rs
@@ -36,6 +36,13 @@ pub trait IError: ReferenceType {
 #[reference(instance_of = "Error")]
 pub struct Error( Reference );
 
+impl Error {
+    #[inline]
+    pub fn new( description: &str ) -> Self {
+        js!( return new Error( @{description} ); ).try_into().unwrap()
+    }
+}
+
 // Error specification:
 // https://www.ecma-international.org/ecma-262/6.0/#sec-error-objects
 
--- a/src/webapi/error.rs
+++ b/src/webapi/error.rs
@@ -37,6 +37,7 @@ pub trait IError: ReferenceType {
 pub struct Error( Reference );
 
 impl Error {
+    ///
     #[inline]
     pub fn new( description: &str ) -> Self {
         js!( return new Error( @{description} ); ).try_into().unwrap()
--- a/src/webcore/promise.rs
+++ b/src/webcore/promise.rs
@@ -157,7 +157,6 @@ pub struct PromiseFuture< A > {
     phantom: PhantomData< A >,
 }
 
-
 impl PromiseFuture< () > {
     /// Asynchronously runs the [`Future`](https://docs.rs/futures/0.1.18/futures/future/trait.Future.html) and immediately returns. This does not block the current thread.
     ///
@@ -187,14 +186,6 @@ impl PromiseFuture< () > {
         } ) );
     }
 }
-
-/*impl< A > PromiseFuture< A > {
-    pub fn new< B >( callback: B ) -> Self
-        where B: FnOnce( FnOnce( A ), FnOnce( JSError ) ) {
-        js!( return new Promise( @{Once( callback )} ); ).try_from().unwrap()
-    }
-}*/
-
 
 impl< A > std::fmt::Debug for PromiseFuture< A > {
     fn fmt( &self, formatter: &mut std::fmt::Formatter ) -> std::fmt::Result {
--- a/src/webcore/promise.rs
+++ b/src/webcore/promise.rs
@@ -173,9 +173,10 @@ impl PromiseFuture< () > {
     ///
     /// ```rust
     /// fn main() {
-    ///     create_some_future()
-    ///         .inspect(|x| println!("Future finished: {:#?}", x))
-    ///         .spawn()
+    ///     PromiseFuture::spawn(
+    ///         create_some_future()
+    ///             .inspect(|x| println!("Future finished: {:#?}", x))
+    ///     )
     /// }
     /// ```
     pub fn spawn< B >( future: B ) where
--- a/src/webcore/promise.rs
+++ b/src/webcore/promise.rs
@@ -1,0 +1,79 @@
+use std;
+use std::marker::PhantomData;
+use {Value, Once};
+use unstable::{TryInto, TryFrom};
+use webcore::value::ConversionError;
+use web::error::Error as JSError;
+use std::error::Error;
+use futures::{future, Future, Poll};
+use futures::sync::oneshot::channel;
+
+
+// TODO split this into Promise and PromiseFuture
+pub struct PromiseFuture< A > {
+    promise: Value,
+    future: Box< Future< Item = A, Error = JSError > >,
+    phantom: PhantomData< A >,
+}
+
+
+impl< A > std::fmt::Debug for PromiseFuture< A > {
+    fn fmt( &self, formatter: &mut std::fmt::Formatter ) -> std::fmt::Result {
+        write!( formatter, "PromiseFuture {:?}", self.promise )
+    }
+}
+
+
+impl< A > Future for PromiseFuture< A > {
+    type Item = A;
+    type Error = JSError;
+
+    fn poll (&mut self ) -> Poll<Self::Item, Self::Error> {
+        self.future.poll()
+    }
+}
+
+
+// TODO this should probably check instanceof Promise
+impl< A: TryFrom< Value > > TryFrom< Value > for PromiseFuture< A > where A: 'static, A::Error: Error {
+    type Error = ConversionError;
+
+    fn try_from( v: Value ) -> Result< Self, Self::Error > {
+        match v {
+            Value::Reference( ref r ) => {
+                let ( sender, receiver ) = channel();
+
+                let callback = |value: Value, success: bool| {
+                    let value: Result< A, JSError > = if success {
+                        let value: Result< A, A::Error > = value.try_into();
+                        value.map_err( |e| JSError::new( e.description() ) )
+                    } else {
+                        let value: Result< JSError, ConversionError > = value.try_into();
+                        value.map_err( |e| JSError::new( e.description() ) ).and_then( Err )
+                    };
+
+                    // TODO is this correct ?
+                    match sender.send( value ) {
+                        Ok( _ ) => {},
+                        Err( _ ) => {},
+                    };
+                };
+
+                Ok( PromiseFuture {
+                    promise: js! {
+                        var callback = @{Once( callback )};
+
+                        return @{r}.then( function (value) {
+                            callback( value, true );
+                        }, function (value) {
+                            callback( value, false );
+                        } );
+                    },
+                    future: Box::new( receiver.map_err( |x| JSError::new( x.description() ) ).and_then( future::result ) ),
+                    phantom: PhantomData,
+                } )
+            },
+            other => Err( ConversionError::Custom( format!( "Expected Promise but got: {:?}", other ) ) ),
+        }
+    }
+}
--- a/src/webcore/promise.rs
+++ b/src/webcore/promise.rs
@@ -17,7 +17,7 @@ reference_boilerplate! {
 }
 
 impl Promise {
-    pub fn done< A, B >( &self, callback: B ) -> Self
+    pub fn done< A, B >( &self, callback: B )
         where A: TryFrom< Value >,
               A::Error: Error,
               B: FnOnce( Result< A, JSError > ) + 'static {
@@ -34,15 +34,16 @@ impl Promise {
             callback( value );
         };
 
-        (js! {
+        js! { @(no_return)
             var callback = @{Once( callback )};
 
-            return @{self}.then( function (value) {
+            // TODO don't swallow any errors thrown inside callback
+            @{self}.then( function (value) {
                 callback( value, true );
             }, function (value) {
                 callback( value, false );
             } );
-        }).try_into().unwrap()
+        }
     }
 
     // We can't use the IntoFuture trait because Promise doesn't have a type argument
@@ -53,14 +54,15 @@ impl Promise {
 
         let ( sender, receiver ) = channel();
 
+        self.done( |value| {
+            // TODO is this correct ?
+            match sender.send( value ) {
+                Ok( _ ) => {},
+                Err( _ ) => {},
+            };
+        } );
+
         PromiseFuture {
-            promise: self.done( |value| {
-                // TODO is this correct ?
-                match sender.send( value ) {
-                    Ok( _ ) => {},
-                    Err( _ ) => {},
-                };
-            } ),
             future: Box::new( receiver.map_err( |x| JSError::new( x.description() ) ).and_then( future::result ) ),
             phantom: PhantomData,
         }
@@ -69,7 +71,6 @@ impl Promise {
 
 
 pub struct PromiseFuture< A > {
-    promise: Promise,
     future: Box< Future< Item = A, Error = JSError > >,
     phantom: PhantomData< A >,
 }
@@ -85,7 +86,7 @@ pub struct PromiseFuture< A > {
 
 impl< A > std::fmt::Debug for PromiseFuture< A > {
     fn fmt( &self, formatter: &mut std::fmt::Formatter ) -> std::fmt::Result {
-        write!( formatter, "PromiseFuture {:?}", self.promise )
+        write!( formatter, "PromiseFuture" )
     }
 }
 
--- a/src/webcore/promise.rs
+++ b/src/webcore/promise.rs
@@ -15,6 +15,7 @@ use webcore::promise_executor::spawn;
 /// In most situations you shouldn't use this, use [`PromiseFuture`](struct.PromiseFuture.html) instead.
 ///
 /// [(JavaScript docs)](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise)
+// https://www.ecma-international.org/ecma-262/6.0/#sec-promise-objects
 pub struct Promise( Reference );
 
 reference_boilerplate! {
@@ -53,6 +54,9 @@ impl Promise {
     /// ```
     ///
     /// [(JavaScript docs)](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise/resolve)
+    // https://www.ecma-international.org/ecma-262/6.0/#sec-promise.resolve
+    // https://www.ecma-international.org/ecma-262/6.0/#sec-promise-resolve-functions
+    // https://www.ecma-international.org/ecma-262/6.0/#sec-promiseresolvethenablejob
     pub fn promisify( input: Value ) -> Promise {
         js!( return Promise.resolve( @{input} ); ).try_into().unwrap()
     }
@@ -79,6 +83,7 @@ impl Promise {
     /// ```
     ///
     /// [(JavaScript docs)](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise/then)
+    // https://www.ecma-international.org/ecma-262/6.0/#sec-performpromisethen
     pub fn done< A, B >( &self, callback: B )
         where A: TryFrom< Value >,
               A::Error: std::error::Error,
--- a/src/webcore/promise.rs
+++ b/src/webcore/promise.rs
@@ -28,7 +28,7 @@ impl< A > Future for PromiseFuture< A > {
     type Item = A;
     type Error = JSError;
 
-    fn poll (&mut self ) -> Poll<Self::Item, Self::Error> {
+    fn poll (&mut self ) -> Poll< Self::Item, Self::Error > {
         self.future.poll()
     }
 }
--- a/src/webcore/promise.rs
+++ b/src/webcore/promise.rs
@@ -113,9 +113,9 @@ impl Promise {
             var callback = @{Once( callback )};
 
             // TODO don't swallow any errors thrown inside callback
-            @{self}.then( function (value) {
+            @{self}.then( function ( value ) {
                 callback( value, true );
-            }, function (value) {
+            }, function ( value ) {
                 callback( value, false );
             } );
         }
--- a/src/webcore/promise.rs
+++ b/src/webcore/promise.rs
@@ -16,12 +16,9 @@ use webcore::promise_executor::spawn;
 ///
 /// [(JavaScript docs)](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise)
 // https://www.ecma-international.org/ecma-262/6.0/#sec-promise-objects
+#[derive(Clone, Debug, ReferenceType)]
+#[reference(instance_of = "Promise")]
 pub struct Promise( Reference );
-
-reference_boilerplate! {
-    Promise,
-    instanceof Promise
-}
 
 impl Promise {
     // https://www.ecma-international.org/ecma-262/6.0/#sec-promise-resolve-functions
--- a/src/webcore/promise.rs
+++ b/src/webcore/promise.rs
@@ -169,6 +169,9 @@ impl Promise {
 /// Convert a JavaScript `Promise` into a `PromiseFuture`:
 ///
 /// ```rust
+/// use stdweb::PromiseFuture;
+/// use stdweb::web::error::Error;
+///
 /// let future: PromiseFuture<String, Error> = js!( return Promise.resolve("foo"); ).try_into().unwrap();
 /// ```
 pub struct PromiseFuture< A, B > {
--- a/src/webcore/promise.rs
+++ b/src/webcore/promise.rs
@@ -6,7 +6,7 @@ use webcore::value::{Value, Reference, ConversionError};
 use webcore::try_from::{TryInto, TryFrom};
 use web::error::Error as JSError;
 use futures::{future, Future, Poll};
-use futures::sync::oneshot::channel;
+use futures::unsync::oneshot::channel;
 
 
 pub struct Promise( Reference );
--- a/src/webcore/promise.rs
+++ b/src/webcore/promise.rs
@@ -10,7 +10,11 @@ use futures::unsync::oneshot::{Receiver, channel};
 use webcore::promise_executor::spawn;
 
 
+/// A `Promise` object represents the eventual completion (or failure) of an asynchronous operation, and its resulting value.
 ///
+/// In most situations you shouldn't use this, use [`PromiseFuture`](struct.PromiseFuture.html) instead.
+///
+/// [(JavaScript docs)](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise)
 pub struct Promise( Reference );
 
 reference_boilerplate! {
@@ -19,12 +23,62 @@ reference_boilerplate! {
 }
 
 impl Promise {
+    /// This function should rarely be needed, use [`PromiseFuture`](struct.PromiseFuture.html) instead.
     ///
+    /// This function is used for two different purposes:
+    ///
+    /// 1. If you have a JavaScript value which is not a `Promise` but you want to wrap it in a `Promise`, you can use `Promise::promisify(value)`.
+    ///    In this situation, it is recommended to use [`futures::future::ok`](https://docs.rs/futures/0.1.18/futures/future/fn.ok.html) instead.
+    ///
+    /// 2. If you have a JavaScript value which is a Promise-like object (it has a `then` method) but it isn't a true `Promise`, you can use
+    ///    `Promise::promisify(value)` to convert it into a true `Promise`. This situation is rare, but it can happen if you are using a Promise
+    ///    library such as jQuery or Bluebird.
+    ///
+    /// # Examples
+    ///
+    /// Convert a JavaScript value to a `Promise`:
+    ///
+    /// ```rust
+    /// Promise::promisify(js!( return 5; ))
+    /// ```
+    ///
+    /// Convert a Promise-like object to a `Promise`:
+    ///
+    /// ```rust
+    /// // jQuery Promise
+    /// Promise::promisify(js!( return $.get("test.php"); ))
+    ///
+    /// // Bluebird Promise
+    /// Promise::promisify(js!( return bluebird_promise.timeout(1000); ))
+    /// ```
+    ///
+    /// [(JavaScript docs)](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise/resolve)
     pub fn promisify( input: Value ) -> Promise {
         js!( return Promise.resolve( @{input} ); ).try_into().unwrap()
     }
 
+    /// This method is usually not needed, use [`PromiseFuture`](struct.PromiseFuture.html) instead.
     ///
+    /// When the `Promise` either succeeds or fails, it calls the `callback` with the result.
+    ///
+    /// It does not wait for the `Promise` to succeed / fail (it does not block the thread).
+    ///
+    /// The `callback` is guaranteed to be called asynchronously even if the `Promise` is already succeeded / failed.
+    ///
+    /// If the `Promise` never succeeds / fails then the `callback` will never be called, and it will leak memory.
+    ///
+    /// # Examples
+    ///
+    /// ```rust
+    /// promise.done(|result| {
+    ///     match result {
+    ///         Ok(success) => { ... },
+    ///         Err(error) => { ... },
+    ///     }
+    /// });
+    /// ```
+    ///
+    /// [(JavaScript docs)](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise/then)
     pub fn done< A, B >( &self, callback: B )
         where A: TryFrom< Value >,
               A::Error: Error,
@@ -54,7 +108,15 @@ impl Promise {
         }
     }
 
+    /// This method should rarely be needed, instead use [`value.try_into()`](unstable/trait.TryInto.html) to convert directly from a [`Value`](enum.Value.html) into a [`PromiseFuture`](struct.PromiseFuture.html).
     ///
+    /// This method converts the `Promise` into a [`PromiseFuture`](struct.PromiseFuture.html), so that it can be used as a Rust [`Future`](https://docs.rs/futures/0.1.18/futures/future/trait.Future.html).
+    ///
+    /// # Examples
+    ///
+    /// ```rust
+    /// promise.to_future().map(|x| x + 1)
+    /// ```
     // We can't use the IntoFuture trait because Promise doesn't have a type argument
     // TODO explain more why we can't use the IntoFuture trait
     pub fn to_future< A >( &self ) -> PromiseFuture< A >
@@ -79,7 +141,17 @@ impl Promise {
 }
 
 
+/// This allows you to use a JavaScript [`Promise`](struct.Promise.html) as if it is a Rust [`Future`](https://docs.rs/futures/0.1.18/futures/future/trait.Future.html).
 ///
+/// The preferred way to create a `PromiseFuture` is to use [`value.try_into()`](unstable/trait.TryInto.html) on a JavaScript [`Value`](enum.Value.html).
+///
+/// # Examples
+///
+/// Convert a JavaScript `Promise` into a `PromiseFuture`:
+///
+/// ```rust
+/// let future: PromiseFuture<String> = js!( return Promise.resolve("foo"); ).try_into().unwrap();
+/// ```
 pub struct PromiseFuture< A > {
     future: Receiver< Result< A, JSError > >,
     phantom: PhantomData< A >,
@@ -87,7 +159,21 @@ pub struct PromiseFuture< A > {
 
 
 impl PromiseFuture< () > {
+    /// Asynchronously runs the [`Future`](https://docs.rs/futures/0.1.18/futures/future/trait.Future.html) and immediately returns. This does not block the current thread.
     ///
+    /// If the [`Future`](https://docs.rs/futures/0.1.18/futures/future/trait.Future.html) errors it will print the error to the console.
+    ///
+    /// This function is normally called once in `main`, it is usually not needed to call it multiple times.
+    ///
+    /// # Examples
+    ///
+    /// ```rust
+    /// fn main() {
+    ///     create_some_future()
+    ///         .inspect(|x| println!("Future finished: {:#?}", x))
+    ///         .spawn()
+    /// }
+    /// ```
     pub fn spawn< B >( future: B ) where
         B: Future< Item = (), Error = JSError > + 'static {
 
--- a/src/webcore/promise.rs
+++ b/src/webcore/promise.rs
@@ -1,6 +1,4 @@
 use std;
-use std::error::Error as _Error;
-use std::marker::PhantomData;
 use webcore::once::Once;
 use webcore::value::{Value, Reference, ConversionError};
 use webcore::try_from::{TryInto, TryFrom};
@@ -92,18 +90,23 @@ impl Promise {
     ///
     /// [(JavaScript docs)](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise/then)
     // https://www.ecma-international.org/ecma-262/6.0/#sec-performpromisethen
-    pub fn done< A, B >( &self, callback: B )
+    pub fn done< A, B, F >( &self, callback: F )
         where A: TryFrom< Value >,
-              A::Error: std::error::Error,
-              B: FnOnce( Result< A, Error > ) + 'static {
+              B: TryFrom< Value >,
+              // TODO these Debug constraints are only needed because of unwrap
+              A::Error: std::fmt::Debug,
+              B::Error: std::fmt::Debug,
+              F: FnOnce( Result< A, B > ) + 'static {
 
         let callback = |value: Value, success: bool| {
-            let value: Result< A, Error > = if success {
-                let value: Result< A, A::Error > = value.try_into();
-                value.map_err( |e| Error::new( e.description() ) )
+            let value: Result< A, B > = if success {
+                // TODO figure out a way to avoid the unwrap
+                let value: A = value.try_into().unwrap();
+                Ok( value )
             } else {
-                let value: Result< Error, ConversionError > = value.try_into();
-                value.map_err( |e| Error::new( e.description() ) ).and_then( Err )
+                // TODO figure out a way to avoid the unwrap
+                let value: B = value.try_into().unwrap();
+                Err( value )
             };
 
             callback( value );
@@ -132,9 +135,12 @@ impl Promise {
     /// ```
     // We can't use the IntoFuture trait because Promise doesn't have a type argument
     // TODO explain more why we can't use the IntoFuture trait
-    pub fn to_future< A >( &self ) -> PromiseFuture< A >
+    pub fn to_future< A, B >( &self ) -> PromiseFuture< A, B >
          where A: TryFrom< Value > + 'static,
-               A::Error: std::error::Error {
+               B: TryFrom< Value > + 'static,
+               // TODO remove these later
+               A::Error: std::fmt::Debug,
+               B::Error: std::fmt::Debug {
 
         let ( sender, receiver ) = channel();
 
@@ -148,7 +154,6 @@ impl Promise {
 
         PromiseFuture {
             future: receiver,
-            phantom: PhantomData,
         }
     }
 }
@@ -163,14 +168,13 @@ impl Promise {
 /// Convert a JavaScript `Promise` into a `PromiseFuture`:
 ///
 /// ```rust
-/// let future: PromiseFuture<String> = js!( return Promise.resolve("foo"); ).try_into().unwrap();
+/// let future: PromiseFuture<String, Error> = js!( return Promise.resolve("foo"); ).try_into().unwrap();
 /// ```
-pub struct PromiseFuture< A > {
-    future: Receiver< Result< A, Error > >,
-    phantom: PhantomData< A >,
+pub struct PromiseFuture< A, B > {
+    future: Receiver< Result< A, B > >,
 }
 
-impl PromiseFuture< () > {
+impl PromiseFuture< (), () > {
     /// Asynchronously runs the [`Future`](https://docs.rs/futures/0.1.18/futures/future/trait.Future.html) and then immediately returns.
     /// This does not block the current thread. The only way to retrieve the value of the future is to use the various
     /// [`Future`](https://docs.rs/futures/0.1.18/futures/future/trait.Future.html) methods, such as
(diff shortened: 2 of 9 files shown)


Diggsey:
There are a few problems here:
- You're sending values of type `Result<T, JsError>` over the channel, but then you're expecting to receive a `T` at the other end.
- There's no way to poll the resulting future because there's no reactor implementation that's compatible with stdweb.
- The inner future does not need to be boxed.
- You can use `futures::unsync::onsehot`.

Before jumping into implementing a promises -> futures adaptor, I think it would be useful to implement a futures-compatible reactor on top of the javascript event loop, that way it's possible to test that everything is working.

Pauan:
> @Diggsey You're sending values of type `Result<T, JsError>` over the channel, but then you're expecting to receive a `T` at the other end.

I have no idea what you're talking about. It uses `.and_then( future::result )` to convert the `Result<T, JSError>` into `T`. I may not have tested the code, but it *does* type-check.

> There's no way to poll the resulting future because there's no reactor implementation that's compatible with stdweb.

Yup! That'll need to be added.

> The inner future does not need to be boxed.

Yes it is needed. You can't use `impl Future` in structs, and if you try to spell out the full type it doesn't work because `FnMut` traits do not have a fixed size.

> You can use `futures::unsync::onsehot`.

That's a good idea, I'll do that.

> I think it would be useful to implement a futures-compatible reactor on top of the javascript event loop, that way it's possible to test that everything is working.

No doubt, which is why I said this is **untested** and **not ready to merge**.

Diggsey:
> I have no idea what you're talking about. It uses .and_then( future::result ) to convert the Result<T, JSError> into T. I may not have tested the code, but it does type-check.

Ah yes

> Yes it is needed. You can't use impl Future in structs, and if you try to spell out the full type it doesn't work because FnMut traits do not have a fixed size.

You can just store the `oneshot::Receiver` future as your "inner" future, and do your mapping in the implementation of `poll` - if you're implementing `Future` manually, you don't have to stick to using the combinators all the time.

Pauan:
> @Diggsey You can just store the oneshot::Receiver future as your "inner" future, and do your mapping in the implementation of poll - if you're implementing Future manually, you don't have to stick to using the combinators all the time.

You're right that is possible, but I was concerned about potential performance issues, since it would be rebuilding the Future every time `poll` is called. If the performance is identical, then I agree that moving the combinators into `poll` is a good idea.

As for not using combinators at all, yes that is possible, but it would require `Rc` + `RefCell` shenanigans, which is basically what `channel` is doing internally anyways, so it's nice to be able to re-use existing well-tested battle-hardened APIs rather than rolling my own.

Diggsey:
>  it would be rebuilding the Future every time poll is called.

What? No, take a look at how the future combinators are actually implemented, you would just do this:

```rust
fn poll( &mut self ) -> Poll< Self::Item, Self::Error > {
    self.future.poll().map_err( |x| JSError::new( x.description() ) ).and_then( future::result )
}
```

There's no performance penalty here: this is exactly how `Future::and_then` and `Future::map_err` are implemented.

> As for not using combinators at all, yes that is possible,

I think you are misunderstanding the terminology - combinators are the methods used to compose futures together, eg. `and_then`, `map_err`, etc. They typically take some number of futures/closures as input and produce a new future. They serve as convenience methods to avoid having to manually implement `Future`. In my example above, I am not using the future combinators because I have implemented `poll()` manually using the similar methods on "Result".

Not using `oneshot` at all *would* be an option, but I agree with you that it's not worth it, because you'd be re-implementing much of its functionality.

Pauan:
> @Diggsey What? No, take a look at how the future combinators are actually implemented, you would just do this:

Yes I know what you were referring to.

> There's no performance penalty here

I will verify that the performance of the *compiled binary code* will be just as fast (or faster). I don't judge performance based upon source code.

> I think you are misunderstanding the terminology

I am not misunderstanding anything, I know exactly what "combinator" means in this context.

> I am not using the future combinators because I have implemented poll() manually using the similar methods on "Result".

You are using the combinators *within the implementation of `poll`*. That is what I meant by "not using the combinators at all".

CryZe:
Alright, so I'm not quite sure why we are boxing the oneshot channel's receiver, as there should be absolutely no reason for doing so. The futures I have in stdweb-io don't do any of this either. Additionally you lose all the OIBITs if you use trait objects, so I wouldn't use them unless there's a good reason for them, which I don't quite see yet.

Regardless, most of this PR looks pretty good to me. This is still missing an Executor. You can take a look at the one I made, or I can make a PR after this that brings the Executor.

Pauan:
> @CryZe Additionally you lose all the OIBITs if you use trait objects, so I wouldn't use them unless there's a good reason for them, which I don't quite see yet.

Yes, I definitely don't like using `Box`, I only did it because I had to in order to make it compile.

> This is still missing an Executor. You can take a look at the one I made, or I can make a PR after this that brings the Executor.

Sure, I'll take a deeper look at your code, I'll let you know if I can't figure it out.

----

@Diggsey @CryZe I tried moving the combinators into `poll`, but:

```rust
error[E0631]: type mismatch in function arguments
   |
   |         self.future.poll().map_err( |x| JSError::new( x.description() ) ).and_then( future::result )
   |                                                                           ^^^^^^^^
   |                                                                           |
   |                                                                           expected signature of `fn(futures::Async<std::result::Result<A, JSError>>) -> _`
   |                                                                           found signature of `fn(std::result::Result<_, _>) -> _`
```

```rust
error[E0507]: cannot move out of borrowed content
   |
   |         self.future.map_err( |x| JSError::new( x.description() ) ).and_then( future::result ).poll()
   |         ^^^^ cannot move out of borrowed content
```

It seems there aren't any combinators that will work for this situation, so I decided to just use manual pattern matching:

```rust
fn poll( &mut self ) -> Poll< Self::Item, Self::Error > {
    match self.future.poll() {
        Ok( Async::Ready( Ok( a ) ) ) => Ok( Async::Ready( a ) ),
        Ok( Async::Ready( Err( e ) ) ) => Err( e ),
        Ok( Async::NotReady ) => Ok( Async::NotReady ),
        Err( e ) => Err( JSError::new( e.description() ) ),
    }
}
```

This should be much better and faster than using `Box` + combinators.

Pauan:
I added in an `Executor` and also documentation. I think this is pretty close to being mergeable.

I did some light testing in the `examples/promise` folder, and everything seems to work correctly!

Thanks a lot @CryZe for your `Executor` implementation! I probably wouldn't have been able to create one by myself. I mostly copied your code as-is, but I did a bit of refactoring on it. You can see the code in `webcore/promise_executor.js`

Pauan:
I added in a new `spawn_print` method, which is similar to `spawn` except that it prints the error to the console (if an error occurs).

This is a necessary convenience, because the alternative is to tell the user to handle the errors themself, and it's easy to get error-handling wrong (e.g. accidentally silently discarding errors, printing the errors in a bad format, etc.)

However, I don't like the type of `spawn_print`. It requires the error type to be `stdweb::web::error::Error`, which is fine for `PromiseFuture` (because the error type of `PromiseFuture` is always `stdweb::web::error::Error`), but that won't be true for other Futures.

The reason it requires `stdweb::web::error::Error` is because the browser can pretty print JavaScript errors better than Rust can (browsers display the stack trace, browsers understand source maps, etc.)

I've thought about possibly adding in a new trait which can be used to give special behavior to `stdweb::web::error::Error` while still allowing for other error types to work as well.

Any ideas or suggestions for how to handle this situation are appreciated.

Pauan:
My current preferred solution is to add in this impl:

```rust
impl From< stdweb::web::error::Error > for () {
    #[inline]
    fn from( error: stdweb::web::error::Error ) -> Self {
        js! { @(no_return)
            console.error( @{error} );
        }

        ()
    }
}
```

And now it's possible to use `.from_err()` to convert from `stdweb::web::error::Error` to `()` (with correct error printing):

```rust
PromiseFuture::spawn(
    create_some_future()
        .from_err()
);
```

I don't like that it has side effects in the implementation of `from`, but the alternative would be to create a new trait.

Pauan:
After thinking about it, I decided on this solution:

I added in a `print` method to `stdweb::web::error::Error` (this can be generalized to a trait later, if needed):

```rust
impl Error {
    #[inline]
    pub fn print( &self ) {
        js! { @(no_return)
            console.error( @{self} );
        }
    }
}
```

I also removed `PromiseFuture::spawn_print`. Now the user can do this:

```rust
PromiseFuture::spawn(
    create_some_future()
        .map_err(|e| e.print())
);
```

I think this is ready to merge now. @koute @CryZe @Diggsey requesting a final review on this.

CryZe:
The huge todo comment in my Executor still needs to be fixed. Basically a Future could resubmit itself while it is currently exeucting, at which point the RefCell will deadlock (panic), which is probably not what we want. The solution would be to try_lock the RefCell and on unsuccessful locking, queuing up the future for automatic re-execution when the current execution is done. That should most likely either be a Cell<usize> or Cell<bool>. I'm not quite sure if we need to count how many times it tried to resubmit, but I think we should probably just count.

Pauan:
@CryZe How would a Future resubmit itself? Ownership is passed into `spawn`, right?

CryZe:
The Future gets the Task Notifier Handle during its execution, at which point at any time it can call that and will deadlock, if it's still executing. I've at least encountered one situation where this did indeed happening. Usually it's not a problem with other executors, as they just queue up the futures to be reexecuted, but here we just lock up, as with this executor, we try to immediately execute everything.

Pauan:
@CryZe I see, I'll try and fix it. Though I don't fully understand the internals of the Future/Executor system, so it would be very helpful if you have a small test case to reproduce it!

CryZe:
If you can give me edit rights on this PR I can fix it real quick if you want.

Pauan:
@CryZe I made you a collaborator on Pauan/stdweb. The code is in the `promises` branch.

CryZe:
Alright, I think I did it. You may want to run some more testing on this, as there was a lot more subtleties involved than I thought.

Pauan:
@CryZe Why is it using `Cell`, rather than simply mutating it directly with `&mut`?

Also, what are you using to test that it's correct?

CryZe:
This needs to be behind a Cell / Lock / Atomic type in order to allow sharing + mutating it. Everything else is undefined behavior. I didn't do much testing yet. I'll probably test it a bit more tomorrow.

Pauan:
@CryZe Ahhh, I see, having multiple `*mut` pointers to the same memory location is undefined behavior. And the reason you're using `Cell` rather than `RefCell` is because `Cell` is faster. And the reason `Cell` can get away without runtime borrow checking is because it only allows `Copy` types.

The changes look good! I'll also see if I can get some tests going tomorrow.

src/webcore/promise.rs
```diff
@@ -0,0 +1,252 @@
...
+    ///
+    /// // Bluebird Promise
+    /// Promise::promisify(js!( return bluebird_promise.timeout(1000); ))
+    /// ```
+    ///
+    /// [(JavaScript docs)](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise/resolve)
+    // https://www.ecma-international.org/ecma-262/6.0/#sec-promise.resolve
+    // https://www.ecma-international.org/ecma-262/6.0/#sec-promise-resolve-functions
+    // https://www.ecma-international.org/ecma-262/6.0/#sec-promiseresolvethenablejob
+    pub fn promisify( input: Value ) -> Promise {
```
Diggsey: Why not just keep the name `resolve`?

Pauan: Because that name is super confusing.

If I see the name `promisify` I have a pretty good idea of what it does (it probably converts something into a Promise).

If I see the name `resolve`, I would never in a million years think that it converts something into a Promise. I would probably assume that it is somehow mutating an existing Promise or something (which isn't what it does).

Also, I have considered making it so that the `promisify` function only works on fake Promises, not arbitrary values, so its implementation won't be as simple as just `Promise.resolve(value)`

I'm open for suggestions on changing its name, but `resolve` is just a bad name for what it does. It's a bad name in JavaScript and an even worse name in Rust.

Pauan:
@CryZe I don't think the executor is correct. When I tried testing the code, I was able to get it to dead-lock, because the executor loop never yields to the JavaScript event loop.

I think we need to insert a `setTimeout` somewhere or other to let the JS event loop breathe a little.

Pauan:
I *think* I fixed the deadlock, but I'm not sure if it's memory safe.

I added in a `MyFuture` struct to `examples/promise/main.rs` which demonstrates the deadlock with the old code (it works fine with my changes).

Pauan:
It seems it isn't memory-safe: I get an "already borrowed: BorrowMutError" error, which I don't think ever happened with the old code.

Pauan:
I think the reason it was causing problems was because `drop_id` was being run before the `setTimeout` finished, and `drop_id` then dropped the `SpawnedTask`, so the `setTimeout` callback ended up reading into dropped memory.

I *think* I fixed it by adding in an `Rc<Cell<bool>>` which indicates whether the `SpawnedTask` is alive or dropped, and then the `setTimeout` callback checks that before calling `run`

src/webcore/promise.rs
Diggsey: There is value in sticking to the original method names though, and I don't think `promisify` is any more useful a name - you can figure that much out by the type signature.

As a compromise, how about `Promise::resolved(..)` - this at least better indicates that it's constructing a new, resolved promise rather than modifying an existing one, while being very close to the original name.

Diggsey:
Rather than implementing a custom ref-counter, it would make more sense to store an `Rc<SpawnedTask>`, remove the ref-count fields from `SpawnedTask`, and use `Rc::into_raw`/`Rc::from_raw` to convert to a task ID and back. Now you can use weak references to safely track whether a task is still alive, although in this case you probably want a strong reference...

The executor doesn't seem quite right, and you definitely don't want to call `setTimeout` before running a task because you will never be able to handle an event synchronously that way (which is important for avoiding flicker).

I think the logic for task notify should look like this:
1) Add the notified task to the back of a global "executor" queue
2) Check if running inside the executor
3) If yes, return
4) If no, run the executor

And the logic for the executor should be:
1) Pop a task from the front of the queue
2) Run the task
3) Repeat until the queue is empty

This ensures that tasks are processed promptly without danger of deadlock (since there's no recursion).

CryZe:
Can we revert the commit. The future you‘ve written is supposed to deadlock. You are not supposed to call task.notify(); unless your future can actually make progress. I don‘t think that‘s a valid future.

Pauan:
> @Diggsey The executor doesn't seem quite right, and you definitely don't want to call setTimeout before running a task because you will never be able to handle an event synchronously that way (which is important for avoiding flicker).

Most of the time it is completely synchronous. It only uses `setTimeout` if the Future calls `notify` at the same time as it's being polled. That generally won't happen except in contrived situations (like the test I created).

> This ensures that tasks are processed promptly without danger of deadlock (since there's no recursion).

The deadlock is caused by the Future synchronously re-queuing itself inside of `poll`, which will still happen with your suggestion.

----

@CryZe Is that really true? Where is that written? What do the other non-JS (e.g. Tokio) Executors do?

[The documentation](https://docs.rs/futures/0.1.18/futures/task/struct.Task.html#method.notify) is quite clear that it is possible to call `notify` while in the middle of `poll`, and I don't see any requirements for making progress or anything like that.

And the documentation strongly implies that enqueuing can be asynchronous, i.e. it isn't guaranteed that the Future will run immediately after calling `notify`

CryZe:
Yeah, it's not explicitly documented anywhere (that I've seen) that you shouldn't call task.notify() unless the future can make progress. It's however also not documented that an executor needs to be able to handle a nonsensical spam of notifications that don't correlate to any actual progress. Pretty much every other executor will have an actual queue where this spam will very quickly bring the whole program down by consuming more and more memory until it crashes, while consuming 100% cpu on that thread during the whole execution. So even with normal executors, this future just doesn't have any reasonable behavior. So if you find a different future, that works just fine with a normal executor, but not with ours, then I would consider that an actual bug.

src/webcore/promise.rs
Pauan: I can compromise with `Promise::resolved`, though I do worry about the confusion for people who are used to JavaScript (they need to remember that it ends in `d` in Rust, but doesn't in JavaScript)

I think for that reason `Promise::resolve` would be better than `Promise::resolved`, though both are quite bad.

The more I think about it, the more I prefer the solution of changing `Promise::promisify` to only convert fake Promises, and then the name makes more sense. Or perhaps rename it to `Promise::convert`, I'm still open to name suggestions.

Pauan:
@CryZe Well, I certainly don't mind reverting the changes (since the older code is a lot simpler, and faster), I just don't want to have Futures that rely upon another Executor's behavior and then it breaks on our Executor. In other words, I want our behavior to match the de-facto behavior of other Executors.

CryZe:
Yeah, that's why I think we should just ask at this point.

Diggsey:
> Pretty much every other executor will have an actual queue where this spam will very quickly bring the whole program down by consuming more and more memory until it crashes,

That's not true, most executors will not re-queue a task multiple times, and do not implement task execution through recursion. If we implement the executor according to the rules I described above, then it's easy to avoid both these problems.

The future @Pauan wrote *will* sit in an infinite loop, because it constantly notifies itself, but it shouldn't consume an increasing amount of resources.

Pauan:
> @Diggsey That's not true, most executors will not re-queue a task multiple times. [...] You just need to check before adding a task to the queue that it's not already in the queue.

That would certainly be easier for us to implement, but that doesn't sound like it's spec-compliant, since [the docs specify quite clearly](https://docs.rs/futures/0.1.18/futures/task/struct.Task.html#method.notify) that you *can* call `task.notify()` while it is being polled, and it *must* call `poll` again (it cannot just ignore it).

Diggsey:
This pseudocode is how most executors are implemented:
```rust
// executor
while <termination_condition> {
    while let Some(task) = queue.pop() {
        task.queued = false;
        task.poll();
    }
    sleep_until_woken();
}

// task::notify
if !task.queued {
    queue.push(task);
    task.queued = true;
    executor.wake_up();
}
```

To translate that to javascript, we have to invert the control flow: we want the `sleep_until_woken()` call to be when the javascript event loop is running, and we don't want to yield control back when we're not inside that call.

This results in the following pseudocode:

```rust
// executor
executor.is_running = true
while let Some(task) = queue.pop() {
    task.queued = false;
    task.poll();
}
executor.is_running = false

// task::notify
if !task.queued {
    queue.push(task);
    task.queued = true;
    if !executor.is_running {
        executor.run();
    }
}
```

Diggsey:
@Pauan if that task is still running, then it's already been popped from the queue, so will be re-queued accordingly. If you call `task::notify()` twice then there's *no need* to poll it twice also.

This doesn't work in the current implementation because you've implement `task::notify()` recursively, so *all* of the notified tasks are always running. This is very bad for performance.

Pauan:
> @Diggsey If you call task::notify() twice then there's no need to poll it twice also.

A literal interpretation of the docs says that it *must* call `poll` twice if it calls `task.notify()` twice. But maybe the docs are wrong. The only way to be sure is to ask Alex.

> This doesn't work in the current implementation because you've implement task::notify() recursively, so all of the notified tasks are always running. This is very bad for performance.

What do you mean? It only recurses when the `poll` implementation calls `task.notify()` *while `poll` is running*, otherwise it's not recursive at all.

Diggsey:
> A literal interpretation of the docs says that it must call poll twice if it calls task.notify() twice. But maybe the docs are wrong. The only way to be sure is to ask Alex.

No, it doesn't say that in the docs at all. It says if you call `notify` from inside `poll`, then the task must be re-polled again at some point in the future. That doesn't mean that if you call `notify` twice it must be re-polled twice.

> The only way to be sure is to ask Alex.

I've already discussed this on the gitter channel, and there's no doubt about the behaviour.

Pauan:
> @Diggsey No, it doesn't say that in the docs at all.

The docs says:

> for ***each*** call to notify, poll will be called ***at least once*** subsequently (unless the future has terminated).

The only way to interpret that is that every single call to `notify` will cause at least one call to `poll`, i.e. 5 calls to `notify` will cause `poll` to be called 5+ times.

> I've already discussed this on the gitter channel, and there's no doubt about the behaviour.

Then the docs should be updated, since they're wrong.

CryZe:
Alright, I wasn‘t sure initially about this behavior either, so it was a matter of resubmission_count: Cell<usize> vs. is_resubmitted: Cell<bool>, however that still deadlocks as the future will always resubmit itself, never allowing the outer execution loop to actually end.

Diggsey:
> he only way to interpret that is that every single call to notify will cause at least one call to poll, i.e. 5 calls to notify will cause poll to be called 5+ times.

That's not the *only* way to interpret it, but it is slightly misleading. To avoid any doubt, here is the implementation from tokio:

https://github.com/tokio-rs/tokio/blob/master/src/executor/scheduler.rs#L613

As you can see, it only queues the task if it hasn't already been queued.

Diggsey:
I opened https://github.com/rust-lang-nursery/futures-rs/pull/740

Diggsey:
@CryZe I also found a better way to trigger the executor. Instead of calling it synchronously, we should do this:
```rust
Promise.resolve().then(executor.run)
```

So the psuedocode for everything becomes:
```rust
// executor
executor.is_running = true
while let Some(task) = queue.pop() {
    task.queued = false;
    task.poll();
}
executor.is_running = false

// task::notify
if !task.queued {
    queue.push(task);
    task.queued = true;
    if !executor.is_running {
        Promise.resolve().then(executor.run);
    }
}
```

This has the effect of running the executor in a micro-task *after* the user's code returns, but before the javascript event loop proceeds.

Pauan:
> @Diggsey To avoid any doubt, here is the implementation from tokio

That is very enlightening, thanks!

> I opened rust-lang-nursery/futures-rs#740

Great, thanks!

> Instead of calling it synchronously, we should do this: `Promise.resolve().then(executor.run)`

What's the benefit of doing this compared to immediately executing it (which is what it does right now)?

Diggsey:
For one, `task::notify` is expected to be lightweight - there are likely to be a few implementations that don't respond well to running the task synchronously. Another is that it avoids polluting the stack trace with other unrelated tasks. Finally it means that you can combine multiple task notifications and do so less total work.

Pauan:
So we got an answer, deadlocks are okay, so I went ahead and changed the implementation. I also changed it so that multiple calls to `notify` (within the `poll` method) will only trigger one call to `poll`.

----

> @Diggsey For one, task::notify is expected to be lightweight - there are likely to be a few implementations that don't respond well to running the task synchronously.

I'm having a hard time thinking of something that would break because `execute` is run synchronously.

In particular, JavaScript is single-threaded so *everything* runs synchronously. And there are other Executors which also run things synchronously.

Do you have any examples of problems (even a contrived example)?

> Another is that it avoids polluting the stack trace with other unrelated tasks. Finally it means that you can combine multiple task notifications and do so less total work.

There are three situations where `notify` is called:

1. Within `poll`. In this case our implementation will already ignore multiple calls to `notify`.

2. Multiple times synchronously outside of `poll`. In this case it *might* call `poll` multiple times, though I don't think that's a big deal.

3. Asynchronously. This is by-far the most common situation, and it already doesn't have the problems you mentioned.

So the only situation that is affected is number 2, and I don't think the extra complexity and non-determinism is worth the benefit of optimizing number 2.

As for the stack trace, I don't think that's an issue: with our implementation, each Future should get its own separate stack trace.

Diggsey:
> In particular, JavaScript is single-threaded so everything runs synchronously. 

You're conflating parallelism with synchronicity, they're very different.

> And there are other Executors which also run things synchronously.

I'm not aware of *any* executors which do this? Neither the tokio executor, the thread pool executor, the new "same thread executor" or the legacy ".wait()" executor or the gtk-based executor do this, so doing it for the javascript one is completely unprecedented.

> Do you have any examples of problems (even a contrived example)?

Yes, if you use a RefCell/Mutex/RwGuard in your code (very common for futures...) it will be susceptible to deadlocks or panics when run on this executor. No other executor has this problem.

> Multiple times synchronously outside of poll. In this case it might call poll multiple times, though I don't think that's a big deal.

Well you'd be wrong. Existing frameworks go to great lengths to reduce the number of updates and coalesce events.

Pauan:
> @Diggsey I'm not aware of any executors which do this?

As far as I know, *all* Executors will start up the event loop as soon as they start running. They don't delay for an arbitrary amount of time first.

And there are quite a few Executors which run entirely synchronously, such as the [`current_thread`](https://docs.rs/tokio/0.1/tokio/executor/current_thread/index.html) executor.

Also, I find it very odd that you claim that `wait()` doesn't run synchronously, because it *literally blocks the entire thread* while it's running, which can very easily prevent progress and cause deadlocks! It is 100% synchronous, by every definition of the word.

> Yes, if you use a RefCell/Mutex/RwGuard in your code (very common for futures...) it will be susceptible to deadlocks or panics when run on this executor. No other executor has this problem.

Everything you mentioned will run synchronously no matter what, because existing Future code doesn't use any of JavaScript's concurrency mechanisms (such as `setTimeout`).

Even if you use `Promise.resolve().then(executor.run)` to shift the `executor.run` method onto the event loop, the `executor.run` method itself will *still be running synchronously*.

Also, when I said "example", I meant *actual code that demonstrates the problem*.

> Well you'd be wrong. Existing frameworks go to great lengths to reduce the number of updates and coalesce events.

To be clear, I am talking about this specific situation:

```rust
// Not inside of the poll method
task.notify();
task.notify();
```

Do *all* existing Executors notify the Task exactly once, in that situation?

You are the one making claims, *you must back up your claims*.

Pauan:
I renamed `Promise::promisify` to `Promise::convert`, and now it only accepts Promise-like objects (it returns `None` if it's not a Promise-like object). I like this a lot better than the old `promisify` API.

CryZe:
My take on this is that I believe the executor should execute be very slim and should just execute everything it can immediately. The choice of deferring some notifications or the initial execution should be up to the user here. In stdweb-io I have a defer() future that you can await / chain to yield execution back to the runtime whenever you think it makes sense. Additionally I have spawn_deferred that does the same thing, but on spawning. So we could just provide these as well and document clearly when it makes sense to use them and when not. Like for example if you are making changes to the DOM right now, you may want to defer calculations that aren‘t necessary for calculating the DOM (like some cleanup routines), to reduce the latency of the frame being rendered. So I feel like chaining these in is probably the best solution here, having a very small and fast runtime when you don‘t need to defer, but allowing the user to defer whenever they deem it useful.

Additionally we may want to consider adding Timeout / Interval futures too, as those are probably very useful, so most people would have to write down the little helper function we have in our example right now, might as well move it over completely.

Pauan:
@CryZe I agree that `timeout` / `interval` / `defer` would be very useful, but I think we should add those in as a separate pull request. This pull request has gotten big enough in my opinion.

Can you do another review of the code? I think it's *finally* ready for merging (famous last words, I know...)

The only thing I can think of that we might want to change is to add in @Diggsey 's suggestion to use `Rc<SpawnedTask>` rather than implementing our own custom ref-counting. But that's an implementation detail which can be changed later.

CryZe:
Yeah I agree. The only problem I still see right now is that the example seems more like a test than an actual user facing example to me. All these .notify() calls don't seem like anything that anyone other than someone implementing a future (and even then not that often) even needs to bother about. So I feel like a nice clean example that actually does something a user would care about makes more sense (but keeping the current more unit-testy example somewhere else).

Diggsey:
> As far as I know, all Executors will start up the event loop as soon as they start running. They don't delay for an arbitrary amount of time first.

I'm not sure what that has to do with anything, we were talking about when a task is executed in relation to it being notified.

> And there are quite a few Executors which run entirely synchronously

Synchronicity is a property that relates to a *function call*: no current executors synchronously execute their tasks in response to the `task.notify` function call.

> Also, when I said "example", I meant actual code that demonstrates the problem.

I really can't be asked writing an entire example just to persuade you that this is an issue, but here's some pseudocode:

```
// TaskA::poll()
let guard = shared_state.borrow_mut();
...
task_b.notify();

// TaskB::poll()
let guard = shared_state.borrow();
```

This example will panic on this executor but will work fine on every other executor.

> Do all existing Executors notify the Task exactly once, in that situation?

I've already pointed you to the source for tokio, go have a look yourself if you're interested in how other executors work.

What I was actually referring to was *javascript frameworks* like react, angular, and the numerous state management libraries commonly used with them: a real potential issue with these libraries is "update amplification": sure it may not look like a big deal that your future calls notify twice, but when you have many futures running, and each one potentially notifies 2-3 other futures, then the entire "update" step grows *exponentially*. By queuing updates, running them sequentially, and removing duplicates, you avoid that exponential time bound.

> You are the one making claims, you must back up your claims.

They're not exactly outrageous claims, I was expecting them to be self-evident once considered.


@CryZe 

> My take on this is that I believe the executor should execute be very slim and should just execute everything it can immediately.

I believe an executor based on `Promise::resolve()` can be simpler than the current executor. I'm also not convinced that not deferring is even a valid executor, and if it is I will argue for it to be changed to disallow that, since it breaks the patterns I described above.

Diggsey:
OK, I've just checked and this implementation is not only disallowed, it's actually currently exhibiting undefined behaviour.

`poll` takes a mutable borrow to `self`: that means that when you call `current_task().notify()`, you have two mutable references to the same task alive at the same time.

CryZe:
> edit: OK, with the more recent changes to prevent recursing it's no longer undefined behaviour, but I'm still not convinced it's a valid executor.

It never had this undefined behavior, it always was in a RefCell, so there was never the possibility of 2 mutable references. The only problem initially was that the RefCell panicked on the second mutable borrow.

Can you post a gist of how your modified Executor would look like? The example you showed is somewhat convincing, but at the same time I'm also not quite sure if this is something that necessarily needs to work for this to be a correct Executor. But probably it should. If your alternative implementation is straight forward enough we can just switch to that I guess.

CryZe:
I have roughly sketched what it would look like: https://gist.github.com/CryZe/81251916e9219db03f96444a96396b6d

I wasn't entirely sure what the Running states around the decrement_ref_count are for, so that may be wrong. Additionally there might be other edge case that I've missed.

Pauan:
>@Diggsey I'm not sure what that has to do with anything, we were talking about when a task is executed in relation to it being notified.

Your pseudo-code initializes the event loop using `Promise.resolve().then(executor.run)`. In particular, it doesn't actually run the individual items concurrently, only the *entire event loop* is being delayed. So it is relevant.

> I'm also not convinced that not deferring is even a valid executor, and if it is I will argue for it to be changed to disallow that, since it breaks the patterns I described above.

I find it odd that earlier I created a deferred implementation, and you didn't like it because you wanted it to be synchronous (to avoid flicker, etc.) but now you're wanting a deferred implementation. So what changed?

> I believe an executor based on Promise::resolve() can be simpler than the current executor

I very highly doubt that. The current implementation is very simple, adding a queue will make it more complicated. Using `Promise.resolve()` might make it *more correct* or *more useful* but I would be extremely surprised if it made the implementation simpler.

> I really can't be asked writing an entire example just to persuade you that this is an issue.

Then I can't really be bothered to listen to you.

I created this example (it didn't take long), and it runs just fine:

```rust
#[cfg(target_arch = "wasm32")]
#[macro_use]
extern crate stdweb;
extern crate futures;

use std::rc::Rc;
use std::cell::RefCell;
use futures::Future;
use futures::{Poll, Async};
use futures::task::{current, Task};

#[cfg(target_arch = "wasm32")]
fn main() {
    use stdweb::{PromiseFuture};

    struct TaskA {
        shared_state: Rc<RefCell<u32>>,
        task_b: Task,
    }

    impl Future for TaskA {
        type Item = ();
        type Error = ();

        fn poll(&mut self) -> Poll<Self::Item, Self::Error> {
            js! { console.log("Poll TaskA"); }

            let foo = self.shared_state.borrow_mut();

            js! { console.log(@{format!("TaskA 1: {:#?}", foo)}); }

            self.task_b.notify();

            js! { console.log(@{format!("TaskA 2: {:#?}", foo)}); }

            Ok(Async::NotReady)
        }
    }

    struct TaskB {
        shared_state: Rc<RefCell<u32>>,
        initialized: bool,
    }

    impl Future for TaskB {
        type Item = ();
        type Error = ();

        fn poll(&mut self) -> Poll<Self::Item, Self::Error> {
            js! { console.log("Poll TaskB"); }

            if !self.initialized {
                self.initialized = true;

                let task_b = current();

                PromiseFuture::spawn(TaskA {
                    shared_state: self.shared_state.clone(),
                    task_b: task_b,
                });
            }

            let foo = self.shared_state.borrow();

            js! { console.log(@{format!("TaskB 1: {:#?}", foo)}); }

            Ok(Async::NotReady)
        }
    }

    PromiseFuture::spawn(TaskB {
        shared_state: Rc::new(RefCell::new(0)),
        initialized: false
    });
}
```

So either you provide an example which proves that our Executor implementation is broken, or I will just assume that it's fine.

CryZe:
The example that fails is that the `let foo = self.shared_state.borrow()` happens before the spawn of the other Task (or rather that's not quite what Diggsey's example was, but it triggers the same bug).

Pauan:
@CryZe Hmmm, when I move the `borrow()` to before it spawns `TaskA`, then I get this:

```
Uncaught (in promise) RuntimeError: unreachable
    at wasm-function[49]:1282
    at wasm-function[60]:115
    at wasm-function[59]:122
    at wasm-function[146]:147
    at wasm-function[173]:140
    at wasm-function[7]:752
    at wasm-function[28]:399
    at wasm-function[8]:2562
    at wasm-function[28]:399
    at wasm-function[6]:235
```

This isn't a Rust error or panic, it's a WebAssembly error. It sounds like either Rust is generating incorrect WebAssembly, or we're somehow screwing something up with unsafe code.

When I run it with the `wasm32-unknown-emscripten` target then I *do* get a proper `BorrowMutError` panic.

I'll investigate whether other Executors also panic or not.

Diggsey:
> Your pseudo-code initializes the event loop

No, the event loop (and executor) always exists, that's why `CORE` is a static variable, whereas other executors need to be instantiated. In javascript you can never run code "outside" the event loop.

> it doesn't actually run the individual items concurrently,

Again, never mentioned concurrency, that's a totally separate concept.

> only the entire event loop is being delayed.

No, polling the outstanding notified tasks is being delayed, which is the important thing. This polling process is something that typically happens on every iteration of the event loop.

> I find it odd that earlier I created a deferred implementation, and you didn't like it because you wanted it to be synchronous (to avoid flicker, etc.) but now you're wanting a deferred implementation. So what changed?

Originally you were using `setTimeout(f, 0)` which runs the callback *after* processing browser events such as redraws. Furthermore, you were only deferring the task execution, rather than the polling step, which means you don't get the benefit of either. At that point I favoured the non-deferred option because it was the best that could be done without `Promise.resolve()`.

I was dissatisfied that and so I did some research and discovered `Promise.resolve()`. Using `Promise.resolve().then()` is perfect because it defers execution but runs it *before* processing browser events. This results in identical behaviour to the existing executors, where all notified futures are polled before yielding.

> Then I can't really be bothered to listen to you.

My original comment was directed at @CryZe because they wrote the executor.

> This isn't a Rust error or panic, it's a WebAssembly error. It sounds like either Rust is generating incorrect WebAssembly, or we're somehow screwing something up with unsafe code.

No, that's just what panics look like on webassembly right now. wasm uses panic = abort, and abort just generates an unreachable block.

CryZe:
So can you guys check my sketch implementation? I honestly believe it's even simpler than what we currently have and it solves the problem.

Pauan:
I just tested it, and this code *also* gives a `BorrowMutError`:

```rust
extern crate futures;

use std::rc::Rc;
use std::cell::RefCell;
use futures::Future;
use futures::{Poll, Async};
use futures::task::{current, Task};

fn main() {
    struct TaskA {
        shared_state: Rc<RefCell<u32>>,
        task_b: Task,
    }

    impl Future for TaskA {
        type Item = ();
        type Error = ();

        fn poll(&mut self) -> Poll<Self::Item, Self::Error> {
            println!("Poll TaskA");

            let foo = self.shared_state.borrow_mut();

            println!("TaskA 1: {:#?}", foo);

            self.task_b.notify();

            println!("TaskA 2: {:#?}", foo);

            Ok(Async::NotReady)
        }
    }

    struct TaskB {
        shared_state: Rc<RefCell<u32>>,
        initialized: bool,
    }

    impl Future for TaskB {
        type Item = ();
        type Error = ();

        fn poll(&mut self) -> Poll<Self::Item, Self::Error> {
            println!("Poll TaskB");

            if !self.initialized {
                self.initialized = true;

                let task_b = current();

                let foo = self.shared_state.borrow();

                println!("TaskB 1: {:#?}", foo);

                TaskA {
                    shared_state: self.shared_state.clone(),
                    task_b: task_b,
                }.wait();
            }

            let foo = self.shared_state.borrow();

            println!("TaskB 1: {:#?}", foo);

            Ok(Async::NotReady)
        }
    }

    TaskB {
        shared_state: Rc::new(RefCell::new(0)),
        initialized: false
    }.wait();
}
```

Our implementation of Executor is valid and correct.

CryZe:
I don't agree with that. Each .wait() call is its own executor. They don't communicate with each other in any way, so that this locks up doesn't surprise me. So to replicate the same situation you'd have to spawn TaskA onto the .wait() executor of TaskB, which is probably not a thing.

Pauan:
> @CryZe I don't agree with that. Each .wait() call is its own executor.

That's true with my stdweb example too: it creates two Executors, one for TaskA and one for TaskB.

And why does it matter? @Diggsey claimed that it would work fine on other Executors, but not on our Executor, therefore our Executor is invalid. But the code is equally broken on other Executors, so our Executor is just as valid.

It might be useful to use a queueing system if it helps to prevent these problems, but that's not the same thing as our current implementation being *wrong* or *invalid*.

CryZe:
Two executors will obviously lock up, as they don't synchronize with each other. So you'd have to find an executor that is single threaded, and allows spawning of futures onto the same executor, which is pretty much only the old tokio core reactor and the new tokio current_thread executor. Can you try these? Both of these shouldn't have the BorrowMut error.

Pauan:
> @CryZe Can you try these? Both of these shouldn't have the BorrowMut error.

Yup, `current_thread` works fine, which suggests that a queue model will be better.

> Also no, stdweb has a single global executor (static CORE = ...). That is different from two separate executors.

Technically speaking `Core` doesn't really exist (it's an empty struct, and it never uses `self`, ever). So it would be more correct to say that we don't have an Executor at all, only ethereal `SpawnedTask` floating around in nothingness. Which is probably pretty close in behavior to spawning a new Executor for each Task.

But you're right that `wait()` is a pretty simplistic executor, which I think is a pretty fair comparison to our current (very simplistic) executor. Tokio's `current_thread` is more comparable to your queueing implementation.

> The difference is that every time you call .wait() you get a different notifier (kind of)

But our implementation creates a new `SpawnedTask` every time you call `spawn`, and each `SpawnedTask` is notified independently, so isn't that the same thing?

> However this also makes me question if this really is a property of Executors in general that needs to work. But it's at least a footgun that you wouldn't necessarily expect, since I'm 99% sure it doesn't apply to tokio.

I agree that it would be nice to have an Executor that works in more cases rather than less cases (even if both are valid). I'll take a deeper look at your queue implementation.

Diggsey:
@CryZe that's definitely an improvement! There's still a danger of conflict if you were to share a RefCell between code not running on `CORE` which calls `task::notify` and the notified task, so I'd still prefer a solution based around `Promise::resolve()` - I believe it can be made very simple so I hope to mock one up this evening (am at work atm).

@Pauan 
> Technically speaking Core doesn't really exist

`Core` is just a proxy for the javascript event loop, which always exists. The difference between having two executors vs one executor is super important, especially without threads.

This is because when you "run" an executor (something which never happens in our implementation, because our executor is a proxy for the javascript event loop which is *always* running) it blocks synchronously until the futures you have spawned to it (which are not daemon futures) have completed. In javascript you never see this synchronous blocking because it happens further up the stack, out of sight.

Diggsey:
Here is my executor implementation:

https://github.com/Pauan/stdweb/compare/promises...Diggsey:promises

It is quite a bit shorter, simpler, uses less unsafe code, and will not panic or deadlock for any valid futures.

CryZe:
That seems good

Pauan:
@Diggsey Using `Rc` is a good idea, and genuinely does simplify things.

Overall I like the code, however, there are two major issues with it:

1. It causes a memory leak. The `notify` method shouldn't use `clone_raw`, it should just use a normal `.clone()`

2. After I fixed the memory leak, it now gives a `BorrowMutError`. That's because you're using `self.spawn.borrow_mut()`. You'll need to fix that in the same way that we did, by using `try_borrow_mut`

Pauan:
Actually, after thinking about it some more, I'm not sure if `Rc` is actually an improvement, either in terms of code size *or* simplicity.

The fact that it sometimes needs to use `.clone()` and other times needs to use `clone_raw` is a source of subtle complexity, and the implementation of `clone_raw` is quite hacky, as you noted.

It's hard to be certain that all of the ref-counting is correct, because of the weird way that we're using `Rc`. So it's probably simpler if we just use the old manual-ref-counting approach.

Diggsey:
> However, using std::mem::forget still seems really bad: doesn't it leak the memory for the cloned Rc?

No, it just avoids calling the destructor on the `Rc`. Here is an equivalent implementation for `clone_raw`:

```rust
unsafe fn clone_raw<T>(ptr: *const T) -> Rc<T> {
    let temp = Rc::from_raw(ptr);
    let result = temp.clone();
    Rc::into_raw(temp);
    result
}
```

The point is that `Rc::into_raw` assumes we're taking ownership of the raw pointer, but that's not what we want: we just want to clone it without taking ownership.

Pauan:
@Diggsey @CryZe Any more suggestions for things to improve?

If not, then @koute this is *finally* ready to merge.

CryZe:
Well I still don't like how our example is a test and not an example that a user would care about

Diggsey:
Yeah that's not ideal. Maybe we could add a couple of new future-based versions of existing web APIs and use them in the example to do something useful?

Another issue is that it imports `Error` from `webapi` into `webcore`. I think it's intended at some point to split them up.

Pauan:
@CryZe Then where do you suggest we put it? @koute said that until we get support for async unit tests, that I should create examples that serve as unit tests.

---

> @Diggsey Maybe we could add a couple of new future-based versions of existing web APIs and use them in the example to do something useful?

That will be done in a separate pull request. This pull request is quite big enough, in my opinion.

> Another issue is that it imports Error from webapi into webcore. I think it's intended at some point to split them up.

What do you mean? The imports are correct. If somebody else wants to move `Error` somewhere else, then *they* will need to migrate all existing code to use the new module path.

Diggsey:
> What do you mean? The imports are correct. 

None of the existing code in `webcore` imports from `webapi`. This PR breaks that separation.

Pauan:
> @Diggsey None of the existing code in webcore imports from webapi. This PR breaks that separation.

Then it would be a good idea to move `Error` into `webcore`, which is where it should be. But I think that should be done in another pull request. Let's try to keep the pull requests as small and self-contained as we can.

Pauan:
Well, if there aren't any other objections, then @koute could you please review and merge this?

The remaining issues (adding in web APIs like `setTimeout`, making *proper* unit tests, and moving `Error`) can be done in a different pull request.

src/webcore/promise.rs
```diff
@@ -0,0 +1,260 @@
...
+    ///     }
+    /// });
+    /// ```
+    ///
+    /// [(JavaScript docs)](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise/then)
+    // https://www.ecma-international.org/ecma-262/6.0/#sec-performpromisethen
+    pub fn done< A, B >( &self, callback: B )
+        where A: TryFrom< Value >,
+              A::Error: std::error::Error,
+              B: FnOnce( Result< A, Error > ) + 'static {
```
koute: The error handling needs some work here.

1. I don't think using `Error` (as in - the JavaScript type `Error`) here is appropriate, as *technically* you can throw things which are not `Error` objects in JavaScript. It should be possible for the user to be able to retrieve the original value from promise's rejection.
2. It would be nice to figure out how to fit the `ConversionError` in this somehow. Perhaps convert it to a JavaScript `TypeError`? (This issue is also relevant in many other places across `stdweb` when we currently panic, but where I'd like to eventually throw a JavaScript exception when a type conversion fails.)

src/webcore/promise.rs
```diff
@@ -0,0 +1,260 @@
...
+    ///
+    /// // Bluebird Promise
+    /// Promise::convert(js!( return bluebird_promise.timeout(1000); ))
+    /// ```
+    ///
+    /// [(JavaScript docs)](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise/resolve)
+    // https://www.ecma-international.org/ecma-262/6.0/#sec-promise.resolve
+    // https://www.ecma-international.org/ecma-262/6.0/#sec-promise-resolve-functions
+    // https://www.ecma-international.org/ecma-262/6.0/#sec-promiseresolvethenablejob
+    pub fn convert( input: Value ) -> Option< Self > {
```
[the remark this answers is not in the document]
Pauan: But if it accepts `Reference` then you would need to do something like this:

```rust
Promise::convert(js!( return ... ).try_into().unwrap())
```

...which is annoying, verbose, and slower. And it doesn't really help, because even with a `Reference` it *still* might return `None`, so you don't gain any real static type safety.

Also, I've noticed that sometimes objects get turned into `Object`, not `Reference`, and the `Promise::convert` function works with *any* object which happens to have a `then` method (even Arrays!). So I think using `Value` is correct.

src/webcore/promise.rs
```diff
@@ -0,0 +1,260 @@
...
+    /// This method converts the `Promise` into a [`PromiseFuture`](struct.PromiseFuture.html), so that it can be used as a Rust [`Future`](https://docs.rs/futures/0.1.18/futures/future/trait.Future.html).
+    ///
+    /// # Examples
+    ///
+    /// ```rust
+    /// promise.to_future().map(|x| x + 1)
+    /// ```
+    // We can't use the IntoFuture trait because Promise doesn't have a type argument
+    // TODO explain more why we can't use the IntoFuture trait
+    pub fn to_future< A >( &self ) -> PromiseFuture< A >
```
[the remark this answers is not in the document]
Pauan: What about the `pub use webcore::promise::{Promise, PromiseFuture};` code in `lib.rs`? Should that also be `cfg`-gated?

Also, is it enough to only `cfg`-gate `PromiseFuture`, and not the `to_future` method?

src/webcore/promise.rs
Pauan: Ah, good catch, you're right that technically any error can be thrown, so it should be `Result<Value, Value>` (which can be auto-coerced into something more precise by the user if they desire).

> It would be nice to figure out how to fit the ConversionError in this somehow.

What do you mean? It doesn't panic, it converts the `ConversionError` into a JavaScript `Error`. So what are you suggesting? To convert it into a `TypeError` instead?

src/webapi/error.rs
```diff
@@ -36,6 +36,26 @@ pub trait IError: ReferenceType {
...
+    #[inline]
+    pub fn new( description: &str ) -> Self {
+        js!( return new Error( @{description} ); ).try_into().unwrap()
+    }
+
+    /// Prints the `Error` to the console (this prints the error's description and also its stack trace).
+    ///
+    /// [(JavaScript docs)](https://developer.mozilla.org/en-US/docs/Web/API/Console/error)
+    #[inline]
+    pub fn print( &self ) {
```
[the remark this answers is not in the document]
Pauan: That's fair, though I had kind of figured that `console.error(e)` was a placeholder, and we might implement better error handling in the future, in which case having `e.print()` would be nice. But I don't have a particularly strong opinion about it, so I don't mind removing the `print` method.

src/webcore/promise.rs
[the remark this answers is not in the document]
koute: Both `PromiseFuture` and `to_future` need to be feature gated. `Promise` itself should not. (You can split the `pub use` line in `lib.rs` into two lines.)

src/webcore/promise.rs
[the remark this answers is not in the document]
koute: > Also, I've noticed that sometimes objects get turned into Object (not Reference)

This is actually a very good point, thank you! Hmm...

The original idea was that `Object` objects represent a plain, dumb, JavaScript object for holding data, but unfortunately a lot of "fat" objects (e.g. like Bluebird promises) in the wild seem to be indistinguishable from those which are just simple `{key: value}` containers for data. I'm thinking that perhaps we should stop treating `Object`s and `Array`s in a special way, make them a normal `#[derive(ReferenceType)]` types and only have a `Reference` variant in `Value`.

I'm starting to really like this idea; then the problem of promises (and other kinds of objects) being converted to an `Object` will disappear.

> And it doesn't really help, because even with a Reference it still might return None, so you don't gain any real static type safety.

(Assuming we'll remove `Value::Object` variant.)

1. Yes, but, conversely if someone already has a `Reference` they have to convert it to a `Value` now to pass into this function.
2. The types are there not only for guaranteeing type safety, but also to serve as documentation.
3. Casting a `Value` to a `Reference` is relatively cheap as it's just a `match`.
4. You could also convert it this way: `js!( ... ).as_reference().and_then( Promise::convert ).unwrap()`, which I don't think is that bad?

src/webcore/promise.rs
koute: > What do you mean? It doesn't panic, it converts the ConversionError into a JavaScript Error. So what are you suggesting? To convert it into a JavaScript TypeError instead?

Yes, that's what I'm suggesting, since essentially a `ConversionError` *is* our counterpart of JS' `TypeError`. Another way to handle this would be to have an enum with two variants: one would be a `Value`, and another would be... something else (perhaps even the `ConversionError` if we make its innards private) which would signify a type error but without converting it to a JS type (keeping it purely in the Rust land).

I'm not sure which one would be better, but I'm *probably* leaning towards the first one's direction.

src/webcore/promise.rs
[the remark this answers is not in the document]
Pauan: If we remove the `Object` and `Array` variants, then yes I agree it would make more sense to make it `Reference`. But it's still a lot more verbose than using `Value`, so I'm not sure.

I'll leave it as `Value` for now, with a TODO note to change it later (if we decide to remove `Object` and `Array` variants).

src/webcore/promise.rs
Pauan: I tried changing it to allow for arbitrary types for the error, but I got a lot of type errors (that I wasn't able to fix), so for now I'm just going to `.unwrap()` the `ConversionError` errors, but I'll add a TODO note to fix it later.

src/webapi/error.rs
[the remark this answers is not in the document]
Pauan: Since we don't currently have a `console::error` function, I'd rather leave this in for now. Afterwards I'll make another pull request which adds in `console::error` and removes `print`

Pauan:
@koute I changed `PromiseFuture` to have two type arguments: one for the success type and one for the error type. In other words, it's now possible to use arbitrary error types with `PromiseFuture`.

Which also means it's now possible to be more precise with the error type (e.g. `PromiseFuture<String, DomException>`)

I also renamed `Promise::convert` to `Promise::from_thenable`, since "thenable" is the standard JavaScript term for "something which has a `then` method".

I also fixed the `cfg`-gating (and verified that it works). It was easier to just move the `futures`-specific stuff into a `promise_future.rs` file, rather than trying to `cfg` individual statements.

Things to do after this pull request is merged:

* Add in `console::error` and remove `Error::print`

* Move `Error` into `webcore`

* Remove `Object` and `Array` variants from `Value`, and change `Promise::from_thenable` to accept a `&Reference` instead of `Value`

* Add in Futurified versions of some web APIs (e.g. `setTimeout`, etc.)

* Make proper async unit tests, rather than using `examples/promise`

* Low priority, but it would be nice to figure out a way to avoid `.unwrap()`

I'll do as many of these as I can, but anybody else can do them if they want to.

koute:
Ok, I'll merge it in as-is, but as discussed this still needs some work so if by the time I'll be releasing `0.4` (which will hopefully be soon-ish, I still have a few APIs go to through) this won't be ready I'll have to temporarily make it private.

Thanks for the great work!

koute:
@Pauan FYI, I've just got rid of the `Object` and `Array` variants in `Value`!

PR #104, opened 2018-02-09
Use 2x version of Patreon button in README.md

jdanford:
It looks a lot nicer on my screen(s) ¯\\\_(ツ)\_/¯

I'm aware that the previous version was mirrored, so you can update it there instead, but either way I wanted the README to look nice because I think this project is great!

### DIFF
--- a/README.md
+++ b/README.md
@@ -14,7 +14,9 @@ a high degree of interoperability between Rust and JavaScript.
 
 ## Donate
 
-[![Become a patron](https://koute.github.io/img/become_a_patron_button.png)](https://www.patreon.com/koute)
+<a href="https://www.patreon.com/koute">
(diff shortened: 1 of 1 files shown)


koute:
Thanks for the suggestion!

Hmm... is there an `svg` version of that button anywhere? I'm reluctant to change it to an `` link since those are not rendered on `crates.io`; an `svg` would be ideal in this case.

jdanford:
Patreon's brand page mentions a widget that can be found on the [widgets dashboard](https://www.patreon.com/dashboard/widgets), but I don't have a creator account so I can't access it. Other than that, I'm not aware of any alternatives.

jdanford:
Closing this since I haven't come across an SVG version of the button.

PR #110, opened 2018-02-10
Added WebGL Rendering Context

liana-p:
First time writing Rust, sorry if I did something wrong

* Added basic `WebGLRenderingContext` with two functions as proof of concept
* It's in a separate file to not pollute the `RenderingContext` file (and I would advise moving the 2d context out of that file too)
* Added an example similar to the canvas one (it just draws a color on screen)

Also added .idea files to `.gitignore` to not commit project files.

### DIFF
--- a/src/webapi/webgl_rendering_context.rs
+++ b/src/webapi/webgl_rendering_context.rs
@@ -1,0 +1,45 @@
+use webcore::value::{Reference, ConversionError};
+use webcore::try_from::TryInto;
+use webapi::rendering_context::RenderingContext;
+use webapi::html_elements::CanvasElement;
+
+/// Used for drawing Webgl content onto the canvas element.
+///
+/// [(JavaScript docs)](https://developer.mozilla.org/en-US/docs/Web/API/WebGLRenderingContext)
+// https://html.spec.whatwg.org/#webglrenderingcontext
+pub struct WebGLRenderingContext(Reference);
+
+reference_boilerplate! {
+	WebGLRenderingContext,
+	instanceof WebGLRenderingContext
+}
+
+impl RenderingContext for WebGLRenderingContext {
+	type Error = ConversionError;
+	fn from_canvas(canvas: &CanvasElement) -> Result<Self, ConversionError> {
+		js!(
+			return @{canvas}.getContext("webgl");
+		).try_into()
+	}
+}
+	
+impl WebGLRenderingContext {
+	/// This specifies what color values to use when calling the clear() method.
+	/// The values are clamped between 0 and 1.
+	///
+	/// [(JavaScript docs)](https://developer.mozilla.org/en-US/docs/Web/API/WebGLRenderingContext/clearColor)
+	pub fn clear_color(&self, red: f64, green: f64, blue: f64, alpha: f64) {
+		js! { @(no_return)
+			@{&self.0}.clearColor(@{red}, @{green}, @{blue}, @{alpha});
+		}
+	}
+
+	/// The WebGLRenderingContext.clear() method of the WebGL API clears buffers to preset values.
+	///
+	/// [(JavaScript docs)](https://developer.mozilla.org/en-US/docs/Web/API/WebGLRenderingContext/clear)
+	pub fn clear(&self) {
+		js! { @(no_return)
+			@{&self.0}.clear(@{&self.0}.COLOR_BUFFER_BIT);
+		}
+	}
+}
--- a/.gitignore
+++ b/.gitignore
@@ -2,3 +2,4 @@ target
 Cargo.lock
 examples/*/target
 examples/*/Cargo.lock
+.idea/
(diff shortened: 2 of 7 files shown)


src/webapi/webgl_rendering_context.rs
```diff
@@ -0,0 +1,45 @@
+use webcore::value::{Reference, ConversionError};
+use webcore::try_from::TryInto;
+use webapi::rendering_context::RenderingContext;
+use webapi::html_elements::CanvasElement;
+
+/// Used for drawing Webgl content onto the canvas element.
+///
+/// [(JavaScript docs)](https://developer.mozilla.org/en-US/docs/Web/API/WebGLRenderingContext)
+// https://html.spec.whatwg.org/#webglrenderingcontext
+pub struct WebGLRenderingContext(Reference);
```
koute: To be consistent with the convention we've picked in the rest of the library this should be named `WebGlRenderingContext` (lowercase `L`)

src/webapi/webgl_rendering_context.rs
```diff
@@ -0,0 +1,45 @@
...
+
+reference_boilerplate! {
```
koute: This needs to be updated to the newest `master` where this macro was replaced by `#[derive(ReferenceType)]`

src/webapi/webgl_rendering_context.rs
```diff
@@ -0,0 +1,45 @@
...
+	WebGLRenderingContext,
+	instanceof WebGLRenderingContext
```
koute: Also, we use spaces everywhere instead of tabs. (:

src/webapi/webgl_rendering_context.rs
```diff
@@ -0,0 +1,45 @@
...
+		).try_into()
+	}
+}
+	
+impl WebGLRenderingContext {
+	/// This specifies what color values to use when calling the clear() method.
+	/// The values are clamped between 0 and 1.
+	///
+	/// [(JavaScript docs)](https://developer.mozilla.org/en-US/docs/Web/API/WebGLRenderingContext/clearColor)
+	pub fn clear_color(&self, red: f64, green: f64, blue: f64, alpha: f64) {
```
koute: Please add link to the IDL entry in the specs (just as for our other APIs.)

src/webapi/webgl_rendering_context.rs
```diff
@@ -0,0 +1,45 @@
...
+		js! { @(no_return)
+			@{&self.0}.clearColor(@{red}, @{green}, @{blue}, @{alpha});
+		}
+	}
+
+	/// The WebGLRenderingContext.clear() method of the WebGL API clears buffers to preset values.
+	///
+	/// [(JavaScript docs)](https://developer.mozilla.org/en-US/docs/Web/API/WebGLRenderingContext/clear)
+	pub fn clear(&self) {
```
koute: This should take an enum as an argument instead of having the `COLOR_BUFFER_BIT` hardcoded.

Diggsey:
@nialna I have implemented an automatic generator for these bindings: https://github.com/brendanzab/gl-rs/pull/447

Example bindings for stdweb:
https://gist.github.com/Diggsey/c8a4cb93c7d5643d20203401ead178a7

NeverGivinUp:
@Diggsey  I havn't really read it, just skimmed over it. So please excuse any unwarranted criticism. 

It seems like the generator outputs Rust code but stays very close to the C feel. Is this a technical limitation or a design decision?

For instance, there are a lot of const GLenum declarations, wheren GLenum is just a renamed u32 -- as it is in the C-oriented JavaScript. They could be broken into groups of related enums, that turn to GLenum only when serialized to JavaScript.
Also, JavaScript uses the C-oriented GLintptr type that [marks it as a platform-specific type for pointer arithmetics](https://www.opengl.org/discussion_boards/showthread.php/135776-glintptr?p=997617&viewfull=1#post997617). I wonder if there is a rustier solution.

Diggsey:
@NeverGivinUp the bindings are intended to be a thin wrapper around the javascript WebGL API. In fact I'd rather move in the opposite direction from what you are suggesting: towards an API that is more compatible with the OpenGL C api, so that it is easier to implement a WebGL backend for eg. gfx.

NeverGivinUp:
@Diggsey I'm glad it's a design decision. 

Would you kindly elaborate on the ease of implementation you have in mind? The closer it is to the C-api, the easier it might be to implement for experienced C programmers. But why would they implement something in Rust anyway? Considering Rust offers zero-cost abstraction and better type safety, wouldn't it be better to make full use of that and gain more self-documenting and more idiomatic code?

Again, I don't mean to criticize. I would like to understand your design rationale, though.

Diggsey:
@NeverGivinUp most people do not use raw OpenGL calls directly, and those that do would probably prefer to use a fully type-safe wrapper like "glium". To that end, it would be better to add support for WebGL to glium/gfx/etc. rather than trying to duplicate the work from those projects here.

Thomspoon:
Is there any work still being done on this? How does this compare to the WebGl example?

PR #112, opened 2018-02-11
Add some window size properties

JulianKniephoff:
For keeping a canvas in sync with the window size in [dubni](https://github.com/JulianKniephoff/dubni), we (the two developers, not the royal we :wink:) use the `window.innerWidth` and `window.innerHeight` properties.

Currently we have to use a `js!`-block; this PR exposes the necessary functions.

Please let me know if I missed something or can improve anything! :smiley:

### DIFF
--- a/src/webapi/window.rs
+++ b/src/webapi/window.rs
@@ -1,4 +1,5 @@
 use webcore::value::Reference;
+use webcore::try_from::TryInto;
 use webapi::event_target::{IEventTarget, EventTarget};
 use webapi::window_or_worker::IWindowOrWorker;
 use webapi::storage::Storage;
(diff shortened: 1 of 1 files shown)


koute:
Thanks a lot for the PR!

1. Please link to the IDL fields in the specs instead of to the text, e.g. instead of:

```
// https://drafts.csswg.org/cssom-view/#dom-window-innerheight
```

please put this:

```
// https://drafts.csswg.org/cssom-view/#ref-for-dom-window-innerheight
```

2. In the specs the type of those fields is `long`, so these should return `i32` instead of `f64`.

JulianKniephoff:
Done! :smiley:

koute:
LGTM, thanks!

PR #113, opened 2018-02-11
Add various console macros and removing the Error::print method

### LINKED ISSUE #55: Implement Console API
We discussed [here](https://github.com/DenisKolodin/yew/pull/39) that it's useful to have [Console](https://developer.mozilla.org/en-US/docs/Web/API/Console) methods in this crate to prevent duplication and make it reusable for other crates.

https://github.com/DenisKolodin/yew/pull/39

Pauan:
Fixes #55 

The reason why they are macros and not functions is because browsers can display objects in a special way (e.g. you can click on the arrow to expand the object's properties, etc.)

However, this only works if you pass in the object directly, so serializing to a string would prevent that.

I *could* make them functions with a type signature like `fn log<A>(input: A)`, but that limits it to a single argument. Using a macro lets it work with an unlimited number of arguments.

### DIFF
--- a/src/webapi/console.rs
+++ b/src/webapi/console.rs
@@ -1,0 +1,46 @@
+#[doc(hidden)]
+#[macro_export]
+macro_rules! js_call {
+	( $f:expr, $( $args:expr ),* ) => {{
+    	js! { @(no_return)
+    		$f( $( @{$args} ),* );
+    	}
+    	()
+    }};
+}
+
+
+#[macro_export]
+macro_rules! console_clear {
+    () => {
+    	js_call!( console.clear, );
+    };
+}
+
+#[macro_export]
+macro_rules! console_log {
+    ( $( $args:expr ),* ) => {
+    	js_call!( console.log, $( $args ),* );
+    };
+}
+
+#[macro_export]
+macro_rules! console_error {
+    ( $( $args:expr ),* ) => {
+    	js_call!( console.error, $( $args ),* );
+    };
+}
+
+#[macro_export]
+macro_rules! console_debug {
+    ( $( $args:expr ),* ) => {
+    	js_call!( console.debug, $( $args ),* );
+    };
+}
+
+#[macro_export]
+macro_rules! console_info {
+    ( $( $args:expr ),* ) => {
+    	js_call!( console.info, $( $args ),* );
+    };
+}
--- a/src/webapi/console.rs
+++ b/src/webapi/console.rs
@@ -1,141 +1,67 @@
-#[doc(hidden)]
-#[macro_export]
-macro_rules! js_call {
-    ( $f:expr, $( $args:expr ),* ) => {{
-        js! { @(no_return)
-            $f( $( @{$args} ),* );
-        }
-        ()
-    }};
-}
-
-
-/// Clears the console.
+/// Calls methods on the JavaScript `console` object.
 ///
-/// If it cannot clear the console then it does nothing.
+/// This should **only** be used for debugging purposes, its behavior is
+/// **not** standardized: it **will** vary with different browsers
+/// and Node.js.
+///
+/// The first argument is the name of the `console` method.
+///
+/// The remaining arguments can be anything which can be sent to JavaScript,
+/// and they do not need to be the same type.
+///
+/// If you want to print things to the console in a standardized way, use
+/// [`println!`](https://doc.rust-lang.org/std/macro.println.html) instead.
+///
+/// [(JavaScript docs)](https://developer.mozilla.org/en-US/docs/Web/API/console)
 ///
 /// # Examples
 ///
+/// ## clear
+///
+/// Clear the console:
+///
 /// ```rust
-/// console_clear!();
+/// console!(clear);
 /// ```
 ///
 /// [(JavaScript docs)](https://developer.mozilla.org/en-US/docs/Web/API/Console/clear)
-#[macro_export]
-macro_rules! console_clear {
-    () => {
-        js_call!( console.clear, );
-    };
-}
-
-
-/// Prints zero or more values to the console.
 ///
-/// The values can be anything that can be sent to JavaScript, and they do not
-/// need to be the same type.
-///
-/// Usually it will print a space in between each value, and it will always print
-/// a newline at the end.
-///
-/// However, this is *not guaranteed*, the printing behavior **will** vary between
-/// different browsers and Node.js, therefore you should use `console_log` *only*
-/// for debugging purposes.
-///
-/// If the first value is a string, then you can use string substitutions to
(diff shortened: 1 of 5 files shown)


Pauan:
@koute I fixed the merge conflicts.

koute:
I'm somewhat conflicted about this. On one hand I'd rather not export any new macros for things like this, on the other hand having the console methods accept multiple arguments would be useful.

Hmm... perhaps a better idea would be to make something like `fn error<T: ConsoleArgs>(args: T)`, and then impl the `ConsoleArgs` trait for `T`, `(A, B)`, `(A, B, C`, etc. so that if you wanted to call it with more than one argument you could just pass a tuple (I'm not sure myself if that is strictly better though, so I'd appreciate any input).

Pauan:
I agree that macros suck, it's unfortunate that macros cannot use Rust's normal module system and visibility rules.

From the user's perspective, using `console::log((1, 2, 3))` is a lot weirder than `console_log!(1, 2, 3)`

And personally, I think it makes sense to use macros for a variable number of arguments, trying to use traits + tuples to accomplish the same thing seems pretty hacky.

I wonder if we can somehow abuse trait overloading to make it work. Something like this...

```rust
struct Console;

trait Console1 {
    fn log<A>(arg1: A) {
        // implementation here
    }
}

trait Console2 {
    fn log<A, B>(arg1: A, arg2: B) {
        // implementation here
    }
}

impl Console1 for Console {}
impl Console2 for Console {}
```

And now you can do this:

```rust
Console.log(1);
Console.log(1, 2);
```

...but then the user has to import the various `Console*` traits to make it work. 😞 

I don't see any clean solution right now, all of the options are bad. I think the macro approach is the least-bad, but it's still pretty bad.

Pauan:
As a compromise, what about having a single `console!` macro, which accepts the log type as its first argument?

```rust
console!(log, 1, 2, 3);
console!(error, 1, 2, 3);
```

This has the benefit that it only needs to export one macro, even if we add more console functions in the future.

Pauan:
Okay, I thought about this quite a bit, and I changed it to use a single `console` macro. Now it's possible to use *all* `console` methods.

This has some advantages:

* We only need to export a single `console` macro, which is a *lot* cleaner than a bunch of `console_*` macros.

* It's **much** easier to implement.

* It's future-proof: all existing and future `console` methods will Just Work(tm) without us needing to change the code at all.

* It allows people to use non-standard `console` methods if they want to.

The documentation makes it clear that `console` should be used **only** for debugging, and that its behavior *cannot* be relied upon. I think that's sufficient warning to prevent it being abused.

koute:
Hmm.. yeah, that looks like a reasonable compromise until properly scoped macros (or variadic functions) are a thing on stable Rust.

One thing I'd like to see though is allowing the user to only call a predefined list of console methods, e.g. this should work:

```rust
console!( error, "Hello world!" );
```

but this should not:

```rust
console!( foobar, "Goodbye!" );
```

You could match on the first argument in the macro, and if it doesn't match then call `compile_error!`.

Pauan:
@koute I originally had it only match specific methods, but I decided against that for a few reasons:

* It's quite a lot more code.

* It means we have to *document* all of the methods, which is tricky because we can't document them separately, because it's a single `console` macro. And there will be a *lot* more documentation.

* It's not future-proof: if more `console` methods are added, then we have to add them into the macro.

* It means we have to make a choice whether to support non-standard methods or not. And I think we *should* have some way to use non-standard methods, because they're useful!

   With my approach, users can use non-standard methods, but it's not "officially" supported, so they use them at their own risk.

* Some environments only support some `console` methods and not others. With my approach it's clear that this macro is use-at-your-own-risk.

   But if we explicitly support specific methods, that gives the impression that those methods will work (but they might not work).

* Because `console` is used ***only*** for debugging, I figured the potential for misuse (i.e. calling it with wrong methods or wrong arguments) doesn't really matter.

But, despite all of that, if you still think that we should only allow certain methods, then I'll make the change.

Should I add in non-standard methods or not? Should I *document* non-standard methods or not?

koute:
You raised some good points, however, you have to remember that the whole point of us creating bindings to these APIs is to (among other things) supply the user with a statically checked interface. Otherwise what's the point of creating a dedicated console interface if it's just going to be a very thin wrapper around the `js!` macro? The user might as well use the `js!` macro directly and get exactly the same guarantees.

We create API bindings *precisely* because by having a dedicated APIs we can limit their domain and gain extra static guarantees that we wouldn't otherwise have. With a limited `console!` macro *if* the user makes a typo (e.g. they write `"errorr"` instead of `"error"`) a compile error will be generated; with the unchecked `console!` they'd get a runtime error, which is no better than using `js!` directly.

And it's fine if we won't support *everything*. For now it would be enough to even have `log` and `error` console APIs. We can always add more in a backward compatible way, and for the non-standard stuff there is always the `js!` macro the user can use, so we don't have to worry about that either.

Pauan:
> however, you have to remember that the whole point of us creating bindings to these APIs is to (among other things) supply the user with a statically checked interface.

I absolutely agree that strong and statically-safe APIs is a *huge* win for stdweb, but in this very *specific* situation I think things are a bit different: the API is already very "unsafe" (because it accepts any number of arguments, *and* it accepts any type, it even accepts multiple different types!) so it's not really a "proper" API in my opinion.

> Otherwise what's the point of creating a dedicated console interface if it's just going to be a very thin wrapper around the js! macro? The user might as well use the js! macro directly and get exactly the same guarantees.

The benefit of including the `console` macro is that it's a ***lot*** more convenient than the user using `js!`.

Also we need it for `PromiseFuture::spawn` (since you didn't like my `Error::print` idea), so saying "just use `js!`" isn't really a good option.

----

I'm still not looking forward to trying to cram the documentation for all the methods into a single macro, but I guess I'll implement your suggestion.

koute:
> the API is already very "unsafe" (because it accepts any number of arguments, and it accepts any type, it even accepts multiple different types!) so it's not really a "proper" API in my opinion.

Well, that's just a variadic function; it doesn't mean that it's unsafe. (:

Again, I'm not against including it, but *if* we're exposing something then I'd like to have at least a rudimentary layer of static type safety. While the *arguments* of these methods can indeed be anything the exact method name itself can *not*, so to me that's an obvious place where we can include some extra checks.

Pauan:
@koute I changed the `console` macro to only accept certain methods.

src/webapi/console.rs
```diff
@@ -0,0 +1,75 @@
+#[doc(hidden)]
+#[macro_export]
+macro_rules! console_unsafe {
```
koute: Could you rename this macro to something that's even more less likely to clash? (I don't know, `__console_internal`, or something like that?)

Pauan: Done.

koute:
LGTM! Thanks!

PR #116, opened 2018-02-12
Changing from_thenable function to use &Reference rather than Value

Pauan:


koute:
The example in the doc comment could be modified into `js!( ... ).as_reference().and_then( Promise::convert ).unwrap()` (which I think is nicer and more Rust-y) instead the current one, but that's just a minor nitpick.

LGTM, thanks!

Issue #117, opened 2018-02-12
example "echo" is broken


Unmoving:
Hey!
https://github.com/koute/stdweb/tree/master/examples/echo doesn't work for me with the newest build

I tried to fix it and open a MR, but sadly I couldn't figure out how `TryInto` is supposed to work now:

```error[E0277]: the trait bound `stdweb::web::html_element::InputElement: stdweb::unstable::TryFrom<std::option::Option<stdweb::web::Element>>` is not satisfied
  --> src/main.rs:97:86
   |
97 |     let text_entry: InputElement = document().query_selector(".form input").unwrap().try_into().unwrap();
   |                                                                                      ^^^^^^^^ the trait `stdweb::unstable::TryFrom<std::option::Option<stdweb::web::Element>>` is not implemented for `stdweb::web::html_element::InputElement`
   |
   = help: the following implementations were found:
             <stdweb::web::html_element::InputElement as stdweb::unstable::TryFrom<&'_r stdweb::Reference>>
             <stdweb::web::html_element::InputElement as stdweb::unstable::TryFrom<stdweb::web::EventTarget>>
             <stdweb::web::html_element::InputElement as stdweb::unstable::TryFrom<stdweb::Value>>
             <stdweb::web::html_element::InputElement as stdweb::unstable::TryFrom<stdweb::web::Element>>
           and 4 others
   = note: required because of the requirements on the impl of `stdweb::unstable::TryInto<stdweb::web::html_element::InputElement>` for `std::option::Option<stdweb::web::Element>```

(More or less the same for `document().query_selector(".output").unwrap().try_into().unwrap();` of course)

PR #118, opened 2018-02-12
Fixing/Updating echo example

Unmoving:
For https://github.com/koute/stdweb/issues/117

### DIFF
--- a/examples/echo/src/main.rs
+++ b/examples/echo/src/main.rs
@@ -4,7 +4,7 @@ use std::rc::Rc;
 
 use stdweb::unstable::TryInto;
 use stdweb::web::{
-	IParentNode,
+    IParentNode,
     IEventTarget,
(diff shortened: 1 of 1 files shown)


examples/echo/src/main.rs
```diff
@@ -4,6 +4,7 @@ use std::rc::Rc;
 
 use stdweb::unstable::TryInto;
 use stdweb::web::{
+	IParentNode,
```
koute: Minor nitpick - could you get rid of this tab and put spaces here instead? (:

koute:
Thanks!

examples/echo/src/main.rs
Unmoving: Hmpf, how that this little fellow pass my eagle eyes? :) Thanks for the headsup

Issue #78, opened 2018-01-14
Idea: track reference counts on rust side


Diggsey:
Currently the javascript runtime maps references to IDs and also maps IDs to reference counts. Whenever a `Reference` is cloned, the code must call out to javascript to update the reference count.

It should be possible to move the reference count onto the rust side: the javascript maps references to IDs, and then rust maps IDs to reference counts. This should make `Reference` as cheap to clone as a normal `Rc`.

When objects are serialised from javascript into rust, the rust side would be expected to increment the reference counts to avoid the javascript runtime having to perform additional calls into rust.

koute:
That might be a pretty good idea and is definitely worth exploring. (Although I have a lot of other things I want to do first before exploring any potential performance optimizations.)

PR #119, opened 2018-02-14
Implementing PartialEq + Eq for Reference, and also for all ReferenceType

### LINKED ISSUE #87: Implement PartialEq and Eq for all (most?) types
Right now `Reference` uses `#[derive(PartialEq, Eq)]`, but a lot of other types don't (e.g. `Node`, `Element`, `NodeList`, etc.)

In most cases implementing it should be as simple as adding `#[derive(PartialEq, Eq)]` to the types.

Assuming you're okay with that, I'll send a pull request adding this in.

----

Also, I took a look at the implementation of `Module.STDWEB`, but I have a question. It implements `Reference` with an `i32`, and the `PartialEq` implementation simply compares the `i32`

Is it possible to have two `Reference` that have different `i32`, but they refer to the same JS object?

In other words: if `ref1 == ref2` is false in Rust, is it possible for `ref1 === ref2` to be true in JavaScript?

Because `Module.STDWEB` uses a `WeakMap` for `ref_to_id_map`, it *seems* that situation will never happen, but is it truly impossible?

Pauan:
Fixes #87

### DIFF
--- a/src/webapi/event_target.rs
+++ b/src/webapi/event_target.rs
@@ -7,6 +7,7 @@ use webapi::event::{ConcreteEvent, IEvent};
 use private::UnimplementedException;
 
 /// A handle to a particular event listener.
+// TODO derive PartialEq, Eq, etc. ?
 pub struct EventListenerHandle {
     event_type: &'static str,
     reference: Reference,
--- a/src/webapi/event_target.rs
+++ b/src/webapi/event_target.rs
@@ -7,7 +7,6 @@ use webapi::event::{ConcreteEvent, IEvent};
 use private::UnimplementedException;
 
 /// A handle to a particular event listener.
-// TODO derive PartialEq, Eq, etc. ?
 pub struct EventListenerHandle {
     event_type: &'static str,
     reference: Reference,
--- a/src/webapi/event_target.rs
+++ b/src/webapi/event_target.rs
@@ -80,7 +80,7 @@ pub trait IEventTarget: ReferenceType {
 ///
 /// [(JavaScript docs)](https://developer.mozilla.org/en-US/docs/Web/API/EventTarget)
 // https://dom.spec.whatwg.org/#eventtarget
-#[derive(Clone, Debug, ReferenceType)]
+#[derive(Clone, Debug, PartialEq, Eq, ReferenceType)]
 #[reference(instance_of = "EventTarget")]
 pub struct EventTarget( Reference );
 
--- a/src/webapi/mutation_observer.rs
+++ b/src/webapi/mutation_observer.rs
@@ -9,7 +9,7 @@ use private::UnimplementedException;
 ///
 /// [(JavaScript docs)](https://developer.mozilla.org/en-US/docs/Web/API/MutationObserver)
 // https://dom.spec.whatwg.org/#mutationobserver
-#[derive(Clone, Debug, ReferenceType)]
+#[derive(Clone, Debug, PartialEq, Eq, ReferenceType)]
 #[reference(instance_of = "MutationObserver")]
 pub struct MutationObserver( Reference );
 
--- a/src/webapi/mutation_observer.rs
+++ b/src/webapi/mutation_observer.rs
@@ -161,6 +161,7 @@ impl MutationObserver {
 ///
 /// When the `MutationObserverHandle` is dropped, the [`disconnect`](#method.disconnect)
 /// method will automatically be called.
+// TODO derive PartialEq, Eq, etc. ?
 #[ derive( Debug ) ]
 pub struct MutationObserverHandle {
     mutation_observer: MutationObserver,
@@ -239,6 +240,7 @@ pub enum MutationRecord {
     },
 }
 
+// TODO create a MutationRecord Reference and use instanceof to verify it
 impl TryFrom< Value > for MutationRecord {
     type Error = ConversionError;
 
--- a/src/webapi/mutation_observer.rs
+++ b/src/webapi/mutation_observer.rs
@@ -161,7 +161,6 @@ impl MutationObserver {
 ///
 /// When the `MutationObserverHandle` is dropped, the [`disconnect`](#method.disconnect)
 /// method will automatically be called.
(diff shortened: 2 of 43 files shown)


src/webapi/event_target.rs
```diff
@@ -7,6 +7,7 @@ use webapi::event::{ConcreteEvent, IEvent};
 use private::UnimplementedException;
 
 /// A handle to a particular event listener.
+// TODO derive PartialEq, Eq, etc. ?
```
koute: It should be enough to compare `listener_reference` (since those are going to be unique for every `EventListenerHandle`) for `PartialEq` for this.

src/webcore/value.rs
```diff
@@ -51,6 +51,15 @@ impl Reference {
     }
 }
 
+impl PartialEq for Reference {
+    fn eq( &self, other: &Reference ) -> bool {
+        js!( return @{self} === @{other}; ).try_into().unwrap()
```
koute: Hmm.. this is probably unnecessary considering the performance hit it's going to incur.

How about something like this?

```rust
let result = self.0 == other.0;
debug_assert_eq!( js! ( return @{self} === @{other}; ).try_into().unwrap(), true );
result
```

In release mode it's going to be super cheap, while in debug mode we'll get an extra check to make sure this behaves as we think it should.

koute: Also, `#[inline]`.

src/webcore/value.rs
```diff
@@ -51,6 +51,15 @@ impl Reference {
...
+    }
+}
+
+// TODO is this guaranteed to be true? e.g. is it possible to have an f64 that is a Reference ?
```
koute: I think it should be safe to have `Eq` implemented for this. (The refids are always going to be integers.)

src/webapi/mutation_observer.rs
```diff
@@ -161,6 +161,7 @@ impl MutationObserver {
 ///
 /// When the `MutationObserverHandle` is dropped, the [`disconnect`](#method.disconnect)
 /// method will automatically be called.
+// TODO derive PartialEq, Eq, etc. ?
```
koute: This can compare the `callback_reference`.

src/webapi/event_target.rs
Pauan: The comment wasn't about *how* to implement `PartialEq`, it was about whether we should implement `PartialEq` *at all* on that type.

src/webcore/value.rs
Pauan: Good catch about `#[inline]`. But I'm confused: `js!` is used *constantly*, why are you worried about the performance *only for `PartialEq`*?

Also, I'm not sure if it actually *is* unnecessary. In issue #92 they pointed out that some JavaScript objects cannot be put into `WeakMap`, so we'll have to rethink the `Reference` implementation. And in any case relying upon `WeakMap` to guarantee correctness feels wrong to me.

Using `===` guarantees that it will ***always*** match the JavaScript behavior, without relying upon things like `WeakMap` (which might change or be refactored in the future!)

src/webcore/value.rs
Pauan: With my implementation of `PartialEq`, it's not comparing refids, it's using JavaScript's `===`, which is why I put that comment there.

src/webapi/mutation_observer.rs
Pauan: Like I said above, the question isn't *how* to implement it, it's whether to implement it *at all*.

src/webcore/value.rs
```diff
@@ -51,6 +51,16 @@ impl Reference {
     }
 }
 
+impl PartialEq for Reference {
+    #[inline]
+    fn eq( &self, other: &Reference ) -> bool {
+        js!( return @{self} === @{other}; ).try_into().unwrap()
```
koute: Hm... I remember putting a comment regarding this, but I can't find it now...? Did it disappear or was it not posted in the first place?

Pauan: You have to click on the "Show outdated" thingy. That happened because I changed the code, so GitHub "helpfully" assumes that your old comments are now outdated because of the code changes.

src/webcore/value.rs
koute: If we have an object which cannot be put in a `WeakMap` then even when comparing it through the `js!` macro it's going to be subtly broken anyway (e.g. some of the conversion functions between Rust and JS won't work)

In general since they're so common I'd like to make `Reference`s cheaper to use in the future (e.g. by moving the refcounting into Rust https://github.com/koute/stdweb/issues/78). As for those special snowflake types which cannot be put in a `WeakMap` - I'd rather come up with some separate mechanism for those, so that only those types will incur the extra overhead instead of everything.

src/webcore/value.rs
koute: Ah right, I didn't notice the second outdated!

src/webcore/value.rs
koute: Well, no primitive types should ever end up as `Reference`s, so in either case it should be safe.

src/webapi/mutation_observer.rs
koute: I don't see why not?

Pauan: It's a commitment to the public API. If we later change the internal implementation it might change the behavior (or even make comparing for equality impossible).

Implementing `PartialEq` for `ReferenceType` is fine, because its behavior is stable, but the Handle types are different.

koute: Since *we* need a stable handle inside of this I think this always should be comparable?

But if you'd rather be conservative on this I don't mind. (I don't have strong feelings either way.)

Pauan:
@koute I improved the performance of `PartialEq`.

I also decided *not* to implement `PartialEq` and `Eq` for the Handle types. I couldn't think of any reason for comparing them for equality, and in any case we can always add in equality later.

I think this is ready to merge now.

koute:
LGTM! Thanks!

PR #122, opened 2018-02-14
Adding in default type argument to PromiseFuture

Pauan:
99+% of the time Promises have a JavaScript `Error` object as their error.

In addition, Futures require the error type to be the same when using combinators (such as `and_then`), so it's beneficial to encourage the use of a single `Error` type (the alternative is to constantly cast between error types, which is pretty verbose).

Therefore, I think it makes sense to make the error type of `PromiseFuture` optional (it defaults to `webapi::error::Error`). This solves both of the above problems, while still allowing for error customization in the rare cases where it is needed.

### DIFF
--- a/examples/promise/src/main.rs
+++ b/examples/promise/src/main.rs
@@ -51,10 +51,10 @@ fn test_error_conversion() {
         } )
     );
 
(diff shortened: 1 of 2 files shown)


koute:
LGTM, thanks!

PR #126, opened 2018-02-15
Add basic geolocation API interface.

brandonedens:


src/webapi/geolocation.rs
```diff
@@ -0,0 +1,135 @@
...
+//! Note that this implementation currently doesn't handle the failure to get position callbacks
+//! nor does it handle geo_options.
+
+use webcore::value::{Reference, Value};
+use webcore::try_from::TryInto;
+
+/// Representation of positional coordinate information.
+#[derive(Clone, Debug, ReferenceType)]
+#[reference(instance_of = "Object")]
+pub struct Coordinates(Reference);
```
koute: Hmm... I'd probably prefer if `Coordinates` and `Position` were normal Rust `struct`s instead of `Reference`s.

src/webapi/geolocation.rs
```diff
@@ -0,0 +1,135 @@
...
+impl Geolocation {
+    /// Attempt to get current position and invoke callback on success.
+    pub fn get_current_position<F: FnOnce(Position) + 'static>(callback: F) {
+        js! (
+            navigator.geolocation.getCurrentPosition(@{Once(callback)})
+            );
+    }
+
+    /// Watch for position changes and call function with updates.
+    pub fn watch_position<F: Fn(Position) + 'static>(callback: F) -> WatchId {
```
koute: This also needs error handling; the callback should probably be something like `FnMut(Result<Position, PositionError>)`.

src/webapi/geolocation.rs
```diff
@@ -0,0 +1,135 @@
...
+            ).try_into()
+            .unwrap()
+    }
+}
+
+/// Watch identifier.
+#[derive(Debug)]
+pub struct WatchId(Value);
+
+impl ::std::default::Default for WatchId {
```
koute: This shouldn't really have `Default` implemented. If the user wants a nullable type they can wrap it in an `Option`.

src/webapi/geolocation.rs
```diff
@@ -0,0 +1,135 @@
...
+    }
+}
+
+/// API that provides interface to geographic location information.
+#[derive(Debug)]
+pub struct Geolocation;
+
+use webcore::once::Once;
+impl Geolocation {
+    /// Attempt to get current position and invoke callback on success.
```
koute: Please add:

1. Links to the MDN (just like in other APIs we have, see them for an example)
2. Links to the IDL in the specs (just like in most of the other APIs we have)

src/webapi/geolocation.rs
brandonedens: Having a hard time implementing this change. My design shown below where I go from Javascript context to Rust to Javascript compiles but fails at runtime with cryptic message that identifier p is not known:

```
 /// Attempt to get current position and invoke callback on success.
-pub fn get_current_position<F: FnOnce(Position) + 'static>(callback: F) {
+pub fn get_current_position<F: FnMut(Result<Position, PositionError>) + 'static>(mut callback: F) {
+
     js! (
-        navigator.geolocation.getCurrentPosition(@{Once(callback)})
+        navigator.geolocation.getCurrentPosition(function(p) {
+            @{callback(Ok(Position(js!(p).into_reference().unwrap())))};
+        }, function(err) {
+            @{callback(Err(position_err_code_to_enum(&js!(err.code))))};
+        })
     );
 }
```

src/webapi/geolocation.rs
brandonedens: They're reference right now because they're wrapping over values returned from the Javascript API. Should I take the values from the Javascript API and reencode them into Rust and pass that to the user?

src/webapi/geolocation.rs
brandonedens: Switched to an Option API per your suggestion.

src/webapi/geolocation.rs
```diff
@@ -0,0 +1,135 @@
...
+
+impl ::std::default::Default for WatchId {
+    fn default() -> Self {
+        WatchId(Value::Undefined)
+    }
+}
+
+/// API that provides interface to geographic location information.
+#[derive(Debug)]
+pub struct Geolocation;
```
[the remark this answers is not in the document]
brandonedens: Done.

src/webapi/geolocation.rs
```diff
@@ -0,0 +1,135 @@
...
+
+use webcore::once::Once;
```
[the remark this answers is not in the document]
brandonedens: Done.

src/webapi/geolocation.rs
```diff
@@ -0,0 +1,135 @@
...
+impl Geolocation {
+    /// Attempt to get current position and invoke callback on success.
+    pub fn get_current_position<F: FnOnce(Position) + 'static>(callback: F) {
+        js! (
+            navigator.geolocation.getCurrentPosition(@{Once(callback)})
+            );
```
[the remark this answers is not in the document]
brandonedens: Done.

src/webapi/geolocation.rs
```diff
@@ -0,0 +1,135 @@
...
+    }
+
+    /// Watch for position changes and call function with updates.
+    pub fn watch_position<F: Fn(Position) + 'static>(callback: F) -> WatchId {
+        WatchId(js! (
+            navigator.geolocation.watchPosition(@{callback})
```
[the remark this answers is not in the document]
brandonedens: Done.

src/webapi/geolocation.rs
```diff
@@ -0,0 +1,135 @@
...
+            ))
+    }
+
+    /// Clear watch disabling callback on position updates.
+    pub fn clear_watch(watch: &WatchId) {
```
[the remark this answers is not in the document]
brandonedens: Done.

src/webapi/geolocation.rs
koute: Yep. In this case I don't really see much point in keeping these as references.

src/webapi/geolocation.rs
koute: This is AFAIK triggered only once so it should take `FnOnce` instead of `FnMut`.

You're getting an error message as everything inside `@{...}` is Rust code, so inside of the `@{...}` block you're using the `js!` macro again and referring to the `p` variable which is not defined in Rust.

You could probably do something like this:

```
let wrapper = move |is_error: bool, value: Value| {
    // ... call callback here ...
});
js!(
    var callback = @{Once(wrapper)};
    navigator.geolocation.getCurrentPosition(function(pos) {
        callback(false, pos);
    }, function(err) {
        callback(true, pos);
    });
);
```

src/webapi/geolocation.rs
```diff
@@ -0,0 +1,116 @@
...
+
+/// Handle to position watch.
+#[derive(Debug)]
+pub struct GeoWatchHandle(Value);
+
+/// Watch for position changes and call function with updates.
+pub fn watch_position<F: FnMut(Position) + 'static>(callback: F) -> GeoWatchHandle {
+    GeoWatchHandle(js! (
+            navigator.geolocation.watchPosition(@{callback})
+            ))
```
koute: Broken indentation. (:

koute:
@brandonedens Ping. Will you be still working on this? (:

Boscop:
Any update on this? :)

koute:
@Boscop Doesn't look like it. If you need this functionality then perhaps you would be interested in continuing this PR? (:

PR #137, opened 2018-02-26
RFC: Web API Guidelines 

teovoinea:
Opening the discussion for code style/guidelines when adding support for new Web APIs.


koute:
This looks like a good start! Thanks!

A few comments:
1. Please change the filename to `CONTRIBUTING.md`.
2. Move the contents of `Contributing` section from the README here, and add a remark in the README that this file exists.
3. I think it would be worthwhile to enumerate the rules/tips in form of a numbered list.
4. For every rule it would be nice to have a short one liner (which would be **bold** or in a bigger font) which summarizes the rule, and then a more thorough explanation in a separate paragraph.
5. The order of the concrete guidelines should be roughly in their order of importance, grouped by category. So, e.g. stuff like "make sure you follow the original JS spec" should be put first, and the explanation on how to write the docs after.

koute:
Although I'll probably edit this a little further it looks good to me now! Thanks!

teovoinea:
@koute I just found a quick issue, the link for CONTRIBUTING.md in the README is broken

koute:
Oops, you're right! That's what I get reviewing stuff at 1 AM. (:

I'll fix it myself real quick so don't worry about it.

PR #139, opened 2018-02-27
Expose WebSocket BinaryType struct

brandonhamilton:
As far as I can tell, the `BinaryType` struct needs to be accessible to pass a value into the `WebSocket::set_binary_type` function.

### DIFF
--- a/src/webapi/web_socket.rs
+++ b/src/webapi/web_socket.rs
@@ -77,7 +77,7 @@ impl IEventTarget for WebSocket {}
 /// [(JavaScript docs)](https://developer.mozilla.org/en-US/docs/Web/API/WebSocket#Attributes)
 // https://html.spec.whatwg.org/#dom-websocket-binarytype
 #[derive(Copy, Clone, Debug, PartialEq, Eq, PartialOrd, Ord, Hash)]
-pub enum BinaryType {
+pub enum SocketBinaryType {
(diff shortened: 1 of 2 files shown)


src/webapi/web_socket.rs
```diff
@@ -72,10 +72,17 @@ pub struct WebSocket( Reference );
...
 // https://html.spec.whatwg.org/#dom-websocket-binarytype
 #[derive(Copy, Clone, Debug, PartialEq, Eq, PartialOrd, Ord, Hash)]
 pub enum BinaryType {
+    /// A Blob object represents a file-like object of immutable, raw data.
+    /// [(Javascript docs)](https://developer.mozilla.org/en-US/docs/Web/API/Blob)
     Blob,
+    /// The ArrayBuffer object is used to represent a generic, fixed-length raw binary data buffer.
+    /// [(Javascript docs)](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/ArrayBuffer)
     ArrayBuffer
 }
```
koute: Can you rename it to `SocketBinaryType` and export it without the `web_socket` module? (Since we **don't** export other websocket related stuff in any namespace it'd be a little strange to make it only for this one single type.)

brandonhamilton: Yes, will do !

koute:
LGTM! Thanks!

PR #142, opened 2018-02-28
Add js getTimezoneOffset() method.

fryguy1013:
Add js getTimezoneOffset() method, so that libraries like `time` can import this for a stdweb implementations.

### DIFF
--- a/src/webapi/date.rs
+++ b/src/webapi/date.rs
@@ -1,13 +1,94 @@
-use std::marker::PhantomData;
+use webcore::value::Reference;
 use webcore::try_from::TryInto;
 
 /// [(JavaScript docs)](https://developer.mozilla.org/en/docs/Web/JavaScript/Reference/Global_Objects/Date)
-#[derive(Debug)]
-pub struct Date {
-    dummy: PhantomData< () >
-}
+/// https://www.ecma-international.org/ecma-262/6.0/#sec-date-constructor
+#[derive(Clone, Debug, PartialEq, Eq, ReferenceType)]
+#[reference(instance_of = "Date")]
+pub struct Date( Reference );
 
 impl Date {
+    /// Creates a JavaScript Date instance that represents a single moment in time.
+    /// Date objects are based on a time value that is the number of milliseconds since 1 January 1970 UTC.
+    ///
+    /// [(JavaScript docs)](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date)
+    // https://www.ecma-international.org/ecma-262/6.0/#sec-date-constructor-date
+    pub fn new() -> Self {
+        js!(
+            return new Date();
+        ).try_into().unwrap()
+    }
+
+   
+    /// Creates a JavaScript Date instance that represents a single moment in time.
+    /// Date objects are based on a time value that is the number of milliseconds since 1 January 1970 UTC.
+    /// 
+    /// year is an integer value representing the year. Values from 0 to 99 map to the years 1900 to 1999.
+    /// month is an integer value representing the month, beginning with 0 for January to 11 for December
+    /// day is an integer value representing the day of the month (normally from 1 to 31)
+    /// hours an integer value representing the minute segment of a time
+    /// seconds an integer value representing the second segment of a time
+    /// milliseconds an integer value representing the millisecond segment of a time
+    ///
+    /// [(JavaScript docs)](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date)
+    // https://www.ecma-international.org/ecma-262/6.0/#sec-date-year-month-date-hours-minutes-seconds-ms
+    pub fn from_datetime(year: i32, month: i32, day: i32, hours: i32, minutes: i32, seconds: i32, milliseconds: i32) -> Self {
+        js!(
+            return new Date(@{year}, @{month}, @{day}, @{hours}, @{minutes}, @{seconds}, @{milliseconds});
+        ).try_into().unwrap()
+    }
+
+    /// Creates a JavaScript Date instance that represents a single moment in time.
+    /// Date objects are based on a time value that is the number of milliseconds since 1 January 1970 UTC.
+    /// 
+    /// String value representing a date. The string should be in a format recognized by
+    /// the Date.parse() method (IETF-compliant RFC 2822 timestamps and also a version of ISO8601).
+    ///
+    /// [(JavaScript docs)](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date)
+    // https://www.ecma-international.org/ecma-262/6.0/#sec-date-value
+    pub fn from_iso8601(date_string: &str) -> Self {
+        js!(
+            return new Date(@{date_string});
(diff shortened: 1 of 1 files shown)


src/webapi/date.rs
```diff
@@ -17,6 +17,18 @@ impl Date {
...
+	///
+	/// [(JavaScript docs)](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/getTimezoneOffset)
+	pub fn timezone_offset_minutes() -> i32 {
+		let ret = js! {
+			return new Date().getTimezoneOffset();
+		};
+		let ret: i32 = ret.try_into().unwrap_or(0);
+		ret
+	}
+
```
koute: 1. This needs a link to the IDL in the specs (see other bindings for how it should be done)
2. The name should be `get_timezone_offset` to stay as close as possible to the original JS name.
3. `unwrap_or` is unnecessary; it should simply be `unwrap`
4. This method should take `&self` (you'd have to add `new()` too to be able to construct a new `Date`)

fryguy1013: I couldn't find an idl for the Date object. The spec linked in the other method was as close as I could find (https://www.ecma-international.org/ecma-262/6.0/#sec-date-constructor).

I didn't realize the intention was to match the API as closely as possible to the JS API for this project. I am in the middle of adding all of the methods on the date object for completeness sake then. I'm assuming that it should also inherit the warts of the JS API. For instance, month being 0-based and the days being 1-based, and calling the get_year method get_full_year because the underlying getYear() isn't Y2K-compliant, and they made a new function called get_full_year.

Also, I'm not a rustacean enough to know what's good names for the various constructors. So far I have:

* new()
* new_with_date(year: i32, month: i32, day: i32)
* new_with_time(year: i32, month: i32, day: i32, hours: i32, minutes: i32, seconds: i32)
* new_with_time_and_millis(year: i32, month: i32, day: i32, hours: i32, minutes: i32, seconds: i32, milliseconds: i32)
* new_with_string(date_string: &str)
* new_with_now(now: f64)

Do these seem good, or should I come up with something else?

fryguy1013:
Ok this should be a much more thorough pass at the Date() JS object.

Some of the setters seem to return the new internal value, which I left out.

Also some of the setters (setFullDate, etc) have optional parameters, which I didn't provide overloads for since I didn't want to come up with names for them, and they're somewhat redundant.

I didn't provide the toLocaleXXXString() methods, as I didn't know the best way to pass along all the additional options from Rust to JS.

I'm not sure about the Date::new_X() naming.

The documentation for the functions is copy/pasted from developer.mozilla.org. I'm not sure about the license for that, but it seemed to be done in other places, so I followed suit.

src/webapi/date.rs
```diff
@@ -1,13 +1,117 @@
...
+    /// 
+    /// Integer value representing the number of milliseconds since January 1, 1970, 00:00:00 UTC,
+    /// with leap seconds ignored (Unix Epoch; but consider that most Unix timestamp functions count in seconds).
+    /// [(JavaScript docs)](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date)
+    // https://www.ecma-international.org/ecma-262/6.0/#sec-date-value
+    pub fn new_with_now(now: f64) -> Self {
+        js!(
+            return new Date(@{now});
+        ).try_into().unwrap()
+    }
```
koute: Hmm.... I'd like a better name for this one, but it's somewhat hard to come up with one. `from_time`? (To be consistent with `set_time`.) The `new_with_now` sounds kinda strange.

src/webapi/date.rs
```diff
@@ -1,13 +1,117 @@
...
+    /// hours an integer value representing the minute segment of a time
+    /// seconds an integer value representing the second segment of a time
+    /// milliseconds an integer value representing the millisecond segment of a time
+    /// [(JavaScript docs)](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date)
+    // https://www.ecma-international.org/ecma-262/6.0/#sec-date-year-month-date-hours-minutes-seconds-ms
+    pub fn new_with_time_and_millis(year: i32, month: i32, day: i32, hours: i32, minutes: i32, seconds: i32, milliseconds: i32) -> Self {
+        js!(
+            return new Date(@{year}, @{month}, @{day}, @{hours}, @{minutes}, @{seconds}, @{milliseconds});
+        ).try_into().unwrap()
+    }
```
koute: Hmm.. could we get rid of `new_with_time` and `new_with_date` for now and keep only this method? And perhaps rename it to something like `from_datetime`? (`datetime` name is also used in `chrono`; we can always add other methods later **if** there will really be a need for them)

src/webapi/date.rs
```diff
@@ -1,13 +1,117 @@
...
+    /// 
+    /// String value representing a date. The string should be in a format recognized by
+    /// the Date.parse() method (IETF-compliant RFC 2822 timestamps and also a version of ISO8601).
+    /// [(JavaScript docs)](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date)
+    // https://www.ecma-international.org/ecma-262/6.0/#sec-date-value
+    pub fn new_with_string(date_string: &str) -> Self {
+        js!(
+            return new Date(@{date_string});
+        ).try_into().unwrap()
+    }
```
koute: `new_with_string` -> `from_iso8601`? (I know that it also support implementation-defined heuristics, but I'd rather not encourage that; the only portable way to use it is by assuming ISO 8601)

src/webapi/date.rs
```diff
@@ -1,13 +1,117 @@
...
+    // https://www.ecma-international.org/ecma-262/6.0/#sec-date-value
+    pub fn new_with_now(now: f64) -> Self {
+        js!(
+            return new Date(@{now});
+        ).try_into().unwrap()
+    }
+
+    /// The Date.UTC() method accepts the same parameters as the longest form of the constructor, and
+    /// returns the number of milliseconds in a Date object since January 1, 1970, 00:00:00, universal time.
+    /// [(JavaScript docs)](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/UTC)
```
koute: Please add an extra newline before `[(JavaScript docs)]` line. (Ditto for other doc comments where it's missing.)

src/webapi/date.rs
```diff
@@ -17,10 +121,703 @@ impl Date {
...
 }
 
 #[test]
 fn test_date_now() {
     let now = Date::now();
     assert!( now > 0.0 );
 }
+
+#[test]
+fn test_date_utc() {
```
koute: Can you put these tests in their own module? (Mostly to be consistent with the rest of the tests we have.)

```
#[cfg(test)]
mod tests {
...
}
```

fryguy1013:
Ok I've updated with your suggestions. I had been looking at the WebSocket implementation which led me a little astray (new_with_protocols, and a test module that was feature bound to web_test which wasn't actually running my tests when I copy/pasted the `use module` lines from it originally before submitting it)

koute:
LGTM; thanks!

PR #144, opened 2018-03-01
Add getBoundingClientRect

teovoinea:
The function isn't part of the HTML Spec yet but all browsers have at least basic support.

https://developer.mozilla.org/en-US/docs/Web/API/Element/getBoundingClientRect


src/webapi/html_element.rs
```diff
@@ -43,6 +116,14 @@ pub trait IHtmlElement: IElement {
         }
     }
 
+    /// Returns the size of an element and its position relative to the viewport.
+    /// [(JavaScript docs)](https://developer.mozilla.org/en-US/docs/Web/API/Element/getBoundingClientRect)
```
koute: Hmm... it looks like it **is** in the specs.

https://drafts.csswg.org/cssom-view/#ref-for-dom-element-getboundingclientrect

teovoinea:
@koute Should be good to go :)

koute:
Thanks!

Issue #148, opened 2018-03-04
Memory leak


kuviman:
Looks like there is a bug in refcounting `Reference`s to JS objects

refcount is [initially set to 1](https://github.com/koute/stdweb/blob/85f7aa484f3be9808b992b9258bd85fa588e05b1/src/webcore/runtime.js#L308), and incremented [when constructing](https://github.com/koute/stdweb/blob/573c7ce5894087c81170bd1eca2a4a5ead8f2d15/src/webcore/value.rs#L33) a `Reference`, but only decremented once [while dropping](https://github.com/koute/stdweb/blob/573c7ce5894087c81170bd1eca2a4a5ead8f2d15/src/webcore/value.rs#L83) the `Reference`

This code leaves `Module.STDWEB_PRIVATE.id_to_refcount_map[1] = 100`:
```rust
fn main() {
    js! { Module.test = {}; }
    for _ in 0..100 {
        js! { return Module.test; }
    }
}
```

That leads to leaking JS objects.

See also [this comment](https://github.com/koute/stdweb/issues/92#issuecomment-370231597), fixing this would probably break IDs for same objects, since [they would be deleted](https://github.com/koute/stdweb/blob/85f7aa484f3be9808b992b9258bd85fa588e05b1/src/webcore/runtime.js#L332) from the WeakMap when reference count reaches zero, but there are still existing references in JS.

Pauan:
I think we just shouldn't guarantee that the same id corresponds to the same JS object.

That does mean that the implementation of `PartialEq` on all `Reference`s will be slower though.

kuviman:
@Pauan I thought it is actually a bug that objects are deleted from `ref_to_id_map`.

If they were not deleted, same ids would be guaranteed, and right now they are the same only because of the incorrect refcounting.

Pauan:
@kuviman You're right, it should never delete from `ref_to_id_map`: it's a `WeakMap`, so it will automatically garbage collect the keys when they are no longer accessible in Rust or JavaScript.

PR #149, opened 2018-03-05
Fix memory leak on js! block

yjh0502:
Fix https://github.com/koute/stdweb/issues/148

It seems that refcount for JS reference is incremented before serialized to wasm buffer[1], so refcount will be incresed twice if `Reference::from_raw_unchecked` is used on `deserialize`.

[1] https://github.com/koute/stdweb/blob/85f7aa484f3be9808b992b9258bd85fa588e05b1/src/webcore/runtime.js#L303-L311

### DIFF
--- a/src/webcore/value.rs
+++ b/src/webcore/value.rs
@@ -36,6 +36,12 @@ impl Reference {
 
     #[doc(hidden)]
     #[inline]
+    pub(super) unsafe fn from_raw_unchecked_noref( refid: i32 ) -> Reference {
+        Reference( refid )
+    }
+
+    #[doc(hidden)]
+    #[inline]
     pub fn as_raw( &self ) -> i32 {
         self.0
     }
@@ -1290,5 +1296,59 @@ mod tests {
 
         let a: Result< (), _ > = js! { return 1; }.try_into();
         assert!( a.is_err() );
+    }
+
+    #[test]
+    fn reference_stable() {
+        js! { Module.__test = {}; }
+        let a = js! { return Module.__test; }.as_reference().unwrap().as_raw();
+        let b = js! { return Module.__test; }.as_reference().unwrap().as_raw();
+        assert_eq!(a, b);
+
+        let c = js! { return {}; }.as_reference().unwrap().as_raw();
+        assert_ne!(a, c);
+
+        js! { delete Module.__test; }
+    }
+
+    fn is_known_reference(refid: i32) -> bool {
+        let has_refcount: bool = js! {
+            return @{refid} in Module.STDWEB_PRIVATE.id_to_refcount_map;
+        }.try_into().unwrap();
+
+        let has_ref: bool = js! {
+            return @{refid} in Module.STDWEB_PRIVATE.id_to_ref_map;
+        }.try_into().unwrap();
+
+        assert_eq!(has_refcount, has_ref);
+        has_refcount
+    }
+
+    #[test]
+    fn reference_refcount() {
+        let obj = js! { return new Object(); };
+        let refid = obj.as_reference().unwrap().as_raw();
+        assert!(is_known_reference(refid));
+
+        drop(obj);
+        assert!(!is_known_reference(refid));
+    }
+
+    #[test]
+    fn reference_refcount_clone() {
+        let obj = js! { return new Object(); };
+        let obj2 = obj.clone();
+
+        let refid = obj.as_reference().unwrap().as_raw();
+        let refid2 = obj.as_reference().unwrap().as_raw();
+
+        assert_eq!(refid, refid2);
+        assert!(is_known_reference(refid));
+
+        drop(obj);
+        assert!(is_known_reference(refid));
+
+        drop(obj2);
+        assert!(!is_known_reference(refid));
     }
 }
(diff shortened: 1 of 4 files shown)


koute:
Thanks for the PR!

Hmm... it'd be great to have a unit test for this so that we can avoid reintroducing something like this by mistake in the future. Would you be up to writing one?

kuviman:
As mentioned in the issue, this breaks stable ids, here's code that breaks after this PR.

```rust
#[macro_use]
extern crate stdweb;

#[test]
fn test() {
    js! { Module.test = {}; }
    let a = js! { return Module.test; }.as_reference().unwrap().as_raw();
    let b = js! { return Module.test; }.as_reference().unwrap().as_raw();
    assert_eq!(a, b);
}
```

This can be fixed though by not deleting from `ref_to_id_map`.

yjh0502:
@koute I added testcases. A testcases relies on runtime details, like `Module.STDWEB_PRIVATE.id_to_refcount_map`. Is it okay?

@kuviman I changed the patch to retain reference ids.

src/webcore/value.rs
```diff
@@ -1291,4 +1297,50 @@ mod tests {
...
+        assert!(!has_ref);
+    }
+
+    #[test]
+    fn reference_refcount() {
+        let refid = {
+            let obj = js! { return new Object(); };
+            obj.as_reference().unwrap().as_raw()
+        };
+        assert_noref(refid);
```
koute: Could you make this more explicit? e.g.:

```
let value = js! { return new Object(); }.into_reference().unwrap();
let refid = value.as_raw();
assert_eq!( is_known_reference(refid), true );
mem::drop( value );
assert_eq!( is_known_reference(refid), false );
```

Basically, to explicitly show that dropping the `Reference` will also remove it from the map.

src/webcore/value.rs
```diff
@@ -1291,4 +1297,50 @@ mod tests {
...
+
+        js! { delete Module.__test; }
+        js! { Module.__test = {}; }
+        let c = js! { return Module.__test; }.as_reference().unwrap().as_raw();
+        assert_ne!(a, c);
+
+        js! { Module.__test = {}; }
+    }
+
+    fn assert_noref(refid: i32) {
```
koute: I think this would be better as a function which returns `bool` - then you could test the inverse. (Also, perhaps a more descriptive name? `is_known_reference`? These are the tests, so we don't care if it's long.)

src/webcore/value.rs
```diff
@@ -1291,4 +1297,50 @@ mod tests {
...
+        let a = js! { return Module.__test; }.as_reference().unwrap().as_raw();
+        let b = js! { return Module.__test; }.as_reference().unwrap().as_raw();
+        assert_eq!(a, b);
+
+        js! { delete Module.__test; }
+        js! { Module.__test = {}; }
+        let c = js! { return Module.__test; }.as_reference().unwrap().as_raw();
+        assert_ne!(a, c);
+
+        js! { Module.__test = {}; }
```
koute: Shouldn't this be `delete`d at the end?

src/webcore/value.rs
```diff
@@ -1291,4 +1297,50 @@ mod tests {
...
+    fn reference_stable() {
+        js! { Module.__test = {}; }
+        let a = js! { return Module.__test; }.as_reference().unwrap().as_raw();
+        let b = js! { return Module.__test; }.as_reference().unwrap().as_raw();
+        assert_eq!(a, b);
+
+        js! { delete Module.__test; }
+        js! { Module.__test = {}; }
+        let c = js! { return Module.__test; }.as_reference().unwrap().as_raw();
+        assert_ne!(a, c);
```
koute: Is there any point in reassigning the `Module.__test` now? This might just as well be replaced with `js! { return {}; }`. (Since the fact that we're returning `Module.__test` doesn't matter; what matters here is that we `deleted`d the original `{}` and we returned a new `{}`.)

koute:
Yeah, depending on the internals in tests is fine.

Also, you might want to get rid of those `.swp` files you've accidentially `git add`ed. (: (You can add `*.swp` to `.gitignore` if you want.)

src/webcore/value.rs
yjh0502: I changed the patch to use `drop` explicitly.

src/webcore/value.rs
yjh0502: Fixed.

src/webcore/value.rs
yjh0502: Fixed. Thanks.

src/webcore/value.rs
yjh0502: I agree.

yjh0502:
@koute Thanks for the review. I updated the patch, and also rebased the PR to `master` branch.

koute:
LGTM! Although I can't merge this as there is still `.runtime.js.swp` file added. (:

So please remove that and it'll be good to go!

yjh0502:
@koute I missed that one, sorry :(.

koute:
No problem!

Thanks! (:

Issue #128, opened 2018-02-15
Async unit tests


Pauan:
Since we now have *excellent* Promise and Future support, we should start thinking about how to write async unit tests.

@koute you said you had some ideas about this, what are they?

koute:
Yes, well, it shouldn't be *super* hard to get async tests working today, it'd just involve quite a bit of legwork.

The biggest issue currently with having async tests is that Rust itself doesn't support custom test harnesses, so we'd have to get around that.

Something like this:

1) Write a procedural attribute macro, in a fashion similar to the current `#[js_export]` - call it `#[async_test]` or something like that. For every function marked with it something like this would be generated: (would require #![feature(linkage)])

```rust
#[linkage = "external"]
#[no_mangle]
#[allow(dead_code)]
extern fn test_foobar() {
    // ...
}
```

Then you could just slap an `#[async_test]` anywhere you want in your project just as you'd do with `#[test]`.

2) When compiling tests have `cargo-web` extract all of the async tests, generate a custom test harness for it and then run them.

I don't think we should involve promises and futures into this though, at least on the lowest level; having `fn test_foobar(done: fn())` would be enough (then you could test promises and futures themselves like this, and most of our async APIs don't need full power of promises/futures to test them anyway)

Diggsey:
> I don't think we should involve promises and futures into this though

I disagree on the futures point - I think we should push towards ubiquitous use of futures and streams for async programming within stdweb - I imagine we'll get *some* form of `await` usable on stable this year, and I'd like to be able to take full advantage of that with stdweb.

I want to be able to slap `#[async_test]` directly on an async function and have it behave as expected. I think we should even push for this async test harness to be supported upstream of stdweb, and I think there will be wide support for this amongst all rust users.

The way I see it, stdweb should be the gateway between the javascript/browser/node world and the rust world, providing interfaces and compatibility layers where appropriate, but on the rust side of that boundary we should have "normal" rust code, where the best practices for rust on stdweb are the same as the best practices for any rust project.

Pauan:
> @koute I don't think we should involve promises and futures into this though, at least on the lowest level

I agree that the lowest level should be very simple, but testing Futures will be *very* common, so we need some sugar on top which makes it easy.

Perhaps something like this:

```rust
#[async_callback_test]
fn foo(done: fn()) {
    // ...
}

#[async_future_test]
fn foo() -> impl Future<Item = (), Error = Error> {
    // ...
}
```

[...]

PR #150, opened 2018-03-05
Adding in Promise::from_future function

Pauan:
I did some light testing to make sure that this works correctly, but no unit tests.

### DIFF
--- a/src/webcore/promise.rs
+++ b/src/webcore/promise.rs
@@ -4,7 +4,13 @@ use webcore::value::{Value, Reference};
 use webcore::try_from::{TryInto, TryFrom};
 
 #[cfg(feature = "futures")]
+use webcore::serialization::JsSerialize;
+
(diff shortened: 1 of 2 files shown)


Pauan:
@koute Ping.

koute:
@Pauan Yes, sorry. I'll take a more thorough look at this very soon!

Pauan:
@koute No rush, just making sure you don't forget it.

koute:
Hmmm.... it looks mostly good to me! I'll merge it, although I'd *really* like some tests before stabilizing this. (That, and I guess wait for `futures 0.2`? I see that there is already `0.2.0-alpha` so it might be worthwhile to slowly start working on it.)

Anyway, thanks!

Pauan:
I agree that unit tests would be great, but that will require #128 to be fixed first.

PR #152, opened 2018-03-05
Escape `<textarea>` in doc comment

Badel2:
Otherwise the docs interpret it as html and it messes up the whole page.


Badel2:
This is the affected page: [link](https://docs.rs/stdweb/*/stdweb/web/html_element/struct.TextAreaElement.html)

And this is what it looks like to me: [Imgur](https://i.imgur.com/XlVGiQ0.png)

koute:
Thanks!

PR #154, opened 2018-03-07
Missed exporting FillRule (#153)

### LINKED ISSUE #153: Cannot currently use JS's equivalent to `context.fill();` on canvas 2D context
I am using `stdweb 0.4.1`.
Currently, when trying to draw a circle filled with color, I make use of `context.beginPath` and `context.closePath`, along with `context.fill` to actually color the shape, which works nicely in JS. However, when using `stdweb`, the `context.fill` method seems to depend on a `FillStyle` which is not currently made public by the API.
Though the documentation says that it is not supported because `Path2D` is experimental, I don't think it is related to `Path2D` itself, since it uses the stable call to a context method. Is there a way to expose this method?

The following code is how I am currently managing to draw a circle on screen.
```rust
    /// Draws a colored circle.
    pub fn draw_circle(&self, color: &str, pos: Vector2, radius: f64) {
        self.ctx.begin_path();
        self.ctx.set_fill_style_color(color);
        self.ctx.arc(pos.x, pos.y, radius, 0.0, PI * 2.0, false);
        js!{ @(no_return) @{&self.ctx}.fill(); }; // workaround
        self.ctx.close_path();
    }
```

teovoinea:
Fixes #153 

Didn't export FillRule from ContextRendering2D.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -226,7 +226,7 @@ pub mod web {
     pub use webapi::file_reader::{FileReader, FileReaderResult};
     pub use webapi::history::History;
     pub use webapi::web_socket::{WebSocket, SocketCloseCode, SocketBinaryType};
-    pub use webapi::rendering_context::{RenderingContext, CanvasRenderingContext2d, CanvasGradient, CanvasPattern, CanvasStyle, ImageData, TextMetrics};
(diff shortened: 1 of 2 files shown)


koute:
Thanks!

PR #155, opened 2018-03-08
Added attribute related DOM operations for Element

csharad:


teovoinea:
Don't forget to export `InvalidCharacterError` in lib.rs :)

https://github.com/koute/stdweb/blob/f6de0e085835dfcf7b39e28534fef30901440d5d/src/lib.rs#L236

src/webapi/element.rs
```diff
@@ -22,6 +24,78 @@ pub trait IElement: INode + IParentNode {
             js!( return @{self.as_ref()}.classList; ).into_reference_unchecked().unwrap()
         }
     }
+
+    /// The Element.hasAttribute() method returns a Boolean value indicating whether
+    /// the specified element has the specified attribute or not.
+    ///
+    /// [(JavaScript docs)](https://developer.mozilla.org/en-US/docs/Web/API/Element/hasAttribute)
+    // https://dom.spec.whatwg.org/#ref-for-dom-element-hasattributes
```
koute: Wrong link. (:

src/webapi/element.rs
```diff
@@ -22,6 +24,78 @@ pub trait IElement: INode + IParentNode {
...
+    // https://dom.spec.whatwg.org/#ref-for-dom-element-getattribute
+    fn get_attribute( &self, name: &str ) -> Option< String > {
+        if self.has_attribute( name ) {
+            let value = js!(
+                return @{self.as_ref()}.getAttribute( @{name} );
+            ).try_into().unwrap();
+            Some(value)
+        } else {
+            None
+        }
```
koute: Is this `if` necessary? `try_into` should properly handle convertion to `Option< String >`.

csharad: AFAIK `try_into` would get empty string as an empty string exactly. So I checked beforehand whether it exists.

> If the given attribute does not exist, the value returned will either be null or "" (the empty string)

Pauan: The `try_into` method knows what the expected type is, so it can do the appropriate conversions.

So because the return type is `Option<String>`, if you use `try_into()` it will return `Some("")`, not `""`, therefore the `if` check is not necessary.

Pauan: In fact the `try_into` method is extremely powerful. For example, you can do stuff like this:

```rust
let future: PromiseFuture<Vec<Option<String>>> = js!( return ... ).try_into().unwrap();
```

In this case it expects `...` to be a JavaScript `Promise` that returns a JavaScript `Array`, and each element in the `Array` can be either a JavaScript `String`, `null`, or `undefined`.

In other words, you only need to specify the return type and `try_into()` then magically figures out how to do the appropriate conversions.

xor:
I've been waiting for these features. Are these be expected to be included in the next release? When will the next release be out? If we make a considerable number of attribute changes, will the constant call to "js_try!" affect performance significantly? Would it make sense to create a function that takes in lists of arguments and their values to set them with a single "js_try!" call?

src/webapi/element.rs
csharad: Yeah exactly because "" in JS is going to be converted to Some(""), I did an `if` condition. I wanted to "" be represented as `None`. The reason being, as per MDN, Firefox returns "" on empty attributes whereas Chrome returns `null` for the same. So to bring them both to the same type I wanted both `null` and "" to be converted to `None`.

koute: @csharad Hmm... can you post a jsfiddle which reproduces this?

On both Chromium and Firefox I get the same result here: https://jsfiddle.net/1a68edng/3/

koute:
> Are these expected to be included in the next release? When will the next release be out?

Yes. Probably this week.

> If we make a considerable number of attribute changes, will the constant calls to "js_try!" affect performance significantly? Would it make sense to create a function that takes in lists of arguments and their values to set them with a single "js_try!" call?

Maybe. I can't tell you how big is the overhead exactly. (I'm planning to make some benchmarks soon.) You probably shouldn't worry about it too much unless you can actually tell that it's too slow.

src/webapi/element.rs
Pauan: @csharad It sounds like all modern browsers correctly return `null`, but you're right that MDN says that older browsers return `""` for non-existing attributes.

So I guess we need to check *which* older browsers are wrong, so we can determine whether this workaround is necessary or not.

csharad: I re-read the MDN docs again it said the newer versions of the browsers correctly return `null`. Sorry my bad. But older Gecko returned "". 
I suppose the older browsers won't be able to run wasm too. So removing the `if` condition wouldn't harm.

Pauan: Older browsers can still run asm.js though, so it's still a valid concern.

Personally, I don't mind dropping support for older browsers, but it's not my decision to make.

csharad: @koute So should I drop the `if` condition considering all this. Also I could not find any docs which specify what DOM implementation a browser version has?

koute: @csharad Both Edge and even IE 11 behave properly here, so I think we can remove this workaround. Ancient browsers which would be affected by this won't support a lot of other more critical funcionality anyway, even when compiling to asm.js (e.g. typed arrays).

koute:
LGTM; thanks!

PR #158, opened 2018-03-11
ready_state() should use attribute readyState instead of protocol

DancingBard:
Should hopefully fix a rather nasty RuntimeError


koute:
Thanks!

PR #159, opened 2018-03-14
Add pointer events

srjek:
This pull request implements most of https://w3c.github.io/pointerevents/.

Some issues that popped up:
- `Element::hasPointerCapture` has no MDN page (yet). MDN links are predictable, so I included one anyways. but is it ok to have a dead link in the documentation?
- `PointerEvent` inherits from `MouseEvent`. In events where no buttons were pressed or depressed, a plain `MouseEvent` will leave .button uninitialized as 0, but a `PointerEvent` will set .button to -1. `MouseEvent::button()` will panic if it sees that -1.
    - `MouseEvent::button()` could be changed to recognize -1 as `MouseButton::Left` (aka 0). It feels wrong, but it's what you would currently get if you tried to access .button on the corresponding MouseEvent that fires after a PointerEvent.
    - I have added a `PointerEvent::button_pointer()` which returns a `Option<MouseButton>`. This could be a change to `MouseEvent::button()` instead. Since every MouseEvent.button will or won't be initialized based on the type of event, the concrete events can provide a correct impl for the trait fn. However, the change in return type would be a breaking change.

### DIFF
--- a/src/webapi/events/mouse.rs
+++ b/src/webapi/events/mouse.rs
@@ -195,6 +195,8 @@ pub enum MouseButton {
     Button4,
     /// The fifth mouse button (browser forward).
     Button5,
+    // /// The sixth mouse button, or the Pen Eraser button
+    //TODO: Eraser,
 }
 
 /// Represents the state of mouse buttons in a `MouseEvent`.
--- a/src/webapi/events/pointer.rs
+++ b/src/webapi/events/pointer.rs
@@ -1,0 +1,491 @@
+use webcore::value::Reference;
+use webcore::try_from::TryInto;
+use webapi::event::{IEvent, IUiEvent, UiEvent, Event, ConcreteEvent};
+use webapi::events::mouse::{IMouseEvent, MouseEvent};
+
+#[cfg(feature = "experimental_features_which_may_break_on_minor_version_bumps")]
+use webapi::events::mouse::MouseButton;
+
+/// The `IPointerEvent` interface represents the state of a DOM event produced by a pointer
+/// such as the geometry of the contact point, the device type that generated the event, the
+/// amount of pressure that was applied on the contact surface, etc.
+///
+/// [(JavaScript docs)](https://developer.mozilla.org/en-US/docs/Web/API/PointerEvent)
+// https://w3c.github.io/pointerevents/#pointerevent-interface
+// https://w3c.github.io/pointerevents/extension.html#extensions-to-the-pointerevent-interface
+pub trait IPointerEvent: IMouseEvent {
+    /// Returns a unique identifier for the pointer causing the event.
+    ///
+    /// [(JavaScript docs)](https://developer.mozilla.org/en-US/docs/Web/API/PointerEvent/pointerId)
+    // https://w3c.github.io/pointerevents/#dom-pointerevent-pointerid
+    #[inline]
+    fn pointer_id( &self ) -> i32 {
+        js!(
+            return @{self.as_ref()}.pointerId;
+        ).try_into().unwrap()
+    }
+
+    /// Returns the width, in CSS pixels, of the contact geometry of the pointer.
+    ///
+    /// [(JavaScript docs)](https://developer.mozilla.org/en-US/docs/Web/API/PointerEvent/width)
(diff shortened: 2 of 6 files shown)


src/webapi/events/pointer.rs
```diff
@@ -0,0 +1,483 @@
...
+            return @{self.as_ref()}.height;
+        ).try_into().unwrap()
+    }
+
+    /// Returns the normalized pressure of the pointer in the range [0, 1]
+    ///
+    /// [(JavaScript docs)](https://developer.mozilla.org/en-US/docs/Web/API/PointerEvent/pressure)
+    // https://w3c.github.io/pointerevents/#dom-pointerevent-pressure
+    #[inline]
+    fn pressure( &self ) -> f64 {
```
koute: Hmm... looking at the specs this should return `f32`, I guess, although right now `try_into` doesn't support it. Can you leave a `TODO` so that we can change it later?

src/webapi/events/pointer.rs
```diff
@@ -0,0 +1,483 @@
...
+        ).try_into().unwrap()
+    }
+
+    /// Returns the normalized tangential pressure of the pointer in the range [-1, 1], where 0 is
+    /// the hardware's neutral position
+    ///
+    /// [(JavaScript docs)](https://developer.mozilla.org/en-US/docs/Web/API/PointerEvent/tangentialPressure)
+    // https://w3c.github.io/pointerevents/#dom-pointerevent-tangentialpressure
+    #[inline]
+    fn tangential_pressure( &self ) -> f64 {
```
koute: This too.

src/webapi/events/pointer.rs
```diff
@@ -0,0 +1,483 @@
...
+            return @{self.as_ref()}.isPrimary;
+        ).try_into().unwrap()
+    }
+
+    /// Returns the sequence of all `PointerEvent` instances that were coalesced into the dispatched `PointerMoveEvent`.
+    ///
+    /// [(JavaScript docs)](https://developer.mozilla.org/en-US/docs/Web/API/PointerEvent/getCoalescedEvents)
+    // https://w3c.github.io/pointerevents/extension.html#dom-pointerevent-getcoalescedevents
+    #[inline]
+    fn get_coalesced_events( &self ) -> Vec<PointerEvent> {
```
koute: > Returns the sequence of all `PointerEvent` instances that were coalesced into the dispatched `PointerMoveEvent`.

Shouldn't this be defined in `PointerMoveEvent` then? Even the specs says that it `always returns at least one coalesced event for pointermove events and an empty list for other types of PointerEvents.`

src/webapi/events/pointer.rs
```diff
@@ -0,0 +1,483 @@
...
+        js!(
+            return @{self.as_ref()}.getCoalescedEvents();
+        ).try_into().unwrap()
+    }
+
+    /// Indicates the mouse button that fired this event. A None value indicates no change since the last PointerEvent.
+    ///
+    /// [(JavaScript docs)](https://developer.mozilla.org/en-US/docs/Web/API/MouseEvent/button)
+    // https://w3c.github.io/pointerevents/#the-button-property
+    fn button_pointer( &self ) -> Option<MouseButton> {
```
koute: Hmm... I think that *probably* it would be better to merge this with the `pointer` method, but as you've said that would be a breaking change. I'm fine with merging this as-is for now, however can you attach `#[cfg(feature = "experimental_features_which_may_break_on_minor_version_bumps")]` to this function? (That's how we mark totally ustable functionality; if we're going to remove this then there is no point in pretending that it's stable.)

src/webapi/events/mouse.rs
```diff
@@ -195,6 +196,8 @@ pub enum MouseButton {
     Button4,
     /// The fifth mouse button (browser forward).
     Button5,
+    /// The sixth mouse button, or the Pen Eraser button
+    Eraser,
```
koute: Adding a new variant is also technically a breaking change, so for now could you comment this out and leave a `TODO`?

srjek:
Done

koute:
Thanks!

Issue #160, opened 2018-03-15
MouseOver MouseOut events?


rickycodes:
How does one add these to href `<a>` tags? Am I just blind? Looking for `onmouseover` equivalent

teovoinea:
Looks like `MouseOver` and `MouseOut` haven't been implemented yet. They should exist here: https://github.com/koute/stdweb/blob/master/src/webapi/events/mouse.rs

It would be pretty easy to implement them yourself following the existing mouse events
https://github.com/koute/stdweb/blob/ba37423bdfd8be85c04f96b22bc9397c1264a09d/src/webapi/events/mouse.rs#L299-L314

If you feel like it, you could open a PR implementing the events your need 😊

rickycodes:
I could do that, sure... Just wanted to make sure they weren't already kicking around already.

koute:
Since it was merged in - closing. (:

PR #161, opened 2018-03-15
Add mouseover and mouseout

rickycodes:
re: https://github.com/koute/stdweb/issues/160

Add `MouseOverEvent` and `MouseOutEvent` events


koute:
LGTM! Thanks!

PR #165, opened 2018-03-17
Changing PromiseExecutor to use a Rust VecDeque

Pauan:
The behavior should be identical, but this *dramatically* improves the performance. In my animation test demo it went from ~10 FPS to ~60 FPS!

There are two reasons for this:

1. It maintains a `VecDeque` in Rust, rather than using the browser's event loop queue. This means ***much*** less interop between Rust -> JS -> Rust, which means less overhead.

2. It avoids creating garbage as much as possible.

   The old implementation created 3 functions and 2 Promises per event tick per Task. So if there were 100 Tasks that means it created 300 functions + 200 Promises on every event tick.

   The new implementation creates 1 Promise per event tick. So if there are 100 Tasks that means it creates 1 Promise on every event tick.

   That means *multiple* orders of magnitude less garbage.

It seems that avoiding garbage is much more important for performance than using `VecDeque`, but either way it's a ***huge*** performance benefit.

After these changes, the biggest performance bottlenecks seem to be from the `to_js` and `to_js_string` functions.

### DIFF
--- a/src/webcore/promise_executor.rs
+++ b/src/webcore/promise_executor.rs
@@ -1,10 +1,16 @@
 use futures::future::{Future, ExecuteError, Executor};
 use futures::executor::{self, Notify, Spawn};
 use futures::Async;
+use std::collections::VecDeque;
 use std::result::Result as StdResult;
 use std::cell::{Cell, RefCell};
 use std::rc::Rc;
-use webcore::once::Once;
+use webcore::try_from::TryInto;
+use webcore::value::Reference;
+
+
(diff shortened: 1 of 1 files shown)


Pauan:
Now that I got this done I can work on getting `Discard` landed.

koute:
Thanks for the PR!

Hmm... this is certainly a creative way to implement a promise queue. Since this functionality is unstable anyway I'll merge it in as-is and we can continue fleshing it out. (We really need some tests *and* benchmarks.)

PR #166, opened 2018-03-20
Export all enums used and returned by CanvasRenderingContext2d functions

srjek:
There were some enums inaccessible from outside the crate, but used as arguments and for return types of functions that were exported. This change also fixes the documentation errors that resulted from exporting the enums, adds a variant that was missing, and extends error_enum_boilerplate! to allow the rustdoc meta tags to be passed though.

### DIFF
--- a/src/webapi/rendering_context.rs
+++ b/src/webapi/rendering_context.rs
@@ -60,36 +60,62 @@ pub struct ImageData(Reference);
 #[reference(instance_of = "TextMetrics")]
 pub struct TextMetrics(Reference);
 
-/// The type of compositing operation to apply when drawing new shapes, where type is a string 
-/// identifying which of the twelve compositing operations to use.
-/// 
+/// The type of compositing operation to apply when drawing new shapes
+///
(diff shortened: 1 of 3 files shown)


src/webapi/rendering_context.rs
```diff
@@ -60,36 +60,62 @@ pub struct ImageData(Reference);
...
+    /// New shapes are drawn where there is no existing canvas content. Everything else (including old canvas content) is made transparent.
     SourceOut,
+    /// Draws new shapes only where there is existing canvas content, over existing canvas content
     SourceAtop,
+    /// Draws new shapes behind existing canvas content
     DestinationOver,
+    /// Keeps existing canvas content where it overlaps with the new shape. Everything outside the overlap is made transparent.
     DestinationIn,
+    /// The existing content is kept where it doesn't overlap with the new shape. Everything else, including the new shape area, is made transparent.
+    DestinationOut,
```
koute: Since introducing an enum value is technically a breaking change can you comment this line out and leave a TODO? I'll uncomment it once I'll be making the next semver-incompatible change.

srjek:
Done

koute:
LGTM; thanks!

PR #170, opened 2018-03-22
Add document().body(), document().head(), document().title(), document().set_title()

rickycodes:
Add `document().body()`

eg: `let body: HtmlElement = document().body();`

Not sure if we want a convenience method for this... I guess you could always use: `query_selector("body")`

### DIFF
--- a/src/webapi/document.rs
+++ b/src/webapi/document.rs
@@ -2,6 +2,7 @@ use webcore::value::Reference;
 use webapi::event_target::{IEventTarget, EventTarget};
 use webapi::node::{INode, Node};
 use webapi::element::Element;
+use webapi::html_element::HtmlElement;
 use webapi::text_node::TextNode;
 use webapi::location::Location;
 use webapi::parent_node::IParentNode;
@@ -65,6 +66,18 @@ impl Document {
             js!(
                 return @{self}.location;
             ).into_reference_unchecked()
+        }
+    }
+
+    /// Returns the <body> or <frameset> node of the current document, or null if no such element exists.
+    ///
+    /// [(JavaScript docs)](https://developer.mozilla.org/en-US/docs/Web/API/Document/body)
+    // https://html.spec.whatwg.org/#dom-document-body
+    pub fn body( &self ) -> HtmlElement {
+        unsafe {
+            js!(
+                return @{self}.body;
+            ).into_reference_unchecked().unwrap()
         }
     }
 }
--- a/src/webapi/document.rs
+++ b/src/webapi/document.rs
@@ -1,4 +1,5 @@
 use webcore::value::Reference;
+use webcore::try_from::TryInto;
 use webapi::event_target::{IEventTarget, EventTarget};
 use webapi::node::{INode, Node};
 use webapi::element::Element;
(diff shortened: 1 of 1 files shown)


koute:
If it's listed in the specs as a dedicated attribute I'm fine with having it!

src/webapi/document.rs
```diff
@@ -67,4 +68,16 @@ impl Document {
             ).into_reference_unchecked()
         }
     }
+
+    /// Returns the <body> or <frameset> node of the current document, or null if no such element exists.
+    ///
+    /// [(JavaScript docs)](https://developer.mozilla.org/en-US/docs/Web/API/Document/body)
+    // https://html.spec.whatwg.org/#dom-document-body
```
koute: We try to link to the IDL in the specs, so this link should be: `https://html.spec.whatwg.org/#the-document-object:dom-document-body`

src/webapi/document.rs
```diff
@@ -67,4 +68,16 @@ impl Document {
...
+    pub fn body( &self ) -> HtmlElement {
```
koute: The specs say that the return type of this is `HTMLElement?`, so shouldn't this be `Option< HtmlElement >`?

rickycodes:
@koute updated re: feedback

Pauan:
I think it'd be awesome if you could add in `document().head()` as well.

Pauan:
@rickycodes Thanks! It's not possible to have overloaded functions like that (unless you use two traits).

So it would probably be `document().title()` and `document().set_title("foo")`

teovoinea:
In other modules, we've done `document().get_title()` and `document.set_title("foo")`
eg:
https://github.com/koute/stdweb/blob/426b025ee0727ccd3b32447bbc74601e6d81f8e0/src/webapi/rendering_context.rs#L360-L380

rickycodes:
@teovoinea I've gone ahead and added `get/set_title()`

lmk if the signatures/documentation need any adjustments!

/cc @koute

ilhamp58:
Yess i want

src/webapi/document.rs
```diff
@@ -67,4 +69,54 @@ impl Document {
...
+            js!(
+                return @{self}.head;
+            ).into_reference_unchecked()
+        }
+    }
+
+    /// Gets the title of the document.
+    ///
+    /// [(JavaScript docs)](https://developer.mozilla.org/en-US/docs/Web/API/Document/title)
+    // https://html.spec.whatwg.org/multipage/semantics.html#the-title-element
```
koute: This should link to the IDL: `https://html.spec.whatwg.org/#the-document-object:document.title`

src/webapi/document.rs
```diff
@@ -67,4 +69,54 @@ impl Document {
...
+    pub fn get_title( &self ) -> String {
```
koute: Hmm... the issue of whenever we should use `get_foo` or just simply `foo` for the reader is definitely something to consider.

Corrently we're somewhat inconsistent - e.g. in `ImageElement` and in `WebSocket` we **don't** use the `get_` prefix while in `CanvasRenderingContext2d` we do. I guess it'd be nice to pick one and be consistent from now on.

Personally I prefer to **not** use the `get_` prefix, so if you could get rid of it it'd be great. (I'll make this consistent across the whole crate on the next semver bump.)

teovoinea: We should also update that in the CONTRIBUTING.md then.

rickycodes:
@koute: updated! (:

src/webapi/document.rs
Pauan: wasm-bindgen uses the `foo` and `set_foo` convention, which I personally also like.

src/webapi/document.rs
```diff
@@ -67,4 +69,54 @@ impl Document {
...
+                return @{self}.title;
+            ).try_into().unwrap()
+        }
+    }
+
+    /// Sets the title of the document.
+    ///
+    /// [(JavaScript docs)](https://developer.mozilla.org/en-US/docs/Web/API/Document/title)
+    // https://html.spec.whatwg.org/#the-document-object:document.title
+    pub fn set_title( &self, title: &str ) -> String {
```
koute: Ah, one last thing - we don't generally return the value itself from the setters (usually there's no point), so please get rid of the `-> String` return value.

rickycodes: ah yea, I had wondered about that... I don't really see it as making a huge difference and it could be useful in some cases? I'll remove it, but I am curious what the reasoning for this is?

koute: > I'll remove it, but I am curious what the reasoning for this is?

The reasoning is very simple - there isn't much point in returning the *exactly* the same thing you're passing into the function. (:

```
foo.set_title( "bar" ); // <-- you already know you're passing "bar", so what's the point in returning it?
```

Besides it's always going to be slightly faster to have a function without a return value. (As then we can skip the serialization of the return value.)

koute:
LGTM; thanks!

src/webapi/document.rs
rickycodes: > The reasoning is very simple - there isn't much point in returning the exactly the same thing you're passing into the function. (:

PR #172, opened 2018-03-24
Add NodeList.item()

rickycodes:
https://developer.mozilla.org/en-US/docs/Web/API/NodeList/item

Not sure how useful this is...

Maybe not very?

### DIFF
--- a/src/webapi/node_list.rs
+++ b/src/webapi/node_list.rs
@@ -35,9 +35,9 @@ impl NodeList {
     /// [(JavaScript docs)](https://developer.mozilla.org/en-US/docs/Web/API/NodeList/item)
     // https://dom.spec.whatwg.org/#dom-nodelist-item
     pub fn item( &self, index: u32 ) -> Node {
-      js!(
-        return @{self}[ @{index} ];
-      ).try_into().unwrap()
+        js!(
+            return @{self}[ @{index} ];
+        ).try_into().unwrap()
     }
 
     /// Returns an iterator over the list.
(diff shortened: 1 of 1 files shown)


Pauan:
This needs to return `Option< Node >`

Also it should use 4 spaces for indentation.

rickycodes:
@Pauan updated!

Pauan:
@rickycodes Awesome, thanks!

Diggsey:
Can we use the `Index` trait for this?

Pauan:
@Diggsey `Index` doesn't return `Option`, instead it panics. So we would ideally want both.

Diggsey:
Never mind - due to lack of HKT the `Index` trait was designed to always return a &reference, which is not possible here.

src/webapi/node_list.rs
```diff
@@ -30,6 +30,16 @@ impl NodeList {
         js!( return @{self}.length; ).try_into().unwrap()
     }
 
+    /// Returns a node from a NodeList by index.
+    ///
+    /// [(JavaScript docs)](https://developer.mozilla.org/en-US/docs/Web/API/NodeList/item)
+    // https://dom.spec.whatwg.org/#dom-nodelist-item
```
koute: This should link to the IDL: `https://dom.spec.whatwg.org/#ref-for-dom-nodelist-item`

rickycodes:
@koute: updated! (:

koute:
LGTM; thanks!

PR #174, opened 2018-03-29
Add scroll_top()

rickycodes:
Just floating this for some early feedback, still need to add docs, but thought I'd open a PR...

example usage (which I am doing for my own website: <a href='ricky.codes'>ricky.codes</a>):

```rust
let top = Some(window().page_y_offset()).unwrap_or(document().document_element().unwrap().scroll_top());
```

and then setting:
```rust
document().document_element().unwrap().set_scroll_top(top.parse::<u32>().unwrap());
```

### DIFF
--- a/src/webapi/document.rs
+++ b/src/webapi/document.rs
@@ -116,4 +116,13 @@ impl Document {
             js!( @(no_return) @{self}.title = @{title}; );
         }
     }
+
+    /// documentElement
+    pub fn document_element( &self ) -> Option< HtmlElement > {
+        unsafe {
+            js!(
+                return @{self}.documentElement;
+            ).try_into().unwrap()
+        }
+    }
 }
--- a/src/webapi/document.rs
+++ b/src/webapi/document.rs
@@ -117,7 +117,11 @@ impl Document {
         }
     }
 
-    /// documentElement
+    /// Returns the Element that is the root element of the document (for example, the <html>
+    /// element for HTML documents).
+    ///
+    /// [(JavaScript docs)](https://developer.mozilla.org/en-US/docs/Web/API/Document/documentElement)
+    // https://www.w3.org/TR/DOM-Level-2-Core/core.html#ID-87CD092
     pub fn document_element( &self ) -> Option< HtmlElement > {
         unsafe {
             js!(
--- a/src/webapi/document.rs
+++ b/src/webapi/document.rs
@@ -70,7 +70,7 @@ impl Document {
         }
     }
 
-    /// Returns the <body> or <frameset> node of the current document, or null if no such element exists.
+    /// Returns the `<body>` or `<frameset>` node of the current document, or null if no such element exists.
     ///
     /// [(JavaScript docs)](https://developer.mozilla.org/en-US/docs/Web/API/Document/body)
     // https://html.spec.whatwg.org/#the-document-object:dom-document-body
@@ -82,7 +82,7 @@ impl Document {
         }
     }
 
-    /// Returns the <head> element of the current document. If there are more than one <head>
+    /// Returns the `<head>` element of the current document. If there are more than one `<head>`
     /// elements, the first one is returned.
     ///
     /// [(JavaScript docs)](https://developer.mozilla.org/en-US/docs/Web/API/Document/head)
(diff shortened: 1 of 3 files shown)


src/webapi/element.rs
```diff
@@ -60,6 +60,20 @@ pub trait IElement: INode + IParentNode {
...
 
+    /// scrollTop
+    fn scroll_top( &self ) -> i32 {
+        js!(
+            return @{self.as_ref()}.scrollTop;
+        ).try_into().unwrap()
+    }
+
+    /// scrollTop
+    fn set_scroll_top( &self, value: u32 ) {
```
koute: Hmm... is there any particular reason why the getter returns `i32` and the setter accepts `u32`?

rickycodes: @koute ah, good catch... I guess I should pick one 🤦‍♂️ 

what do you think makes the most sense here?

koute: In general we should stick with the types from the specs, so it'd be the best if you'd find the relevant IDL in the specs and use the type specified there. (:

rickycodes: @koute the IDL states "unrestricted double"

teovoinea: Unrestricted double should be `f64` no?
If you scroll down to `TypedArray objects` table, you can see what the Web IDL types represent.
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/TypedArray

Pauan: @rickycodes You can click on the type to get more information.

The [unrestricted double](https://heycam.github.io/webidl/#idl-unrestricted-double) type is 64-bit floating point, so the type should be `f64`

rickycodes: ah, ok. thanks for clarification! I'll update in the PR

rickycodes: Just ant to do some additional clarification... Here's some example code of mine:

Pauan: @rickycodes [The `double` type](https://heycam.github.io/webidl/#idl-double) should also be `f64`, therefore `page_y_offset` should return `f64`.

The only difference between `double` and `unrestricted double` is that in the WebIDL they guarantee that `double` will not be `NaN`, `-Infinity`, or `Infinity` (but the type is the same).

rickycodes: Ok, cool. I just updated that as well. If things look good I can update with documentation/links and this should be good to go :)

koute: Yep. Still needs the docs and the links to the specs. (:

rickycodes: Ok, this should be all good!

src/webapi/document.rs
```diff
@@ -116,4 +116,17 @@ impl Document {
             js!( @(no_return) @{self}.title = @{title}; );
         }
     }
+
+    /// Returns the Element that is the root element of the document (for example, the <html>
```
koute: The `<html>` has to be escaped in backticks or it will mess up the docs.

src/webapi/document.rs
```diff
@@ -116,4 +116,17 @@ impl Document {
...
+    /// element for HTML documents).
+    ///
+    /// [(JavaScript docs)](https://developer.mozilla.org/en-US/docs/Web/API/Document/documentElement)
+    // https://www.w3.org/TR/DOM-Level-2-Core/core.html#ID-87CD092
```
koute: Please link to this spec instead: https://dom.spec.whatwg.org/#ref-for-dom-document-documentelement

Ditto for other links.

src/webapi/document.rs
```diff
@@ -116,4 +116,17 @@ impl Document {
...
+    pub fn document_element( &self ) -> Option< HtmlElement > {
```
koute: According to specs shouldn't this return `Option< Element >`?

src/webapi/document.rs
rickycodes: this should be good now, I've also gone back and added backticks that were missed in some of my previous other work.

rickycodes: also updated doc links and the change to `document_element()` (:
lmk if you spot anything else!

src/webapi/window.rs
```diff
@@ -190,4 +190,27 @@ impl Window {
             return @{self}.outerHeight;
         ).try_into().unwrap()
     }
+
+    /// he read-only Window property pageYOffset is an alias for scrollY; as such, it returns
```
koute: `he` -> `The` (:

koute:
Besides the typo LGTM. (:

Issue #164, opened 2018-03-17
MouseEvent.{client,screen}{X,Y} are defined as producing double, not long


chris-morgan:
I was puzzled to see this item from the 0.4 changelog:

> - The following now return an `i32` instead of `f64`:
>   - `IMouseEvent::client_x`
>   - `IMouseEvent::client_y`
>   - `IMouseEvent::movement_x`
>   - `IMouseEvent::movement_y`
>   - `IMouseEvent::screen_x`
>   - `IMouseEvent::screen_y`

… because that is heading in the opposite direction to a draft spec change that happened a while ago, from using `long` (i32) to using `double` (f63) for four of these: see https://drafts.csswg.org/cssom-view/#extensions-to-the-mouseevent-interface. https://developer.mozilla.org/en-US/docs/Web/API/MouseEvent also talks about the change, and shows that at least Chrome ≥56 is using double rather than long for these things.

(Note that that is specifically for *client* and *screen* X and Y, not *movement* X and Y, which comes from the pointer lock API and is still defined as `long`. Shrug.)

koute:
A while ago I was going through all of our APIs and aligning them to the current version of the specs, so I didn't know that in the newest draft those are actually defined as `doubles.

Thanks for letting me know! I'll change it back on the next semver-incompatible bump.

PR #177, opened 2018-03-30
Added offset_x and offset_y to IMouseEvent

Tacklebox:


src/webapi/events/mouse.rs
```diff
@@ -72,6 +72,28 @@ pub trait IMouseEvent: IUiEvent {
         ).try_into().unwrap()
     }
 
+    /// Returns the X position on the target element where this event occured.
+    ///
+    /// [(JavaScript docs)](https://developer.mozilla.org/en-US/docs/Web/API/MouseEvent/offsetX)
+    // Not in w3c standard
```
koute: Please add links to the specs. (: 

While it's not in the official standard (yet?) there is a draft spec, e.g. here https://drafts.csswg.org/cssom-view/#ref-for-dom-mouseevent-offsetx

src/webapi/events/mouse.rs
```diff
@@ -72,6 +72,28 @@ pub trait IMouseEvent: IUiEvent {
...
+    #[inline]
+    fn offset_x( &self ) -> i32 {
```
koute: This should return `f64` according to the specs' IDL.

Tacklebox:
I added links to the w3c spec as requested.
About the return type,
https://developer.mozilla.org/en-US/docs/Web/API/MouseEvent/offsetX#Browser_compatibility
shows only chrome and android webview use doubles and the other browsers use longs.
Combined with https://github.com/koute/stdweb/issues/164 I thought it might be better to wait till the next semver to sort that out. Do you still want me to change it to a double?

Pauan:
@Tacklebox Browser compatibility isn't a problem, because `f64` (`double`) is a strict super-set of `i32` (`long`). In other words, it's possible to losslessly convert from `i32` to `f64` (but not vice versa).

And in fact stdweb should automatically do that conversion (since all numbers in JavaScript are `f64` anyways).

Given #164, I personally vote to make them `f64`.

koute:
Yep. Basically what @Pauan said.

LGTM; thanks for the PR!

PR #178, opened 2018-04-01
Implement fallback for references which cannot be stored in a WeakMap

### LINKED ISSUE #92: Some WebGL objects cannot put in the weakmap in firefox

Diggsey:
Fixes #92 

This implements a fallback mechanism using a normal `Map` if attempting to store an object in the `WeakMap` throws an exception - objects stored here are explicitly removed when their reference count reaches zero, meaning that there are no leaks, but that the "stable ID" guarantee does not hold for these specific objects.

This likely comes with a performance penalty due to the use of exception handling.

Personally I would prefer the "stable ID" guarantee to be dropped entirely from `webcore`, and instead a `WeakMap` type was exposed so that stable IDs could still be achieved by opting in for specific objects.

### DIFF
--- a/src/webcore/discard.rs
+++ b/src/webcore/discard.rs
@@ -99,6 +99,7 @@ mod tests {
     }
 
     #[test]
+    #[allow(unused_must_use)]
     fn unused_discard_on_drop() {
         DiscardOnDrop::new( Foo::new() );
     }
--- a/src/webcore/runtime.js
+++ b/src/webcore/runtime.js
@@ -290,6 +290,8 @@ Module.STDWEB_PRIVATE.to_js_string = function to_js_string( index, length ) {
 Module.STDWEB_PRIVATE.id_to_ref_map = {};
 Module.STDWEB_PRIVATE.id_to_refcount_map = {};
 Module.STDWEB_PRIVATE.ref_to_id_map = new WeakMap();
+// Not all types can be stored in a WeakMap
+Module.STDWEB_PRIVATE.ref_to_id_map_fallback = new Map();
 Module.STDWEB_PRIVATE.last_refid = 1;
 
 Module.STDWEB_PRIVATE.id_to_raw_value_map = {};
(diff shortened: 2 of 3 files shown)


koute:
Do you know which objects can't be put in a `WeakMap`? (Besides WebGL-related ones in Firefox, which as far as I understand it is a bug that will be fixed?)

Also it would be nice to know how big exactly is the performance penalty (e.g. compared to crossing the JS FFI boundary).

koute:
Hmm... as I personally wasn't bitten by `WeakMap` rejecting any of my objects and the fact that you could *technically* always wrap such an object in an `{}` I'm kinda ambivalent on this, however I understand that in some cases that might be a problem so I'm willing to accept this, but since this touches a core functionality I'd really like to see at least one test.

Is there any object across all JS implementation which isn't `WeakMap`-insertable? If there is then could we use it to write a test to verify this behavior?

Diggsey:
> Is there any object across all JS implementation which isn't WeakMap-insertable? If there is then could we use it to write a test to verify this behavior?

No, there isn't - AFAIK chrome accepts anything. I've personally tested my WebGL example in firefox under this change, but I don't know of a way to set up an automated test independent of browser implementation. (Yet another reason I would prefer to avoid `WeakMap` entirely)

Diggsey:
Hmm, I could monkey-patch the `WeakMap.set` method to make it deterministically throw an exception if certain objects are added and use that in the test?

koute:
Hmm... sounds good to me!

Diggsey:
@koute Added a test

kuviman:
Since stable IDs are no longer guaranteed for such objects, maybe it is better to drop the guarantee?

It would remove the overhead of catching the exception like so, but will add overhead when checking references for equality though.

I don't know which is more important, but I think that webgl apps are those that usually require more optimization, and they shouldn't need equality checks.

Anyway, having stable id guarantee that is not actually guaranteed seems wrong.

Diggsey:
@kuviman stable IDs are not needed for efficient/correct comparison of references, they're not even really exposed by stdweb.

Diggsey:
To clarify, the "stable ID" guarantee itself is almost entirely useless except for a few niche cases, and I've never seen any code actually relying on it.

kuviman:
I mean, `impl PartialEq for Reference` would need to interop with JS, which would be slower, [here](https://github.com/koute/stdweb/blob/master/src/webcore/value.rs#L64)

Diggsey:
@kuviman No, it wouldn't, I think you misunderstand what the "stable ID" guarantee means: it means that if the reference is *dropped* on the rust side, and then passed back in a second time, it will receive the same ID number as it had originally.

kuviman:
Oh, right, my bad, there is no way to have two references on rust side with different IDs.

Then I don't see why it may be needed at all.

koute:
@Diggsey Just for the record - you *did* verify that the test you've added fails without your changes, right?

As far as stable IDs go - I originally misunderstood what you meat by it, but, yes, I also don't think that keeping the ID the same **after** every reference to a given value is dropped on the Rust side is very useful. (Not that we ever really guaranteed that anywhere.) I'll consider switching to a plain `Map` for `0.5`.

Diggsey:
@koute yes, I ran it without my changes and the test failed.

Diggsey:
> (Not that we ever really guaranteed that anywhere.)

There is an explicit test for it, so I assumed it was intentional.

koute:
> There is an explicit test for it, so I assumed it was intentional.

Well, the tests are not only there for behavior we want to *officially* guarantee, but also to make sure that we won't inadvertently change.

koute:
Anyhow, LGTM. Thanks!

Issue #162, opened 2018-03-15
Implement stricter type-checking for event types


koute:
Currently the `#[reference(instance_of = "MouseEvent")]` line automatically generates an implementation of the [InstanceOf trait](https://docs.rs/stdweb/*/stdweb/trait.InstanceOf.html) for a given type. `stdweb` uses that trait to decide whenever a given `Reference` is convertible to that type.

The default `InstanceOf` implementation does `instanceof TypeName`, which works pretty well for most types, however many events have the same type - e.g. a lot of mouse-related events are all `MouseEvent`s, which means that it's possible to cast between them willy-nilly, even in cases where such a cast wouldn't make any sense.

So it would be nice to replace that with a manual implementation of `InstanceOf` which would check not only the type with `instanceof`, but would also check that its `type` field is properly set.

koute:
Fixed on `master`.

PR #181, opened 2018-04-02
Added reference attribute constraint parameter to address #162

dnorman:
Re #162

I have amended stdweb-derive such that constraints may be added to the instance_of attribute thus avoiding a bunch of boilerplate. Opted to make this a separate meta item rather than creating additional complexity via nesting.

Go easy on me, I'm pretty new to stdweb ;)

### DIFF
--- a/src/webapi/events/socket.rs
+++ b/src/webapi/events/socket.rs
@@ -12,7 +12,7 @@ use webapi::event::{IEvent, Event, ConcreteEvent};
 /// [(JavaScript docs)](https://developer.mozilla.org/en-US/docs/Web/Events/close)
 // https://html.spec.whatwg.org/multipage/web-sockets.html#closeevent
 #[derive(Clone, Debug, PartialEq, Eq, ReferenceType)]
-#[reference(instance_of = "CloseEvent")] // TODO: Better type check.
+#[reference(instance_of = "CloseEvent", constraint="type=close")]
 #[reference(subclass_of(Event))]
 pub struct SocketCloseEvent( Reference );
 
@@ -63,7 +63,7 @@ impl ConcreteEvent for SocketCloseEvent {
 // https://html.spec.whatwg.org/#event-error
 // https://html.spec.whatwg.org/multipage/web-sockets.html#handler-websocket-onerror
 #[derive(Clone, Debug, PartialEq, Eq, ReferenceType)]
-#[reference(instance_of = "Event")] // TODO: Better type check.
+#[reference(instance_of = "Event", constraint="type=error")]
 #[reference(subclass_of(Event))]
 pub struct SocketErrorEvent( Reference );
 
@@ -78,7 +78,7 @@ impl ConcreteEvent for SocketErrorEvent {
 // https://html.spec.whatwg.org/#event-open
 // https://html.spec.whatwg.org/multipage/web-sockets.html#handler-websocket-onopen
 #[derive(Clone, Debug, PartialEq, Eq, ReferenceType)]
-#[reference(instance_of = "Event")] // TODO: Better type check.
+#[reference(instance_of = "Event", constraint="type=open")]
 #[reference(subclass_of(Event))]
 pub struct SocketOpenEvent( Reference );
 
--- a/stdweb-derive/src/lib.rs
+++ b/stdweb-derive/src/lib.rs
@@ -54,6 +54,7 @@ pub fn derive_reference_type( input: TokenStream ) -> TokenStream {
 
     let mut instance_of = None;
     let mut subclass_of = Vec::new();
+    let mut constraints = Vec::new();
 
     for meta_items in input.attrs.iter().filter_map( get_meta_items ) {
         for meta in meta_items {
@@ -67,6 +68,18 @@ pub fn derive_reference_type( input: TokenStream ) -> TokenStream {
                         instance_of = Some( str.value() );
                     } else {
                         panic!( "The value of '#[reference(instance_of = ...)]' is not a string!" );
+                    }
+                },
+                syn::NestedMeta::Meta( syn::Meta::NameValue( ref meta ) ) if meta.ident == "constraint" => {
+                    if let syn::Lit::Str( ref str ) = meta.lit {
+                        let val = str.value();
+                        let parts : Vec<&str> = val.splitn(2,"=").collect();
+                        if parts.len() != 2 {
+                            panic!("The value of '#[reference(..., constraint = ...)]' must be in the form of \"js_field=Ident!\"" );
+                        }
+                        constraints.push(( parts[0].to_owned(), parts[1].to_owned()) );
+                    } else {
+                        panic!( "The value of '#[reference(..., constraint = ...)]' is not a string!" );
                     }
                 },
                 syn::NestedMeta::Meta( syn::Meta::List( ref meta ) ) if meta.ident == "subclass_of" => {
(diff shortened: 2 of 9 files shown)


stdweb-derive/src/lib.rs
```diff
@@ -69,6 +70,18 @@ pub fn derive_reference_type( input: TokenStream ) -> TokenStream {
                         panic!( "The value of '#[reference(instance_of = ...)]' is not a string!" );
                     }
                 },
+                syn::NestedMeta::Meta( syn::Meta::NameValue( ref meta ) ) if meta.ident == "constraint" => {
+                    if let syn::Lit::Str( ref str ) = meta.lit {
+                        let val = str.value();
+                        let parts : Vec<&str> = val.splitn(2,"=").collect();
+                        if parts.len() != 2 {
```
koute: AFAIR won't `splitn(2, ...)` always return two elements? If so then this will never trigger.

stdweb-derive/src/lib.rs
```diff
@@ -69,6 +70,18 @@ pub fn derive_reference_type( input: TokenStream ) -> TokenStream {
...
+                            panic!("The value of '#[reference(..., constraint = ...)]' must be in the form of \"js_field=value\"" );
+                        }
+                        constraints.push(( parts[0].to_owned(), parts[1].to_owned()) );
```
koute: For now I think it'd be better to be more conservative and make it accept only one constraint, and error out in case of a duplicate. (So `Option` instead of a `Vec`.)

koute: Could you also check that the field is a valid JS identifier? (It'd be fine to make sure that it starts with a letter and the rest contains only `[a-zA-Z0-9_]`.)

src/webapi/events/socket.rs
```diff
@@ -78,7 +78,7 @@ impl ConcreteEvent for SocketErrorEvent {
 // https://html.spec.whatwg.org/#event-open
 // https://html.spec.whatwg.org/multipage/web-sockets.html#handler-websocket-onopen
 #[derive(Clone, Debug, PartialEq, Eq, ReferenceType)]
-#[reference(instance_of = "Event")] // TODO: Better type check.
+#[reference(instance_of = "Event", constraint="type=open")]
```
koute: Hmm... now that I think about it I guess I'd really like if whatever is in constraint would be an actual JS expression (right now this supports only strings, and is not forward compatible with adding e.g. integers or other types), which means that in case of a string it would have to contain `"`, which means that we *probably* would want a different outer delimiter than `"`? `constraint=(type = "open")` maybe? (As long as it's reasonable anything should be fine with me.)

Or maybe make it standalone, e.g. `#[reference(constraint(type = "open"))]`? That would probably work better if we'd want to support multiple constraints (since the user would then just simply duplicate the `#[reference(constraint(...))]` line)

stdweb-derive/src/lib.rs
```diff
@@ -165,14 +178,25 @@ pub fn derive_reference_type( input: TokenStream ) -> TokenStream {
...
+            let mut js_constraints : Vec<String> = constraints.into_iter().map(|constraint|{
+                let js_field = constraint.0.replace("\"", "\\\"");
+                let js_value = constraint.1.replace("\"", "\\\"");
+                format!("_ref[\"{}\"] === \"{}\"", &js_field, &js_value)
+            }).collect();
+
+            js_constraints.insert( 0, format!("_ref instanceof {}", &js_name) );
+
+            let js_constraints = js_constraints.join(" && ");
+
```
koute: Could we have semi-separate codepaths for this? (That is - replace the whole body of `instance_of` with `#instance_of_body`, have the old case of having only `instance_of` generate code exactly as it used to, and only generate a differente `instance_of_body` in case of a constraint.)

stdweb-derive/src/lib.rs
dnorman: Either way is fine by me. Will make the change

src/webapi/events/socket.rs
dnorman: @koute I'm not sure I entirely understand what you're suggesting. 

I weighed the pros and cons of making it a JS expression, but constraint="type='open' "  or constraint(type = " 'open' ") seemed like overkill, given the dearth of non-string scenarios. Also, I'm uncertain if constraint(type=123) is discernible from constraint(type="123"). As far as I can tell, there should be no breaking changes if we decide to change it later to address future scenarios.

Additionally, I was unable to immediately conceive of any directives other than instance_of for which the constraint attribute would be relevant.

With the above having been said, I'm happy to make whatever changes you feel are appropriate if you can help me understand which specific syntax you would prefer :)

dnorman:
Will apply commensurate changes once I have clarity on the above question :)

src/webapi/events/socket.rs
koute: My point is not to support arbitrary JS expressions right now (we can ignore other types and support only strings for now), but to have a forward compatible syntax. (:

> Also, I'm uncertain if constraint(type=123) is discernible from constraint(type="123"). As far as I can tell, there should be no breaking changes if we decide to change it later to address future scenarios.

*Currently* as this PR stands `type=123` means that `type` field needs to be equal to `"123"` (that is - to a string). So, say, we'd want to extend this with support for integers - how will we be able to tell if `type=123` means `123` (an integer) or `"123"` (a string)?

Basically, have the most forward compatible syntax possible, but be as conservative as possible with it.

> Additionally, I was unable to immediately conceive of any directives other than instance_of for which the constraint attribute would be relevant.

How about having a constraint *without* an `instance_of`? (: I mean, we *probably* won't use it internally, but this is also usable by our downstream users, and they might find a use for it. (That is - make the object convertible based on a value of a field *without* discriminating based on what it's an instance of.)

And, I think that purely *syntax-wise* is should be cleaner to have the constraint in its own separate thing than shoehorned into the `instance_of`. Especially in case of multiple constraints, e.g.

```
#[reference(instance_of = "Event", constraint=..., constraint=..., constraint=...)]
```

vs

```
#[reference(constraint=...)]
#[reference(constraint=...)]
#[reference(constraint=...)]
```

It might not be the same for everyone, but for me *personally* the second one is more obvious in that it's actually *adding* constraints instead of overriding them.

That said, of course, I might be wrong, and if you disagree with some of my reasoning then I'm more than happy to discuss it.

stdweb-derive/src/lib.rs
dnorman: The JS valid identifier check seems reasonable. As to the question of Option vs Vec, I think the prior comment about reference(constraint...) being separate from reference(instance_of...) is germane. If they are to be in a single line, then it should be a vec, else an Option is fine. ( I do mildly prefer the single line approach for reasons stated in my prior comment )

src/webapi/events/socket.rs
koute: > If they were to put reference(contraint...) after the reference(instance_of) then they would get an undesirable outcome, insofar as the constraints would be silently ignored.

I don't understand - how so? The constraints would never be silently ignored, and it shouldn't matter whenever the user specifies the `constraint` or `instance_of` first - it should work exactly the same. *If* the `instance_of` is specified then that is always checked first, then every `constraint`s are checked in the order of their appearance. Nothing overrides anything - those are declarative attributes, not imperative commands. (This would apply equally in both the single-line case and the multiple-line case.)

> Additionally, I think it may make sense to change constraint to constraint_string and also add constraint_js for raw js. I don't anticipate we would need integer comparisons, but conceivably one could add constraint_int as well. Does that make sense?

But, you see, why not simply make it *look* like it's a string in the first place? (: I would rather avoid having multiple `constraint_foo`, `constraint_bar` keys, and I think it just makes more sense to have a single one and just figure out the type from the value. `"123"` is a string, `123` is a number, etc.

I do agree that something like `constraint="foo=\"foo\""` or `constraint="foo='foo'"` looks super awkward; perhaps this is why you don't like the idea of putting a quoted value in there? In that case as I've previously suggested we can just use a different syntax and have the leading `="` and trailing `"` replaced by `(` and `)` respectively.

koute:
`master` has now a very similar mechanism to this specific to events, so I'm closing this PR. If anyone in the future wants to resurrect it in the future be my guest.

PR #183, opened 2018-04-03
Add tokenlist toggle

rickycodes:
seems like this should be here

### DIFF
--- a/src/webapi/token_list.rs
+++ b/src/webapi/token_list.rs
@@ -50,4 +50,13 @@ impl TokenList {
     pub fn contains( &self, token: &str ) -> bool {
         js!( return @{self}.contains( @{token} ); ).try_into().unwrap()
     }
+
+    /// The toggle() method of the DOMTokenList interface removes a given token from the
+    /// list and returns false. If token doesn't exist it's added and the function returns true.
+    ///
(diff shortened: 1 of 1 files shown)


src/webapi/token_list.rs
```diff
@@ -50,4 +50,13 @@ impl TokenList {
     pub fn contains( &self, token: &str ) -> bool {
         js!( return @{self}.contains( @{token} ); ).try_into().unwrap()
     }
+
+    /// The toggle() method of the DOMTokenList interface removes a given token from the
+    /// list and returns false. If token doesn't exist it's added and the function returns true.
+    ///
+    /// [(JavaScript docs)](https://developer.mozilla.org/en-US/docs/Web/API/DOMTokenList/toggle)
+    // https://dom.spec.whatwg.org/#ref-for-dom-domtokenlist-toggle
+    pub fn toggle( &self, token: &str ) -> bool {
```
koute: 1. This should return a `Result` (see [here](https://dom.spec.whatwg.org/#dom-domtokenlist-toggle) for what it can throw; see other cases where we've used `js_try!` on how to make an error type for it).
2. We should add the optional argument while we're at it (`force: bool`)

PR #192, opened 2018-04-06
fix #190: add missing methods to Location and use SecurityError

### LINKED ISSUE #190: Add pathname to Location
Here is the full `window.location` object in javascript 
```
>> print(JSON.stringify(window.location, null, 4))
"{
    \"href\": \"http://krasimirtsonev.com/blog/article/A-modern-JavaScript-router-in-100-lines-history-api-pushState-hash-url\",
    \"origin\": \"http://krasimirtsonev.com\",
    \"protocol\": \"http:\",
    \"host\": \"krasimirtsonev.com\",
    \"hostname\": \"krasimirtsonev.com\",
    \"port\": \"\",
    \"pathname\": \"/blog/article/A-modern-JavaScript-router-in-100-lines-history-api-pushState-hash-url\",
    \"search\": \"\",
    \"hash\": \"\"
}"
```

Currently `Location` only includes `hash` and `href`. It seems to me like it should be zero-cost to include at _least_ pathname as well. Also, I think including _all_ of them would be at least useful.

Thoughts are welcome. I suppose an argument could be made that the user can parse the `href` directly if they want more information.

Thanks.

vitiral:


vitiral:
this fixes #192

src/webapi/location.rs
```diff
@@ -1,38 +1,120 @@
...
 pub struct Location( Reference );
 
+macro_rules! jsattr {
+    [$s:ident, $attr:ident] => {{
+        js_try!(
+            return @{$s}.$attr;
+        ).unwrap()
+    }}
+}
+
```
koute: Hmm... I'm not sure if having such a macro is worth it? You might as well just paste `js_try!( return @{self}.attr; ).unwrap()` everywhere and it's not going to be much different.

src/webapi/location.rs
```diff
@@ -1,38 +1,120 @@
...
 impl Location {
     /// The entire URL.
     ///
     /// [(JavaScript docs)](https://developer.mozilla.org/en-US/docs/Web/API/Location/href)
     // https://html.spec.whatwg.org/#dom-location-href
```
koute: Can you change all of the specs' links to point to the IDL in the specs? (Just as the `hash` has.)

src/webapi/location.rs
```diff
@@ -1,38 +1,120 @@
...
 /// The `Location` interface represents the location (URL) of the object it
 /// is linked to. Changes done on it are reflected on the object it relates
 /// to. Both the [Document](struct.Document.html) and [Window](struct.Window.html)
 /// interface have such a linked `Location`, accessible via [Document::location](struct.Document.html#method.location)
 /// and [Window::location](struct.Window.html#method.location) respectively.
 ///
+/// Note that all `Location` methods can return a `SecurityError` if the `Location` object's
+/// relevant `Document`'s origin is not same origin-domain with the entry settings object's origin.
+/// See: https://html.spec.whatwg.org/#dom-location-href
+///
```
koute: Hmm... I wonder if it would be possible to write a test (which would run locally, totally offline and without setting up dummy servers to fetch something cross-origin) to check that one of the methods actually throws a security error?

These are simple enough that I'm fine with not having a test, but *if* you could think of a way to to write such a test I'd certainly be very nice. (:

vitiral: I have absolutely no clue... haha.

It would be nice to be able to spin up such tests easily without something like selenium. With selenium it would probably be possible though...

src/webapi/location.rs
```diff
@@ -18,21 +22,91 @@ impl Location {
...
+    ///
+    /// - For URL using the http or https, the scheme followed by `'://'`, followed by the domain,
+    ///   followed by `':'`, followed by the port (the default port, 80 and 443 respectively, if
+    ///   explicitely specified);
+    /// - For URL using `file: scheme`, the value is browser dependant.
+    /// - For URL using the blob: scheme, the origin of the URL following blob:. E.g
+    ///   "blob:https://mozilla.org" will have "https://mozilla.org".
+    ///
+    /// [(JavaScript docs)](https://developer.mozilla.org/en-US/docs/Web/API/Location/origin)
+    // https://html.spec.whatwg.org/#dom-location-origin
```
koute: This and some of the other links still do not point to the IDL. (:

vitiral:
@koute should be G2G now!

koute:
LGTM; thanks!

PR #195, opened 2018-04-08
fix #191: clarify PopStateEvent documentation

### LINKED ISSUE #191: event::PopStateEvent erroniously states that it can be created by push_state or replace_state
https://docs.rs/stdweb/*/stdweb/web/event/struct.PopStateEvent.html#method.state

>  If the history entry being activated was created by a call to `history.push_state()` or was affected by a call to `history.replace_state()`, the popstate event's state property contains a copy of the history entry's state object.

However, the `PopStateEvent` (`window.onpopstate`) will in fact never be triggered by `history.{push, replace}State()`, which is what [stdweb calls](https://docs.rs/stdweb/*/src/stdweb/webapi/history.rs.html#36).

https://developer.mozilla.org/en-US/docs/Web/API/WindowEventHandlers/onpopstate

> Note: calling history.pushState() or history.replaceState() won't trigger a popstate event. The popstate event is only triggered by performing a browser action, such as clicking on the back button (or calling history.back() in JavaScript), when navigating between two history entries for the same document.

vitiral:
Fix for #191

One thing I noticed is a difference within mozilla's documentation.

https://developer.mozilla.org/en-US/docs/Web/API/WindowEventHandlers/onpopstate

> Browsers tend to handle the popstate event differently on page load. Chrome (prior to v34) and Safari (prior to 10.0) have always emitted a popstate event on page load, but Firefox didn't for a long time; it does in modern versions.

https://developer.mozilla.org/en-US/docs/Web/API/PopStateEvent

> Browsers tend to handle the popstate event differently on page load. Chrome and Safari always emit a popstate event on page load, but Firefox doesn't.

I have found the second one to be more accurate (I am using firefox 58).

### DIFF
--- a/src/webapi/events/history.rs
+++ b/src/webapi/events/history.rs
@@ -43,10 +43,19 @@ impl HashChangeEvent {
     }
 }
 
-/// A popstate event is dispatched to the window every time the active history entry changes
+/// A `PopStateEvent` is dispatched to the window every time the active history entry changes
 /// between two history entries for the same document. If the history entry being activated was
-/// created by a call to history.push_state() or was affected by a call to history.replace_state(),
-/// the popstate event's state property contains a copy of the history entry's state object.
+/// created by a call to `history.push_state()` or was affected by a call to
+/// `history.replace_state()`, the `PopStateEvent`'s state property contains a copy of the history
+/// entry's state object.
+///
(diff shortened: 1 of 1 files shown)


koute:
I haven't checked the accuracy of the changes myself, but I trust they're correct. LGTM; thanks!

PR #200, opened 2018-04-11
fix #199: add Node::from_html

### LINKED ISSUE #199: Add Node::from_html
https://docs.rs/stdweb/0.4.2/stdweb/web/struct.Node.html

I think a worthwhile addition would be the ability to get a `Node` from raw html

Something like this code works for me:
```
pub fn from_html(html: &str) -> Result<Node, ConversionError> {
        Node::try_from(js! {
            var div = document.createElement("div");
            div.innerHTML = @{html};
            return div;
        })
}
```

vitiral:
fix #199 

Outstanding question:
- Should the "root element" be ~~`div`~~ `span`? 
- Should it be configurable in some way (`id`, let the user specify the `tag`, let the user insert a node?)
- should this be implemented as `Node::set_inner_html` instead (or can we have both)?

### DIFF
--- a/src/webapi/node.rs
+++ b/src/webapi/node.rs
@@ -363,19 +363,22 @@ impl IEventTarget for Node {}
 impl INode for Node {}
 
 impl Node {
-    /// Attempt to create the `Node` from raw html.
+    /// Attempt to create the `Node` from raw html. The html string must contain **exactly one**
+    /// root node.
     ///
-    /// # Panics
+    /// Returns a `SyntaxError` if:
     ///
-    /// This method panics if the resulting number of nodes != 1. To ensure this
-    /// it is recommended to at _least_ have programatic control of your root
-    /// HTML node.
+    /// - There is not **exactly one** root node.
+    /// - The html syntax is wrong. However, on most browsers the html parsing algorighm is
+    ///   _unbelievably_ forgiving and will just turn your html into text or maybe even an empty
+    ///   string.
     ///
-    /// # Examples
+    /// It is recommended to have control over the html being given to this function as not
+    /// having control is a security concern.
     ///
-    /// ```
-    /// let node = Node::from_html("<div>Some text, horray!</div>").unwrap();
-    /// ```
+    /// For more details, see information about setting `innerHTML`:
+    ///
+    /// <https://developer.mozilla.org/en-US/docs/Web/API/Element/innerHTML>
     pub fn from_html(html: &str) -> Result<Node, SyntaxError> {
         js_try!(
             var span = document.createElement("span");
--- a/src/webapi/node.rs
+++ b/src/webapi/node.rs
@@ -363,11 +363,14 @@ impl IEventTarget for Node {}
 impl INode for Node {}
 
 impl Node {
+    /// Attempt to create the `Node` from raw html.
+    ///
+    /// Note that this uses a `span` node as the parent.
     pub fn from_html(html: &str) -> Result<Node, ConversionError> {
         Node::try_from(js! {
-            var div = document.createElement("div");
-            div.innerHTML = @{html};
-            return div;
+            var span = document.createElement("span");
+            span.innerHTML = @{html};
+            return span;
         })
     }
 }
@@ -826,7 +829,7 @@ mod tests {
         let inner = node.first_child().unwrap();
         let text = inner.first_child().unwrap();
 
-        assert_eq!(node.node_name(), "DIV");
+        assert_eq!(node.node_name(), "SPAN");
         assert_eq!(node.last_child().unwrap(), inner);
 
         assert_eq!(inner.node_name(), "DIV");
--- a/src/webapi/node.rs
+++ b/src/webapi/node.rs
@@ -377,40 +377,30 @@ impl Node {
     /// let node = Node::from_html("<div>Some text, horray!</div>").unwrap();
     /// ```
     pub fn from_html(html: &str) -> Result<Node, SyntaxError> {
-        let result = js! {
-            var span = document.createElement("span");
-            try {
-                span.innerHTML = @{html};
-            } catch( error ) {
-                return {
-                    error: error,
-                    success: false,
-                };
-            }
+        let result: Result<Value, Value> = js_try!(
+            var xmldoc = document.implementation.createDocument(null, null);
+            var span = xmldoc.createElement("span");
+            span.innerHTML = @{html};
 
             if( span.childNodes.length != 1 ) {
-                return {
-                    panic: "HTML had ${span.childNodes.length} nodes but must have 1",
-                    success: false,
-                };
+                throw SyntaxError("HTML had ${span.childNodes.length} nodes but must have 1");
             }
             return span.childNodes[0];
-        };
+            return "foo";
+        ).unwrap();
 
-
-        if let Value::Bool(false) = js!{ return @{&result}.success; } {
-            if let Value::String(e) = js! { return @{&result}.panic; } {
-                panic!(e);
-            };
-
-            let err = SyntaxError::try_from(result)
-                .expect("was error but not syntax error");
-            return Err(err);
+        match result {
+            Ok(v) => {
+                let node = Node::try_from(v)
+                    .expect("inner type was not a node.");
+                Ok(node)
+            }
+            Err(e) =>  {
+                let err = SyntaxError::try_from(e)
+                    .expect("was error but not syntax error");
+                Err(err)
+            }
         }
-
-        let node = Node::try_from(result)
-            .expect("inner type was not a node.");
-        Ok(node)
     }
 }
 
@@ -873,11 +863,8 @@ mod tests {
         assert_eq!(text.node_name(), "#text");
         assert_eq!(text.node_value().unwrap(), "Some text, horray!");
         assert!(text.first_child().is_none());
-    }
 
-    #[test]
-    #[should_panic]
-    fn from_html_panic() {
-        let _ = Node::from_html("<div>foo</div><div>bar</div>");
+        assert!(Node::from_html("<div>foo</div><div>bar</div>").is_err());
+        assert!(Node::from_html("<di").is_err());
     }
 }
--- a/src/webapi/node.rs
+++ b/src/webapi/node.rs
@@ -1,6 +1,6 @@
 use std::mem;
 
-use webcore::value::{Reference, Value};
+use webcore::value::Reference;
 use webcore::try_from::{TryFrom, TryInto};
 use webapi::document::Document;
 use webapi::dom_exception::{HierarchyRequestError, NotFoundError, SyntaxError};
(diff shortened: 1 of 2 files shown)


vitiral:
`set_inner_html` requires you to create a node first which is kind of annoying IMO. I think there's room for both :smile:

I could just return `span.child_nodes()`. That would make a lot of sense.

vitiral:
of course, the caller could just call `child_nodes()` pretty easily as well

src/webapi/node.rs
```diff
@@ -362,6 +362,19 @@ pub struct Node( Reference );
...
+    /// Attempt to create the `Node` from raw html.
+    ///
+    /// Note that this uses a `span` node as the parent.
+    pub fn from_html(html: &str) -> Result<Node, ConversionError> {
+        Node::try_from(js! {
+            var span = document.createElement("span");
+            span.innerHTML = @{html};
+            return span;
+        })
+    }
```
koute: 1. Automatically creating a `span` here seems kind-of arbitrary to me and potentially confusing, e.g. if I do:

```
Node::from_html( "<div>Hello!</div>" )
```

I'd expect it to return a `<div>`; the fact that it returns a `<span>` is non-obvious from the method signature itself, so *if* we're going to add a method like this then I think it should require that it contains only a single root node and return that it.

2. `innerHTML` this can throw a `SyntaxError` if the `document` is XML (AFAIK this happens when we have an XHTML page), e.g. similar to this:

```
var xmldoc = document.implementation.createDocument(null, null);
var span = xmldoc.createElement("span"); 
span.innerHTML = "<div";
```

vitiral: 1. I hadn't considered that idea, but that sounds good.
2. We could use [`DOMParser`](https://developer.mozilla.org/en-US/docs/Web/API/DOMParser) and force the user to specify the type: normally `"text/html"`, but can also be `"image/svg+xml"` or `"application/xml"`.

However, I'm perplexed since your code _is_ a syntax error. The folowing works for me:

```
        var xmldoc = document.implementation.createDocument(null, null);
        var span = xmldoc.createElement("span");
        span.innerHTML = "<div></div>";·
```

koute: > However, I'm perplexed since your code is a syntax error. The folowing works for me:

That is the point. (: (Although AFAIK the code itself is valid, it's just that a `SyntaxError` is thrown. Unless you meant something else here?)

In my snippet the assignment to `innerHTML` throws a `SyntaxError` exception. That was just an example to illustrate that we need to handle that `SyntaxError` here somehow if we're going to use `innerHTML`.

vitiral:
Oh yes, i understand now

vitiral:
So I was thinking of returning a different error type if `childNodes.length != 1`, but that would mean that I have to create an error enum like this:

```
/// Errors that can be encountered when converting HTML.
pub enum HtmlConversionError {
    NumNodes(String),
    Syntax(SyntaxError),
}
```

I don't think the `NumNodes` error is all that interesting so I'm opting to panic in that situation. My reasoning is the same reason we panic if you are accessing an array element that doesn't exist -- you better have control of the html you are creating!

vitiral:
hmm... I just realized that `js_try!` cannot actually return any values -- so this is going to be considerably more difficult than I thought...

vitiral:
okay, it looks like other areas use custom `catch` -- so I will do that as well.

vitiral:
@koute issues should be fixed (and also rebased).

vitiral:
Okay, this is bizare:

```
>>> var span = document.createElement("span");
undefined
>>> span.innerHTML = "<div>foo</di";
"<div>foo</di"
>>> span.innerHTML
"<div>foo</div>"
```

It looks like it auto-fixed it. Same with `<div>foo`. Setting to `<di` just sets it to nothing `""`. I cannot actually trigger a syntax error here. Which makes me think that we should _only_ panic. Thoughts?

Pauan:
That's because the HTML parsing algorithm is intentionally designed to be ***unbelievably*** forgiving, so that way even novices can write HTML and have it work (for a given definition of "work"...)

src/webapi/node.rs
```diff
@@ -362,6 +362,58 @@ pub struct Node( Reference );
...
+    ///
+    /// ```
+    /// let node = Node::from_html("<div>Some text, horray!</div>").unwrap();
+    /// ```
+    pub fn from_html(html: &str) -> Result<Node, SyntaxError> {
+        let result = js! {
+            var span = document.createElement("span");
+            try {
+                span.innerHTML = @{html};
+            } catch( error ) {
```
koute: You can probably use the `js_try!` macro here, you know. It should be simpler. (:

src/webapi/node.rs
```diff
@@ -362,6 +362,58 @@ pub struct Node( Reference );
...
+                    success: false,
+                };
+            }
+            return span.childNodes[0];
+        };
+
+
+        if let Value::Bool(false) = js!{ return @{&result}.success; } {
+            if let Value::String(e) = js! { return @{&result}.panic; } {
+                panic!(e);
```
koute: Hmmm... do we actually want to panic here?

I mean - putting a snippet of HTML here with multiple roots is *kind-of* a syntax error, so perhaps we should just return `SyntaxError` in such a case with our own error message?

It might not make much sense to treat the case where the HTML has an actual syntax error and the case where the HTML has multiple roots separately. *Both* cases should (in theory) never happen if we control the HTML (assuming we didn't make a mistake), and both *can* happen if we just paste the HTML from an external source.

And in general I don't think we currently don't have any other panics like this one anywhere, so it'd would be a first.

vitiral: I think that reasoning makes sense. I guess I just create it [using the js! macro with SyntaxError](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/SyntaxError).

I'm kind of concerned that such an error gets created for number of nodes but is NOT created if the html is invalid (see above examples) -- but I guess that is just the harm of working with javascript!

src/webapi/node.rs
vitiral: `js_try!` doesn't return a value currently and I couldn't get it to without a lot of fiddling (and breaking tests... and regressing it's performance), so I opted for this route. It could be replaced in the future if it _did_ return a value.

src/webapi/node.rs
koute: Well, yeah, if your document is HTML then the snippet you pass to `innerHTML` is treated very loosely. That's an unfortunate historic artifact.

vitiral: I mean, we could always validate it as valid html _in the future_ if such a thing were useful (I think it would be).

src/webapi/node.rs
koute: But it does return a value? (: (See the `js_try` test in `macros.rs`)

src/webapi/node.rs
koute: I thought about it too, but I'm not sure whenever something like that would pull its own weight. Maybe in debug mode? A full HTML parser for this little snippet seems like an overkill. (Alternatively we could use the `document.implementation.createDocument(null, null);` trick to create a dummy XHTML document and abuse the browser to do the validation for us, even where our document is HTML.)

src/webapi/node.rs
vitiral: ah crap, I only looked at the `no_return` pattern :laughing:

src/webapi/node.rs
vitiral: Just to have a reference, that would be:

```
var xmldoc = document.implementation.createDocument(null, null);
var span = xmldoc.createElement("span"); 
span.innerHTML = "<div"; // Will throw SyntaxError
```
Seems hacky, but I much prefer requiring valid syntax. So sounds good.

koute: Yeah, something like that; I think that would be fine to have in debug mode.

vitiral:
So I tried a couple of things and failed.

- `js_try!` macro _kind of_ works, but I'm getting an error that the error is not a syntax error
  - `thread 'main' panicked at 'was error but not syntax error: Custom("reference is of a different type")', libcore/result.rs:945:5`

- the xml stuff was not good for me:
  - changed the name of the element from `"div"` to `"DIV"`
  - had another issue that I forgot now...

Help with `js_try!` would be much appreciated. I can't seem to figure out the type signature or something.

src/webapi/node.rs
```diff
@@ -362,6 +362,45 @@ pub struct Node( Reference );
...
+    ///
+    /// ```
+    /// let node = Node::from_html("<div>Some text, horray!</div>").unwrap();
+    /// ```
+    pub fn from_html(html: &str) -> Result<Node, SyntaxError> {
+        let result: Result<Value, Value> = js_try!(
+            var span = document.createElement("span");
+            span.innerHTML = @{html};
+            if( span.childNodes.length != 1 ) {
+                throw SyntaxError("HTML had ${span.childNodes.length} nodes but must have 1");
```
koute: We should not use ES6 features in `js!` snippets. (:

Also, perhaps an error like this would be better? `"Node::from_html called on a snippet with multiple root nodes"`

vitiral: what's the ES6 feature?

vitiral: Also, you would get the error if you have 0 nodes as well, which is why I included the number.

koute: > what's the ES6 feature? (I'm such a js noob)

The string interpolation. `${span.childNodes.length}`

> Also, you would get the error if you have 0 nodes as well, which is why I included the number.

Hmm.. yes, you're right. Then perhaps we could have an inner `if` which would check for that? Or just say that `"Node::from_html requires a single root node"`? I don't really care *that* much either way though, as long as the error message is clear what went wrong.

vitiral: I did

                throw SyntaxError(
                    "Node::from_html requires a single root node but has: "
                    + toString(span.childNodes.length)
                );

vitiral: Still having issues with the test failing though. Don't know why (says that the resulting error is `CustomError` instead of `SyntaxError`, even when I'm throwing `SyntaxError`)

Pauan: I don't think `toString` is valid, you should just remove it:
```
throw SyntaxError(
    "Node::from_html requires a single root node but has: " + span.childNodes.length
);
```

JS automatically converts things into strings when you try to add it to a string.

src/webapi/node.rs
```diff
@@ -362,6 +362,48 @@ pub struct Node( Reference );
...
+                    .expect("inner type was not a node.");
+                Ok(node)
+            }
+            Err(e) =>  {
+                let err = SyntaxError::try_from(e)
+                    .expect("was error but not syntax error");
+                Err(err)
+            }
+        }
+    }
```
koute: This whole `match` is unnecessary with `js_try!` and you can simply remove it.

src/webapi/node.rs
```diff
@@ -362,6 +362,48 @@ pub struct Node( Reference );
...
+    /// it is recommended to at _least_ have programatic control of your root
+    /// HTML node.
+    ///
+    /// # Examples
+    ///
+    /// ```
+    /// let node = Node::from_html("<div>Some text, horray!</div>").unwrap();
+    /// ```
+    pub fn from_html(html: &str) -> Result<Node, SyntaxError> {
+        let result: Result<Value, Value> = js_try!(
```
koute: The `js_try!` can convert to the necessary types by itself, so you can get rid of `let: Result<Value, Value>`.

src/webapi/node.rs
```diff
@@ -362,6 +362,48 @@ pub struct Node( Reference );
...
+            var span = document.createElement("span");
+            span.innerHTML = @{html};
+            if( span.childNodes.length != 1 ) {
+                throw SyntaxError(
```
koute: Shouldn't this have `new` before `SyntaxError`?

koute:
The reason why your test doesn't work is that `SyntaxError` isn't actually a `DOMException` (we have defined it incorrectly.)

I'll fix it in a minute.

koute:
Actually, there are two `SyntaxError`s, one which is a `DOMException`, and one which isn't. *Sigh*...

koute:
Ok, so, the issue here is that you're throwing the ECMAScript's `SyntaxError` (which is **not** an `DOMException`) while the `innerHTML` throws the `DOMException` variant of `SyntaxError`. So, instead of `new SyntaxError` you should use `new DOMException("...message here...", "SyntaxError")`, and then it should work.

vitiral:
Wow, thanks for triaging that. I'll get to it soon.

koute:
LGTM; thanks!

PR #201, opened 2018-04-11
fix #73: add docs for no_return

### LINKED ISSUE #73: Document @(no_return) flag for the js! macro

vitiral:
also mentioned the importance of including a `return` value for other cases.

### DIFF
--- a/src/webcore/macros.rs
+++ b/src/webcore/macros.rs
@@ -464,6 +464,10 @@ macro_rules! _js_impl {
 ///
 /// # Examples
 ///
+/// ## Regular Usage
+///
+/// Note: you **must** include the `return ...;` statement to get a value.
+///
 /// ```
 /// let name = "Bob";
 /// let result = js! {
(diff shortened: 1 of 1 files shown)


src/webcore/macros.rs
```diff
@@ -464,6 +464,10 @@ macro_rules! _js_impl {
 ///
 /// # Examples
 ///
+/// ## Regular Usage
+///
+/// Note: you **must** include the `return ...;` statement to get a value.
```
koute: Hmm... wouldn't it be better to move this to the bottom of this subsection?

src/webcore/macros.rs
```diff
@@ -473,6 +477,17 @@ macro_rules! _js_impl {
 ///
 /// println!( "2 + 2 = {:?}", result );
 /// ```
+///
+/// ## No Return
+///
+/// This will moderately increase performance when there is no return value.
```
koute: Hmm... I'm not sure if `moderately` is appropriate here, as I haven't really benchmarked it so I would rather not oversell it. (:

Perhaps a phrasing such as this would be better?

> If you don't need to return a value from your snippet you can add a `@(no_return)` attribute to slightly improve performance.

Side note: once procedural macros stabilize I'm planning on automatically derive whenever a given snippet returns a value or not; this attribute will then become unnecessary.

vitiral:
@koute thanks, done!

koute:
Thanks!

PR #202, opened 2018-04-15
Add DocumentFragment, get_attribute in Node, fix compiling warning

togekk:
- Add DocumentFragment
- Add create_document_fragment in Node
- Add get_attribute in Node
- Mark TryInto line in Location, so that the warning won't show up while compiling

### DIFF
--- a/src/webapi/document.rs
+++ b/src/webapi/document.rs
@@ -35,12 +35,11 @@ pub fn document() -> Document {
 }
 
 impl Document {
-    /// In an HTML document, the Document.createElement() method creates the HTML
-    /// element specified by `tag`, or an HTMLUnknownElement if `tag` isn't
-    /// recognized. In other documents, it creates an element with a null namespace URI.
+    /// In an HTML document, the Document.createDocumentFragment() method creates a
+    /// new empty DocumentFragment.
     ///
-    /// [(JavaScript docs)](https://developer.mozilla.org/en-US/docs/Web/API/Document/createElement)
-    // https://dom.spec.whatwg.org/#ref-for-dom-document-createelement
+    /// [(JavaScript docs)](https://developer.mozilla.org/en-US/docs/Web/API/Document/createDocumentFragment)
+    // https://dom.spec.whatwg.org/#ref-for-dom-document-createdocumentfragment
     pub fn create_document_fragment( &self ) -> Result< DocumentFragment, TODO > {
         unsafe {
             Ok( js!( return @{self}.createDocumentFragment(); ).into_reference_unchecked().unwrap() )
--- a/src/webapi/document.rs
+++ b/src/webapi/document.rs
@@ -4,6 +4,7 @@ use webapi::event_target::{IEventTarget, EventTarget};
 use webapi::node::{INode, Node};
 use webapi::element::Element;
 use webapi::html_element::HtmlElement;
+use webapi::document_fragment::DocumentFragment;
 use webapi::text_node::TextNode;
 use webapi::location::Location;
 use webapi::parent_node::IParentNode;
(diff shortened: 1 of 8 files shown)


src/webapi/document.rs
```diff
@@ -34,6 +35,17 @@ pub fn document() -> Document {
 }
 
 impl Document {
+    /// In an HTML document, the Document.createDocumentFragment() method creates a
+    /// new empty DocumentFragment.
+    ///
+    /// [(JavaScript docs)](https://developer.mozilla.org/en-US/docs/Web/API/Document/createDocumentFragment)
+    // https://dom.spec.whatwg.org/#ref-for-dom-document-createdocumentfragment
+    pub fn create_document_fragment( &self ) -> Result< DocumentFragment, TODO > {
```
koute: Why does this return a `Result`? Can `createDocumentFragment` actually throw?

src/webapi/location.rs
```diff
@@ -1,5 +1,5 @@
 use webcore::value::Reference;
-use webcore::try_from::TryInto;
+// use webcore::try_from::TryInto;
```
koute: This should probably not be in this PR (:

src/webapi/node.rs
```diff
@@ -341,6 +341,18 @@ pub trait INode: IEventTarget {
...
+    /// If the given attribute does not exist, the value returned will either be
+    /// null or "" (the empty string);
+    ///
+    /// [(Javascript docs)](https://developer.mozilla.org/en-US/docs/Web/API/Element/getAttribute)
+    // https://dom.spec.whatwg.org/#ref-for-dom-element-getattribute
+    fn get_attribute( &self, name: &str ) -> Option< String > {
+        js!(
+            return @{self.as_ref()}.getAttribute( @{name} );
+        ).try_into().unwrap()
+    }
```
koute: We already have this method in `IElement::get_attribute`, and AFAIK the `get_attribute` is not part of the `Node` interface so this should be removed.

Any reason why did you add it here?

src/webapi/document.rs
togekk: Sorry I'm new to this and I don't quite understand what Result does to be honest....but I guess DocumentFragment has the basic functions of Element so I copied and modified create_element function. Perhaps there's a better way to write it?

But I've tested it in my project and the DocumentFragment did actually created. And it does append_child and be appended too. Here's an example:

Live: https://togekk.github.io/stdweb-docfrag/
Souce Code: https://github.com/togekk/rust-mvc/blob/documentfragment/src/main.rs

Pauan: @togekk Unlike most other languages, Rust doesn't have exceptions. So the normal way of handling errors is to use `Result`, which makes it clear which things can error and which things cannot error.

I just now looked at the spec, it seems that `createDocumentFragment` never throws errors, so this should return `DocumentFragment`, not `Result<DocumentFragment, TODO>`

Pauan: Also, please do ask if you do not understand anything, we're happy to help!

src/webapi/node.rs
togekk: Sorry my mistake. I skipped the help message and finally get it work now. So yes this shoud be removed

src/webapi/document.rs
togekk: Thanks for explaining! So should I commit again or...?

src/webapi/node.rs
```diff
@@ -341,6 +341,7 @@ pub trait INode: IEventTarget {
             @{self.as_ref()}.normalize();
         }
     }
+
```
koute: Stray newline. (Sorry. (:)

togekk: No worries my bad. Now it's all done :D

PR #206, opened 2018-04-19
Futures 0.2

Pauan:
I've tested that it works, and the performance is the same as the old executor.

Breaking changes:

* `PromiseFuture::spawn` has been renamed to `PromiseFuture::spawn_local` (for consistency with LocalExecutor)

* `PromiseFuture::spawn_local` now requires the error type to be `Never` (it used to be `()`)

### DIFF
--- a/src/webcore/promise_future.rs
+++ b/src/webcore/promise_future.rs
@@ -2,14 +2,15 @@ use std;
 use webcore::value::{Value, ConversionError};
 use webcore::try_from::{TryInto, TryFrom};
 use webapi::error;
-use futures::{Future, Poll, Async};
-use futures::unsync::oneshot::Receiver;
+use futures_core::{Future, Poll, Async, Never};
+use futures_core::task::Context;
+use futures_channel::oneshot::Receiver;
 use webcore::executor::spawn;
 use webcore::discard::DiscardOnDrop;
 use super::promise::{Promise, DoneHandle};
 
 
-/// This allows you to use a JavaScript [`Promise`](struct.Promise.html) as if it is a Rust [`Future`](https://docs.rs/futures/0.1.*/futures/future/trait.Future.html).
+/// This allows you to use a JavaScript [`Promise`](struct.Promise.html) as if it is a Rust [`Future`](https://docs.rs/futures/0.2.*/futures/future/trait.Future.html).
 ///
 /// The preferred way to create a `PromiseFuture` is to use [`value.try_into()`](unstable/trait.TryInto.html) on a JavaScript [`Value`](enum.Value.html).
 ///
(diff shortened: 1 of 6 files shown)


src/webcore/promise_future.rs
```diff
@@ -83,6 +84,17 @@ impl PromiseFuture< (), Never > {
         B: Future< Item = (), Error = Never > + 'static {
         spawn_local( future );
     }
+
+    /// This is used to print an error to the console.
+    ///
+    /// See the documentation for [`spawn_local`](#method.spawn_local) for more details.
+    #[inline]
+    pub fn print_error< A: JsSerialize >( value: A ) -> Never {
```
koute: I feel like the naming of this method is slightly misleading, as it doesn't just print the error - it also panics. Perhaps simply calling it `panic` would be better?

Pauan: I think the `Never` return type is good enough to indicate that it panics/never returns.

Calling it `panic` is wrong because it doesn't just panic (and in fact panicking is not its primary purpose).

koute: But you don't see the `Never` type when reading code, you see `.map_err(PromiseFuture::print_error)` and it's not obvious that it's going to panic. And the `Never` type (or `!`) for that matter *in theory* can also mean that it goes into an infinite loop.

The user should be always fully aware that their code can panic, especially because currently panics can cause issues like memory leaks and `RefCell`s getting stuck in a borrowed state.

Pauan: > But you don't see the Never type when reading code, you see .map_err(PromiseFuture::print_error) and it's not obvious that it's going to panic.

Indeed, though that's just how Futures work: they require `Never` for the error type, so if you're using `map_err` it's pretty safe to assume it panics. So anybody with familiarity with Futures 0.2 should understand what's going on.

> The user should be always fully aware that their code can panic

A good principle, yes. Though in both the standard library *and* stdweb that isn't enforced *by the name*, instead it's usually done by adding in a `# Panics` section to the docs, which I will add.

In any case, `panic` alone is a really bad name. `error_panic` isn't good either. `print_error_panic` is okay, but quite long. Maybe it's acceptable.

Pauan: > especially because currently panics can cause issues like memory leaks and RefCells getting stuck in a borrowed state.

That doesn't sound right. Right now in wasm panic = abort, so that won't ever happen.

And even with normal panics I can't think of any way to abuse panics like that: when a panic happens, it drops everything in scope.

Do you have any examples (even theoretical/hypothetical ones)?

Pauan:
I made it clear in the docs that `print_error` panics.

src/webcore/promise_future.rs
koute: > In any case, panic alone is a really bad name. error_panic isn't good either. print_error_panic is okay, but quite long. Maybe it's acceptable.

Or maybe `print_and_panic`?

> That doesn't sound right. Right now in wasm panic = abort, so that won't ever happen. And even with normal panics I can't think of any way to abuse panics like that: when a panic happens, it drops everything in scope.

A panic will, not drop anything and generate a `RuntimeException` on the JS side. On other targets this is not a problem since an abort will kill the problem, but here there is nothing preventing the JS code from calling into the Rust code again.

koute: E.g. a program like this:

```
#[macro_use]
extern crate stdweb;

struct Dummy;
impl Drop for Dummy {
    fn drop( &mut self ) {
        console!( log, "Drop called" );
    }
}

fn main() {
    let dummy = Dummy;
    panic!("");
}
```

On emscripten targets the object will be properly dropped, but on `wasm32-unknown-unknown` it will not.

Pauan: Hmmm, you're right that JS can call back into wasm again, I hadn't thought of that.

Pauan:
I renamed it to `print_error_panic`. Because of the way that it's used (only once in `main`, and using `map_err`), giving it a particularly long name isn't a big deal.

koute:
Hmm... I'm not entirely satisfied by the name, but since this isn't stable anyway it can stay this way for now.

Thanks!

PR #209, opened 2018-04-26
Add support for HTMLCollection, <option>, <select>

limira:
Tests are in `select.rs`.

### DIFF
--- a/src/webapi/html_collection.rs
+++ b/src/webapi/html_collection.rs
@@ -2,11 +2,11 @@ use webcore::value::Reference;
 use webcore::try_from::TryInto;
 use webapi::element::Element;
 
-/// The HTMLCollection interface represents a generic collection
+/// The `HtmlCollection` interface represents a generic collection
 /// (array-like object similar to arguments) of elements (in document order)
 /// and offers methods and properties for selecting from the list.
 /// 
-/// An HTMLCollection in the HTML DOM is live; it is automatically
+/// An `HtmlCollection` in the HTML DOM is live; it is automatically
 /// updated when the underlying document is changed.
 /// 
 /// [(JavaScript docs)](https://developer.mozilla.org/en-US/docs/Web/API/HTMLCollection)
@@ -24,7 +24,7 @@ impl HtmlCollection {
         js!( return @{self}.length; ).try_into().unwrap()
     }
 
-    /// Returns an element from an HtmlCollection by index.
+    /// Returns an element from an `HtmlCollection` by index.
     ///
     /// [(JavaScript docs)](https://developer.mozilla.org/en-US/docs/Web/API/HTMLCollection/item)
     // https://dom.spec.whatwg.org/#ref-for-dom-htmlcollection-item
@@ -34,7 +34,7 @@ impl HtmlCollection {
         ).try_into().unwrap()
     }
 
-    /// Returns an iterator over the list.
+    /// Returns an iterator over the collection.
     pub fn iter( &self ) -> ElementIter {
         ElementIter {
             list: self.clone(),
--- a/src/webapi/html_collection.rs
+++ b/src/webapi/html_collection.rs
@@ -1,0 +1,96 @@
+use webcore::value::{Value, Reference};
+use webcore::reference_type::ReferenceType;
+use webcore::try_from::TryInto;
+use webapi::element::Element;
+
+/// The HTMLCollection interface represents a generic collection
+/// (array-like object similar to arguments) of elements (in document order)
+/// and offers methods and properties for selecting from the list.
+/// 
+/// An HTMLCollection in the HTML DOM is live; it is automatically
+/// updated when the underlying document is changed.
+/// 
+/// [(JavaScript docs)](https://developer.mozilla.org/en-US/docs/Web/API/HTMLCollection)
+// https://dom.spec.whatwg.org/#interface-htmlcollection
+#[derive(Clone, Debug, PartialEq, Eq, ReferenceType)]
+#[reference(instance_of = "HTMLCollection")]
+pub struct HtmlCollection( Reference );
+
+impl HtmlCollection {
+    /// Returns the number of elements in the collection.
+    ///
+    /// [(JavaScript docs)](https://developer.mozilla.org/en-US/docs/Web/API/HTMLCollection)
+    // https://dom.spec.whatwg.org/#dom-htmlcollection-length
+    pub fn len( &self ) -> u32 {
+        js!( return @{self}.length; ).try_into().unwrap()
+    }
+
+    /// Returns an element from an HtmlCollection by index.
+    ///
+    /// [(JavaScript docs)](https://developer.mozilla.org/en-US/docs/Web/API/HTMLCollection/item)
+    // https://dom.spec.whatwg.org/#dom-htmlcollection-item
+    pub fn item( &self, index: u32 ) -> Option< Element > {
+        js!(
+            return @{self}[ @{index} ];
+        ).try_into().unwrap()
+    }
+
+    /// Returns an iterator over the list.
+    pub fn iter( &self ) -> ElementIter {
+        ElementIter {
+            list: self.clone(),
+            index: 0
+        }
+    }
+}
+
+
+impl IntoIterator for HtmlCollection {
+    type Item = Element;
+    type IntoIter = ElementIter;
+
+    #[inline]
+    fn into_iter( self ) -> Self::IntoIter {
+        ElementIter {
+            list: self,
+            index: 0
+        }
+    }
+}
+
+impl< 'a > IntoIterator for &'a HtmlCollection {
+    type Item = Element;
+    type IntoIter = ElementIter;
+
+    #[inline]
+    fn into_iter( self ) -> Self::IntoIter {
+        ElementIter {
+            list: self.clone(),
+            index: 0
+        }
+    }
+}
+
+#[derive(Debug)]
+pub struct ElementIter {
+    list: HtmlCollection,
+    index: i32
+}
+
+impl Iterator for ElementIter {
+    type Item = Element;
+    fn next( &mut self ) -> Option< Self::Item > {
+        let value = js!(
+            return @{&self.list}[ @{self.index} ];
+        );
+
+        let element = match value {
+            Value::Undefined => return None,
+            Value::Reference( reference ) => unsafe { Element::from_reference_unchecked( reference ) },
+            _ => unreachable!()
+        };
+
+        self.index += 1;
+        Some( element )
+    }
+}
--- a/src/webapi/html_collection.rs
+++ b/src/webapi/html_collection.rs
@@ -1,5 +1,4 @@
-use webcore::value::{Value, Reference};
-use webcore::reference_type::ReferenceType;
+use webcore::value::Reference;
 use webcore::try_from::TryInto;
 use webapi::element::Element;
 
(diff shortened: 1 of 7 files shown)


src/webapi/html_elements/select.rs
```diff
@@ -0,0 +1,164 @@
...
+        let selected_index = selected_index.unwrap_or(-1);
+        js!{
+            @(no_return)
+            @{self}.selectedIndex = @{selected_index};
+        }
+    }
+
+    /// Returns the `Some(value)` of the first selected item, if any, or `None` if there is no selected item.
+    // https://html.spec.whatwg.org/#dom-select-value
+    pub fn value(&self) -> Option<String> {
```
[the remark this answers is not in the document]
Pauan: Rather than using `selectedOptions` I think it would be better to use `selectedIndex`, like this:

```rust
pub fn value(&self) -> Option<String> {
    js!(
        var self = @{self};

        if (self.selectedIndex < 0) {
            return null;

        } else {
            return self.value;
        }
    ).try_into().unwrap()
}
```

src/webapi/html_collection.rs
```diff
@@ -0,0 +1,96 @@
...
+        js!( return @{self}.length; ).try_into().unwrap()
+    }
+
+    /// Returns an element from an HtmlCollection by index.
+    ///
+    /// [(JavaScript docs)](https://developer.mozilla.org/en-US/docs/Web/API/HTMLCollection/item)
+    // https://dom.spec.whatwg.org/#dom-htmlcollection-item
+    pub fn item( &self, index: u32 ) -> Option< Element > {
+        js!(
+            return @{self}[ @{index} ];
```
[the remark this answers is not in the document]
Pauan: I'm not so sure about that: I know that [some browsers can optimize `[]` a *lot* better than `.item()`](https://jsperf.com/nodelist-item-x-vs-nodelist-x2). (e.g. Chrome runs `[]` over twice as fast!)

Of course [microbenchmarks should be taken with a *very* hefty grain of salt](https://www.youtube.com/watch?v=g0ek4vV7nEA), but still.

src/webapi/html_elements/select.rs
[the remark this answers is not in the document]
koute: Aye, that would be better.

src/webapi/html_collection.rs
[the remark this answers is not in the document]
limira: NodeList (node_list.rs) also use `[]`, do you what to change it to `.item` as well?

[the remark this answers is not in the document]
limira: Run benchmark linked by @Pauan on my system: 

* Firefox 59: `[]` 41% slower than `.item`
* Chromium 65: `.item` 46% slower than `[]`

So, I will use `.item` for an `.item`.

Note that what I have done for `html_collection.rs` is copy from `node_list.rs`. After this discussion, I found out that I can use `HtmlCollection::item` in its iterator (in `next` method). Please take a look to `html_collection.rs` (`item`, `ElementIter`), are they ok? (they are covered by tests in `select.rs`)

Side note: While working on this I found out that `node_list.rs` use an `i32` in its iterator instead of `u32`.

src/webapi/html_elements/select.rs
```diff
@@ -0,0 +1,164 @@
...
+        if value == "" {
+            None
+        } else {
+            Some(value)
+        }
+    }
+
+    /// Change the selected value to the given value.
+    // https://html.spec.whatwg.org/#dom-select-value
+    pub fn set_value(&self, value: Option<String>) {
```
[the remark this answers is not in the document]
limira: Since `pub fn value(&self) -> Option<String>` is still return `Option<String>`, I will `set_value` as:
```rust
    pub fn set_value(&self, value: Option<String>) {
        match value{
            Some(value) => js!{
                @(no_return)
                @{self}.value = @{value};
            },
            None => js!{
                @(no_return)
                @{self}.selectedIndex = -1;
            }
        };
    }
```

limira:
## Summary changes:

1. Update specs links
2. Run benchmark (`[]` vs `.item`) linked by @Pauan on my system: 

* Firefox 59: `[]` 41% slower than `.item`
* Chromium 65: `.item` 46% slower than `[]`

So, I will use `.item` for an `.item`.

Note that what I have done for `html_collection.rs` is copy from `node_list.rs`. After this discussion, I found out that I can use `HtmlCollection::item` in its iterator (in `next` method). Please take a look to `html_collection.rs` (`item`, `ElementIter`), are they ok? (they are covered by tests in `select.rs`)

3. Keep `pub fn value(&self) -> Option<String>`, implement as @Pauan suggested.
4. Since `SelectElement::value` is still return `Option<String>`,  `set_value` will still receive `Option<String>` and its new implement is as below:
```rust
    pub fn set_value(&self, value: Option<String>) {
        match value{
            Some(value) => js!{
                @(no_return)
                @{self}.value = @{value};
            },
            None => js!{
                @(no_return)
                @{self}.selectedIndex = -1;
            }
        };
    }
```

src/webapi/html_elements/select.rs
```diff
@@ -0,0 +1,181 @@
...
+    /// An [HtmlCollection](struct.HtmlCollection.html) representing
+    /// the set of <option> elements that are selected.
+    // https://html.spec.whatwg.org/#the-select-element:dom-select-selectedoptions
+    pub fn selected_options(&self) -> HtmlCollection {
+        js!(
+            return @{self}.selectedOptions;
+        ).try_into().unwrap()
+    }
+
+    /// A convenient method to get values of all selected <option> elements
```
koute: `convenient` -> didn't you mean `convenience`? (:

Also `<option>` here is unquoted too, so it will break the docs. (Ditto for the next method.)

src/webapi/html_elements/select.rs
```diff
@@ -0,0 +1,181 @@
...
+                return null;
+            } else {
+                return self.value;
+            }
+        ).try_into().unwrap()
+    }
+
+    /// Change the selected value to the given value.
+    // https://html.spec.whatwg.org/#the-select-element:dom-select-value
+    pub fn set_value(&self, value: Option<String>) {
```
koute: Hmm.... I'm not convinced we actually need to take the `Option<String>` here? I mean, there is already `set_selected_index(None)` that one can use. I know that `value` returns `Option<String>`, but that's normal for a getter; setters usually don't take an `Option`.

Pauan: I think it should be consistent with the getter, and I also think it's very useful to be able to (easily and intuitively) unset the value.

Pauan: Having the setter take `Option<String>` signals that it is possible for it to *not* have a value at all.

If it takes `String` then that implies that it always has a value, which is not true. And then if the user *wants* to unset the value, they need to use a completely different API, which doesn't sound good to me.

Pauan: Also, the decision for `get_value` to return an `Option<String>` (rather than requiring the user to use `get_selected_index`) was to prevent mistakes, and make the API more intuitive. I believe the same argument applies to `set_value`

src/webapi/html_elements/select.rs
limira: * What I means is `convenient method` (`adj+noun`). But my English skill is just average, so I am not sure any thing that involves `grammar`.
* About: `<option>`: Do you have a method to detect things like this? After this catch by you, I actually ran a `cargo doc`, go seeing the result docs, and came back to correct the source code docs... then repeat the tasks until not seeing something weird in the result docs.

koute: 1. In these kinds of contexts you'd usually use `convenience` instead of `convenient` (:

2. I don't; I just usually catch these kinds of issues during the review. Usually you can just simply run `cargo doc`, open the docs and look through the docs for the code you've added/changed if it looks correct.

src/webapi/html_elements/select.rs
koute: Okay, fine, you can keep it `Option<String>`; just:

1. Call `set_selected_index` inside of setting it through `js!`.
2. In the `Some` case add a `debug_assert!` after setting the value to verify that `self.selected_index().is_some()` - if we're going to make the interface more bullet proof let's go all the way (otherwise both `None` and `Some("insert gibberish here")` will result in the value being unset). Or perhaps we could instead always check that and return a `Result`? (Basically make setting an invalid value into an error.) I don't like messing with those APIs too much, but they *are* pretty poorly designed. (@Pauan Your thoughts?)

Pauan: Hmmm... these are my thoughts on it:

1. Add in `raw_value` and `set_raw_value` which only work with `String` (not `Option<String>`), and they never error. So they're just a thin wrapper over the underlying DOM APIs.

2. Change `get_selected_index` and `set_selected_index` to use `u32` (with a `debug_assert` to ensure that it doesn't wrap around to negative)

3. Change `value` and `set_value` to call `get_selected_index` + `raw_value`, and `set_selected_index` + `set_raw_value`

4. Change `set_value` to return `Result`

koute: Hmm... sounds good to me!

@limira Let's do what Pauan proposed. For the error type returned by `value()` you can create a dedicated type for it and return it when the user passes an unknown value. (`UnknownVariantError`, perhaps?)

Pauan: I think `UnknownValueError` or `InvalidValueError` is more intuitive, since you're setting a value which doesn't exist.

koute: Those also sound fine to me!

Pauan:
@jaroslaw-weber This PR uses `js!` to add in the desired functionality, but there's nothing special about it: you can also use `js!` in your code as well:

```rust
fn raw_value(node: &Node) -> String {
    js!( return @{node}.value; ).try_into().unwrap()
}
```

So you don't need to wait: just use `js!` in your own code.

src/webapi/html_elements/select.rs
```diff
@@ -22,49 +39,45 @@ impl IElement for SelectElement {}
...
     }
 
-    /// Change the selected value to the given value.
-    // https://html.spec.whatwg.org/#the-select-element:dom-select-value
-    pub fn set_value(&self, value: Option<String>) {
-        match value{
-            Some(value) => js!{
+    /// Change selected index to the given value.
+    // https://html.spec.whatwg.org/#the-select-element:dom-select-selectedindex
+    pub fn set_selected_index(&self, selected_index: &Option<u32>) {
```
koute: Taking the `Option` by reference (`&`) is unnecessary.

Is there any particular reason why you made it so?

src/webapi/html_elements/select.rs
```diff
@@ -73,6 +86,35 @@ impl SelectElement {
...
+        match self.selected_index(){
+            None => None,
+            Some(_) => Some(self.raw_value())
+        }
+    }
+
+    /// Change the selected value to the given value. If you provide an invalid value,
+    /// the `<select>` element will have no item selected, and an `UnknownValueError` is returned.
+    // https://html.spec.whatwg.org/#the-select-element:dom-select-value
+    pub fn set_value(&self, value: &Option<String>) -> Result<(), UnknownValueError> {
```
koute: This should take `Option<&str>` instead of `Option<String>`.

limira:
I hope this is the last commit require for this PR!

Thank you for all your guide and help! This is my very first contribution to opensoure software/Rust ecosystem. Think you have very careful attention while reviewing my work. Thanks again!

Of course, if you find out any new issue, I will fix it!

koute:
LGTM! Thanks a lot! (:

PR #218, opened 2018-05-13
Fix typo: `toDataUrl` -> `toDataURL`

Victor-Savu:
The canvas element method for getting the image data as a URL-encoded string is [`toDataURL`](https://developer.mozilla.org/en-US/docs/Web/API/HTMLCanvasElement/toDataURL)


koute:
Good catch! Thanks!

PR #223, opened 2018-05-26
Add Gamepad API

coryshrmn:
This PR fully wraps the [current Gamepad API spec](https://w3c.github.io/gamepad/).

The [Gamepad API](https://developer.mozilla.org/en-US/docs/Web/API/Gamepad_API) is important for Rust browser games. It is [supported on all major browsers](https://caniuse.com/#feat=gamepad).

`GamepadEvent` and `GamepadButtonMapping` are unit-tested. The rest of the API is not testable, because `Gamepad` and `GamepadButton` are not constructible. 

In lieu of tests, I added a Gamepad example ([deployed here](https://coryshrmn.github.io/stdweb/examples/gamepad/deploy/)). It demonstrates the entire API (except `GamepadButton::touched()`, which is not yet supported in Chrome). I verified the example works perfectly on Firefox, Chrome, and Safari.

Please let me know if there is anything else I can do.
Thanks!

### DIFF
--- a/src/webapi/gamepad.rs
+++ b/src/webapi/gamepad.rs
@@ -191,3 +191,27 @@ pub fn get_gamepads() -> Vec<Option<Gamepad>> {
         return Array.from(navigator.getGamepads());
     ).try_into().unwrap()
 }
+
+#[cfg(test)]
+mod tests {
+    use super::GamepadMappingType;
+
+    use webcore::try_from::TryInto;
+    use webcore::value::{ConversionError, Value};
+
+    #[test]
+    fn test_value_into_gamepad_mapping() {
+
+        let to_mapping = |v: Value| -> Result<GamepadMappingType, ConversionError> {
+            v.try_into()
+        };
+
+        assert_eq!(to_mapping("standard".into()), Ok(GamepadMappingType::Standard));
+        assert_eq!(to_mapping("".into()), Ok(GamepadMappingType::NoMapping));
+        assert!(to_mapping("fakemapping".into()).is_err());
+        assert!(to_mapping(Value::Null).is_err());
+    }
+
+    // most of the Gamepad API is not testable,
+    // because Gamepad and GamepadButton are not constructible
+}
--- a/src/webapi/gamepad.rs
+++ b/src/webapi/gamepad.rs
@@ -1,8 +1,15 @@
 //use std::fmt::Debug;
 
 use webcore::reference_type::ReferenceType;
-use webcore::try_from::TryInto;
-use webcore::value::Reference;
+use webcore::try_from::{
+    TryFrom,
+    TryInto,
+};
+use webcore::value::{
+    ConversionError,
+    Reference,
+    Value,
+};
 
 /// The set of known gamepad layout mappings.
 ///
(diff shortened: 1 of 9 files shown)


src/webapi/gamepad.rs
```diff
@@ -0,0 +1,217 @@
...
+#[derive(Clone, Debug, Eq, PartialEq, ReferenceType)]
+#[reference(instance_of = "Gamepad")]
+pub struct Gamepad( Reference );
+
+impl IGamepad for Gamepad {}
+
+/// Retrieve all connected gamepads, in an array indexed by each gamepad's `index` member.
+///
+/// [(JavaScript docs)](https://developer.mozilla.org/en-US/docs/Web/API/Navigator/getGamepads)
+// https://w3c.github.io/gamepad/#dom-gamepad-axes
```
koute: Wrong specs link. (:

src/webapi/gamepad.rs
```diff
@@ -0,0 +1,217 @@
...
+pub fn get_gamepads() -> Vec<Option<Gamepad>> {
```
koute: This would be better as `Gamepad::get_all` or something like that.

src/webapi/gamepad.rs
```diff
@@ -0,0 +1,217 @@
...
+/// A reference to a JavaScript object which implements the [IGamepad](trait.IGamepad.html)
+/// interface.
+///
+/// [(JavaScript docs)](https://developer.mozilla.org/en-US/docs/Web/API/Gamepad)
+// https://w3c.github.io/gamepad/#gamepad-interface
+#[derive(Clone, Debug, Eq, PartialEq, ReferenceType)]
+#[reference(instance_of = "Gamepad")]
+pub struct Gamepad( Reference );
+
+impl IGamepad for Gamepad {}
```
koute: Hmm... since those are the only impementations we probably should get rid of the `IGamepadButton` and `IGamepad` interfaces and just define all of those methods in a normal `impl` block.

src/webapi/gamepad.rs
coryshrmn: fixed, thanks :)

src/webapi/gamepad.rs
coryshrmn: sounds good. done!

src/webapi/gamepad.rs
koute: If we absolutely want to adhere to the specs we could add a `navigator` method (just as we do for e.g. [document](https://docs.rs/stdweb/*/stdweb/web/fn.document.html)) and then define `get_gamepads` on it. Then it would be obvious that it's a 1-to-1 call.

In extra convenience methods like this one we usually just pick a name that's more Rust-y, and in this case since the call will be prefixed with `Gamepad::` there is no point in repeating it. 

That said, it's not that much of a big deal, so I leave this decision to you.

coryshrmn: Okay. `Gamepad::get_all()` does sound nicer. I renamed to `Gamepad::get_all()`, and documented it should be called every frame.

koute:
LGTM! Thanks a lot!

PR #228, opened 2018-06-01
Use correct height in clear_rect

casey:
A small typo in clear_rect was making it use width for height. Fixed to use the height instead.


casey:
You're welcome, and no worries! I hope it was a nice tech-free two weeks ^_^

PR #231, opened 2018-06-04
[Looking for feedback] Custom instanceof for input elements/implement FileInputElement

### LINKED ISSUE #226: FileList API
I noticed that the `files` property of `InputElement` was removed some time ago. Are there any thoughts on how this should be reimplemented? In the meantime, are there any quick workarounds to get access to a `FileList`?

bgourlie:
I'm looking for feedback on the approach to subclassing input elements. This PR adds a custom `instanceof` implementation for input elements, and implements `FileInputElement`. Fixes #226.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -270,6 +270,7 @@ pub mod web {
     pub mod html_element {
         pub use webapi::html_elements::ImageElement;
         pub use webapi::html_elements::InputElement;
+        pub use webapi::html_elements::FileInputElement;
         pub use webapi::html_elements::TextAreaElement;
         pub use webapi::html_elements::CanvasElement;
         pub use webapi::html_elements::SelectElement;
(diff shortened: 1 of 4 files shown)


koute:
Sorry for the delay; I had no access to a computer for the last two weeks.

Yep, something like this! However I feel that the usecase here is way too specific to hardcode it into the `#[reference]` macro. What would be preferable is one of these two things:

a) a simple, internal `macro_rules!` macro which will inplement `InstanceOf` for these elements,
b) or (and this is the better option) extend the `#[reference]` macro in a way which doesn't hardcode anything, basically something like this:

```
#[reference(instance_of = "HTMLInputElement")]
#[reference(constraint=(type="file"))]
#[reference(subclass_of(EventTarget, Node, Element, InputElement))]
pub struct FileInputElement( Reference );
```

(This was started in https://github.com/koute/stdweb/pull/181 but not finished.)

I'm willing to accept (a) if you want to get this out of the door, but obviously (b) would be the best (and we can use the same mechanism to improve the `Event` casting.)

PR #232, opened 2018-06-06
Expose FileReaderReadyState as public

ctjhoa:
This is mandatory to check the FileReader state after using it.


koute:
Sorry for the delay; I had no access to a computer for the last two weeks.

LGTM! Thanks!

PR #240, opened 2018-06-24
Add support for XhrResponseType and XmlHttpRequest::response

ryanisaacg:
This allows loading binary files with XmlHttpRequset

### DIFF
--- a/src/webapi/xml_http_request.rs
+++ b/src/webapi/xml_http_request.rs
@@ -22,7 +22,7 @@ pub struct XmlHttpRequest( Reference );
 /// An enum indicating the state of the `XmlHttpRequest`.
 ///
 /// [(JavaScript docs)](https://developer.mozilla.org/en-US/docs/Web/API/XMLHttpRequest/readyState)
-// https://xhr.spec.whatwg.org/#dom-xmlhttprequest-readystate
+/// https://xhr.spec.whatwg.org/#dom-xmlhttprequest-readystate
 #[derive(Copy, Clone, PartialEq, Eq, Debug)]
 pub enum XhrReadyState {
     /// Client has been created. [open()](struct.XmlHttpRequest.html#method.open) not called yet.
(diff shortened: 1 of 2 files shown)


src/webapi/xml_http_request.rs
```diff
@@ -22,7 +22,7 @@ pub struct XmlHttpRequest( Reference );
 /// An enum indicating the state of the `XmlHttpRequest`.
 ///
 /// [(JavaScript docs)](https://developer.mozilla.org/en-US/docs/Web/API/XMLHttpRequest/readyState)
-// https://xhr.spec.whatwg.org/#dom-xmlhttprequest-readystate
+/// https://xhr.spec.whatwg.org/#dom-xmlhttprequest-readystate
```
koute: This should stay `//` (:

src/webapi/xml_http_request.rs
```diff
@@ -64,6 +82,23 @@ impl XmlHttpRequest {
         }
     }
 
+    /// Returns the type of the request as a [XhrResponseType](enum.XhrResponseType.html)
+    ///
+    /// [(JavaScript docs)](https://developer.mozilla.org/en-US/docs/Web/API/XMLHttpRequest/readyState)
+    // https://xhr.spec.whatwg.org/#dom-xmlhttprequest-readystate
```
koute: Wrong specs link.

src/webapi/xml_http_request.rs
```diff
@@ -78,6 +113,14 @@ impl XmlHttpRequest {
         }
     }
 
+    /// Returns the object representing the response
+    ///
+    ///[(JavaScript docs)](https://developer.mozilla.org/en-US/docs/Web/API/XMLHttpRequest/response)
+    // https://xhr.spec.whatwg.org/#ref-for-dom-xmlhttprequest-response
+    pub fn response(&self) -> Result< Value, TODO > {
```
koute: 1. `response` -> `raw_response` since it returns a raw `Value`; I'd like to reserve the `response` name for a more Rust-y method which returns an enum instead of a `Value`.

2. You return a `Result` here; can this actually throw an exception? [It doesn't look like it can throw to me from the specs](https://xhr.spec.whatwg.org/#dom-xmlhttprequest-response), so it should return `Option` instead.

src/webapi/xml_http_request.rs
```diff
@@ -64,6 +82,23 @@ impl XmlHttpRequest {
         }
     }
 
+    /// Returns the type of the request as a [XhrResponseType](enum.XhrResponseType.html)
+    ///
+    /// [(JavaScript docs)](https://developer.mozilla.org/en-US/docs/Web/API/XMLHttpRequest/readyState)
+    // https://xhr.spec.whatwg.org/#dom-xmlhttprequest-readystate
+    pub fn response_type(&self) -> XhrResponseType {
+        use self::XhrResponseType::*;
+        let type: String = js! ( return @{self}.responseType; ).try_into().unwrap();
```
koute: You can't name variables `type`. (It doesn't compile as `type` is a keyword.)

src/webapi/xml_http_request.rs
ryanisaacg: Question: Rather than Option, should it not just return a Value?

PR #243, opened 2018-06-28
 Add support for the wheel event

### LINKED ISSUE #230: Add wheel event
Hi! It seems it is not possible to add an handler for the mouse wheel. This could be useful, e.g., for zooming on a WebGL-based game.

ryanisaacg:
So the `wheel` event is a special event that inherits from `MouseEvent.` I have two main questions: 

- Should I impl the `wheel` properties directly on the event, or make a trait? AFAIK only one event is a "Wheel Event" so I don't think a trait is necessary. 

- Should the event be named `WheelEvent` or `MouseWheelEvent`? `WheelEvent` is more in keeping with the DOM api name, but `MouseWheelEvent` has greater consistency with the rest of the module.

Resolves #230

### DIFF
--- a/src/webapi/events/mouse.rs
+++ b/src/webapi/events/mouse.rs
@@ -447,13 +447,77 @@ impl ConcreteEvent for MouseLeaveEvent {
 #[derive(Clone, Debug, PartialEq, Eq, ReferenceType)]
 #[reference(instance_of = "MouseEvent")] // TODO: Better type check.
 #[reference(subclass_of(Event, UiEvent, MouseEvent))]
-pub struct WheelEvent( Reference );
+pub struct MouseWheelEvent( Reference );
 
-impl IEvent for WheelEvent {}
-impl IUiEvent for WheelEvent {}
-impl IMouseEvent for WheelEvent {}
-impl ConcreteEvent for WheelEvent {
+impl IEvent for MouseWheelEvent {}
+impl IUiEvent for MouseWheelEvent {}
+impl IMouseEvent for MouseWheelEvent {}
+impl ConcreteEvent for MouseWheelEvent {
     const EVENT_TYPE: &'static str = "mouseleave";
+}
+
+impl MouseWheelEvent {
+    /// The change in X of the wheel
+    ///
+    /// [(Javascript docs)](https://developer.mozilla.org/en-US/docs/Web/API/WheelEvent/deltaX)
+    // https://w3c.github.io/uievents/#dom-wheelevent-deltax
+    pub fn delta_x(&mut self) -> f64 {
+        js! (
+            @{&self.0}.deltaX
+        ).try_into().unwrap()
+    }
+
+    /// The change in Y of the wheel
+    ///
+    /// [(Javascript docs)](https://developer.mozilla.org/en-US/docs/Web/API/WheelEvent/deltaY)
+    // https://w3c.github.io/uievents/#dom-wheelevent-deltay
+    pub fn delta_y(&mut self) -> f64 {
+        js! (
+            @{&self.0}.deltaY
+        ).try_into().unwrap()
+    }
+
+    /// The change in Z of the wheel
+    ///
+    /// [(Javascript docs)](https://developer.mozilla.org/en-US/docs/Web/API/WheelEvent/deltaZ)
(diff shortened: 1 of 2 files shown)


koute:
If there's only one implementer of an interface then it's usually better to just make an inherent impl without adding any new traits.

Please name it `MouseWheelEvent` for consistency.

src/webapi/events/mouse.rs
```diff
@@ -439,6 +439,87 @@ impl ConcreteEvent for MouseLeaveEvent {
...
+#[derive(Clone, Debug, PartialEq, Eq, ReferenceType)]
+#[reference(instance_of = "MouseEvent")] // TODO: Better type check.
+#[reference(subclass_of(Event, UiEvent, MouseEvent))]
+pub struct MouseWheelEvent( Reference );
+
+impl IEvent for MouseWheelEvent {}
+impl IUiEvent for MouseWheelEvent {}
+impl IMouseEvent for MouseWheelEvent {}
+impl ConcreteEvent for MouseWheelEvent {
+    const EVENT_TYPE: &'static str = "mouseleave";
```
koute: Isn't thins wrong? (:

ryanisaacg: Whoops, the dangers of copy-paste I guess! Should be fixed with latest commit.

src/webapi/events/mouse.rs
```diff
@@ -439,6 +439,87 @@ impl ConcreteEvent for MouseLeaveEvent {
...
+pub enum MouseWheelDeltaMode {
+    /// The unit of measurement for the delta is pixels
+    // https://w3c.github.io/uievents/#dom-wheelevent-dom_delta_pixel
+    DomDeltaPixel = 0,
+     /// The unit of measurement for the delta is lines
+    // https://w3c.github.io/uievents/#dom-wheelevent-dom_delta_line
+    DomDeltaLine = 1,
+     /// The unit of measurement for the delta is pages
+    // https://w3c.github.io/uievents/#dom-wheelevent-dom_delta_page
+    DomDeltaPage = 2
```
koute: 1. Get rid of `= <number>` to make the enum opaque.
2. We can probably delete the `DomDelta` prefix.

src/webapi/events/mouse.rs
```diff
@@ -439,6 +439,87 @@ impl ConcreteEvent for MouseLeaveEvent {
...
+impl ConcreteEvent for MouseWheelEvent {
+    const EVENT_TYPE: &'static str = "wheel";
+}
+
+impl MouseWheelEvent {
+    /// The change in X of the wheel
+    ///
+    /// [(Javascript docs)](https://developer.mozilla.org/en-US/docs/Web/API/WheelEvent/deltaX)
+    // https://w3c.github.io/uievents/#dom-wheelevent-deltax
+    pub fn delta_x(&mut self) -> f64 {
```
koute: These should take `&self` instead of `&mut self`.

src/webapi/events/mouse.rs
```diff
@@ -439,6 +439,87 @@ impl ConcreteEvent for MouseLeaveEvent {
...
+
+/// What unit of measure the mouse wheel delta is in
+///
+/// [(JavaScipt docs)](https://developer.mozilla.org/en-US/docs/Web/API/WheelEvent/deltaMode)
+#[derive(Clone, Copy, Debug, PartialEq, Eq)]
+pub enum MouseWheelDeltaMode {
+    /// The unit of measurement for the delta is pixels
+    // https://w3c.github.io/uievents/#dom-wheelevent-dom_delta_pixel
+    DomDeltaPixel = 0,
+     /// The unit of measurement for the delta is lines
```
koute: Extra space. (:

src/webapi/events/mouse.rs
```diff
@@ -439,6 +439,87 @@ impl ConcreteEvent for MouseLeaveEvent {
     const EVENT_TYPE: &'static str = "mouseleave";
 }
 
+/// The `WheelEvent` is fired when a pointing device's wheel button (usually a mousewheel)
```
koute: `WheelEvent` -> `MouseWheelEvent`

ryanisaacg:
I should have noticed the issues before I made the PR. I think I fixed all of them, let me know if there are more.

koute:
Thanks!

PR #247, opened 2018-07-03
Implement the experimental pointer lock API

ryanisaacg:


src/webapi/events/pointer.rs
```diff
@@ -358,6 +358,40 @@ impl ConcreteEvent for LostPointerCaptureEvent {
...
+// http://www.w3.org/TR/pointerlock/#pointerlockchange-and-pointerlockerror-events
+#[derive(Clone, Debug, PartialEq, Eq, ReferenceType)]
+#[reference(instance_of = "Event")] // TODO: Better type check.
+#[reference(subclass_of(Event))]
+pub struct PointerLockChangeEvent( Reference );
+
+impl IEvent for PointerLockChangeEvent {}
+impl IUiEvent for PointerLockChangeEvent {}
+impl IMouseEvent for PointerLockChangeEvent {}
+impl IPointerEvent for PointerLockChangeEvent {}
```
koute: Hmm... I don't think these events are supposed to derive from these? (The specs are somewhat unclear regarding that.)

Taking a peek a Firefox sources [it looks like it should just be an `Event`](https://github.com/mozilla/spiderweb/blob/2eca983d249b0c7573534d59a008747073aa5a33/dom/events/EventNameList.h#L345)? Can you double check?

ryanisaacg: ...Huh. That's what I get for not checking my assumptions about web api's I guess. Yeah those derivations shouldn't be there.

koute:
Also it'd be nice if you could add a few really basic tests just as we have for other events. (:

ryanisaacg:
Sure, no problem.

koute:
You also should export those events in `lib.rs` or else they'll be useless. (:

ryanisaacg:
Whoops! My bad.

koute:
LGTM; thanks!

PR #255, opened 2018-07-30
Add `ScrollEvent`

hulthe:


src/webapi/events/dom.rs
```diff
@@ -100,6 +100,21 @@ impl ConcreteEvent for ResizeEvent {
     const EVENT_TYPE: &'static str = "resize";
 }
 
+/// The scroll event is fired when the document view or an element has been scrolled.
+///
+/// [(JavaScript docs)](https://developer.mozilla.org/en-US/docs/Web/Events/scroll)
+// https://drafts.csswg.org/cssom-view/#eventdef-document-scroll
+#[derive(Clone, Debug, PartialEq, Eq, ReferenceType)]
+#[reference(instance_of = "UIEvent")] // TODO: Better type check.
```
koute: Doesn't it say in the specs that this uses the `Event` interface?

koute: If they're only sometimes `UIEvent`s then it'd be better to make this one derive from `Event`. (It'd be nice to have concrete examples when it's an `UIEvent` and when it's only `Event`.) If need be we can always add `UiScrollEvent` later and have that derive from `UiEvent`.

hulthe: Alright, its done

koute:
LGTM; thanks!

PR #266, opened 2018-08-25
Allow whitespace around element for Node::from_html().

kevincox:
Often it is easier to call this function with a string which has
whitespace around the outside.

### DIFF
--- a/src/webapi/node.rs
+++ b/src/webapi/node.rs
@@ -875,9 +875,7 @@ mod tests {
 
     #[test]
     fn from_html_trimmed() {
-        let node = Node::from_html("
-          <div>Some text, horray!</div>
-        ").unwrap();
+        let node = Node::from_html("\n<div>Some text, horray!</div>\n").unwrap();
         let text = node.first_child().unwrap();
 
         assert_eq!(node.node_name(), "DIV");
--- a/src/webapi/node.rs
+++ b/src/webapi/node.rs
@@ -383,13 +383,27 @@ impl Node {
         js_try!(
             var span = document.createElement("span");
             span.innerHTML = @{html};
-            if( span.childNodes.length != 1 ) {
+
+            // Fast path, if we have one childNode just return it.
+            if( span.childNodes.length > 1 ) {
+                // We will drop one leading and training whitespace-only text node.
+                if( span.firstChild.nodeType == Node.TEXT_NODE &&
+                    !span.firstChild.textContent.trim() ) {
+                    span.removeChild(span.firstChild);
+                }
+                if( span.lastChild.nodeType == Node.TEXT_NODE &&
+                    !span.lastChild.textContent.trim() ) {
+                    span.removeChild(span.lastChild);
+                }
+            }
+
+            if( span.childNodes.length != 1) {
                 throw new DOMException(
                     "Node::from_html requires a single root node but has: "
                     + span.childNodes.length,
                     "SyntaxError");
             }
-            return span.childNodes[0];
+            return span.firstChild;
         ).unwrap()
     }
 }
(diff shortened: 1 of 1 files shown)


src/webapi/node.rs
```diff
@@ -383,13 +383,27 @@ impl Node {
...
+                    !span.firstChild.textContent.trim() ) {
+                    span.removeChild(span.firstChild);
+                }
+                if( span.lastChild.nodeType == Node.TEXT_NODE &&
+                    !span.lastChild.textContent.trim() ) {
+                    span.removeChild(span.lastChild);
+                }
+            }
+
+            if( span.childNodes.length != 1) {
```
koute: Hm... all of this seems unnecessary. I mean - wouldn't replacing `@{html}` with `@{html.trim()}` achieve the same thing? (:

src/webapi/node.rs
```diff
@@ -858,4 +872,28 @@ mod tests {
         assert!(format!("{}", err).contains("requires a single root node"));
         assert!(Node::from_html("<di").is_err());
     }
+
+    #[test]
+    fn from_html_trimmed() {
+        let node = Node::from_html("
+          <div>Some text, horray!</div>
```
koute: Please make the newlines explicit in these tests. (Replace actual newlines with `\n`.)

src/webapi/node.rs
kevincox: I originally used `.trim()` in rust but this changes the behavior for people generating text nodes. For example

```rust
stdweb::web::Node::from_html(r#"
  foobar
"#);
```

Currently this has a text node surrounded by whitespace but if we used trim it would just be the word. The checks here ensure that I am only changing behavior which was previously an error.

src/webapi/node.rs
kevincox: Done.

src/webapi/node.rs
koute: Hmm.. in which case can you add a test for the text-surrounded-by-whitespace case?

I think we still may be able to do this on Rust's side (which should be faster) - find first non-whitespace byte, if it's `b'<'` then it's not a text node and should be trimmed (and since we've found the first non-whitespace byte we already have it trimmed from the left side), otherwise do not touch it. Also, Pauan is right in that Rust's `trim` is Unicode-aware, which we don't really need here; only `b'\n'`, `b'\r'`, `b'\t'` and `b' '` should suffice.

kevincox: @koute 

I'll try to implement this rust side.

> Rust's trim is Unicode-aware

Is this a problem? Should I use a trim which explicitly isn't unicode aware?

@Pauan I could be convinced not to however there are two reasons I didn't do this.

1. This is a change in behaviour to existing users.
1. Sometimes you do need to have space around text as it will affect the DOM and rendering.

If we are okay with these points adding `.trim()` is trivial.

koute: > Is this a problem? Should I use a trim which explicitly isn't unicode aware?

It's not strictly a problem, but it's totally unnecessary.

I'm generally pretty ambivalent as to which behavior we pick here, however since this method's *current* behavior is that it leaves text nodes as-is we shouldn't change that without a major version bump, so I'm okay with trimming only actual non-text nodes here provided we'll do it on Rust's side with a custom non-Unicode trimmer (which should be fairly easy to write).

Pauan: @kevincox Just to be clear, I never meant for stdweb to automatically call `trim`, instead I meant that the *user* of stdweb would call `trim`:

```rust
Node::from_html(foo.trim())
```

This allows the user to call `trim` only when it makes sense to do so.

kevincox:
Since there is no obviously correct solution and support for `insertAdjacentHTML` solves my use case I am going to abandon this. If anyone wants to pick it up feel free.

PR #267, opened 2018-08-25
Add insertAdjacentHTML binding and helpers.

kevincox:
This function provides an efficient way to add nodes to the DOM. This
provides a binding to the raw function as well as helper functions which
make it easier to provide a correct `position` argument.

### DIFF
--- a/src/webapi/dom_exception.rs
+++ b/src/webapi/dom_exception.rs
@@ -44,6 +44,17 @@ impl IDomException for InvalidAccessError {}
 
 error_boilerplate! { InvalidAccessError, name = "InvalidAccessError" }
 
+/// Occurs when the object can not be modified.
+// NoModificationAllowedError
+#[derive(Clone, Debug, PartialEq, Eq, ReferenceType)]
+#[reference(subclass_of(Error, DomException))]
+pub struct NoModificationAllowedError( Reference );
+
+impl IError for NoModificationAllowedError {}
+impl IDomException for NoModificationAllowedError {}
+
+error_boilerplate! { NoModificationAllowedError, name = "NoModificationAllowedError" }
+
 /// Occurs when the specified object cannot be found.
 // https://heycam.github.io/webidl/#notfounderror
 #[derive(Clone, Debug, PartialEq, Eq, ReferenceType)]
--- a/src/webapi/dom_exception.rs
+++ b/src/webapi/dom_exception.rs
@@ -45,7 +45,7 @@ impl IDomException for InvalidAccessError {}
 error_boilerplate! { InvalidAccessError, name = "InvalidAccessError" }
 
 /// Occurs when the object can not be modified.
-// NoModificationAllowedError
+// https://heycam.github.io/webidl/#nomodificationallowederror
 #[derive(Clone, Debug, PartialEq, Eq, ReferenceType)]
 #[reference(subclass_of(Error, DomException))]
 pub struct NoModificationAllowedError( Reference );
--- a/src/webapi/element.rs
+++ b/src/webapi/element.rs
@@ -168,10 +168,10 @@ pub trait IElement: INode + IParentNode + IChildNode {
     /// Insert nodes from HTML fragment into specified position.
     ///
     /// [(JavaScript docs)](https://developer.mozilla.org/en-US/docs/Web/API/Element/insertAdjacentHTML)
-    // https://dom.spec.whatwg.org/#dom-element-insertadjacentelement
+    // https://dom.spec.whatwg.org/#ref-for-dom-element-insertadjacentelement
     fn insert_adjacent_html( &self, position: &InsertPosition, html: &str ) -> Result<(), InsertAdjacentError> {
         js_try!( @(no_return)
-            @{self.as_ref()}.insertAdjacentHTML( @{position.keyword()}, @{html} );
+            @{self.as_ref()}.insertAdjacentHTML( @{position.as_str()}, @{html} );
         ).unwrap()
     }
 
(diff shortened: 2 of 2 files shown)


src/webapi/element.rs
```diff
@@ -163,8 +163,46 @@ pub trait IElement: INode + IParentNode + IChildNode {
     fn has_pointer_capture( &self, pointer_id: i32 ) -> bool {
         js!( return @{self.as_ref()}.hasPointerCapture( @{pointer_id} ); ).try_into().unwrap()
     }
+
+    /// Insert nodes from HTML fragment into specified position.
+    ///
+    /// [(JavaScript docs)](https://developer.mozilla.org/en-US/docs/Web/API/Element/insertAdjacentHTML)
+    fn insert_adjacent_html( &self, position: &str, html: &str ) -> Result<(), SyntaxError> {
```
koute: 1. The `position` parameter should be an enum.
2. This can also throw `NoModificationAllowedError` besides `SyntaxError`. (See `WebSocket::new` for an example on how to handle this. We don't have bindings for `NoModificationAllowedError` so you have to add that too.)
3. Please add a link to the specs' IDL just as we have for the most of other methods.

src/webapi/element.rs
```diff
@@ -163,8 +163,46 @@ pub trait IElement: INode + IParentNode + IChildNode {
...
+    ///
+    /// [(JavaScript docs)](https://developer.mozilla.org/en-US/docs/Web/API/Element/insertAdjacentHTML)
+    fn insert_html_before_end( &self, html: &str ) -> Result<(), SyntaxError> {
+        self.insert_adjacent_html("beforeend", html)
+    }
+
+    /// Insert nodes from HTML fragment after element.
+    ///
+    /// [(JavaScript docs)](https://developer.mozilla.org/en-US/docs/Web/API/Element/insertAdjacentHTML)
+    fn insert_html_after_end( &self, html: &str ) -> Result<(), SyntaxError> {
```
koute: Hmmm.. are those necessary?

I guess I'm not entirely against having them for conveniance, however if we're going to invent new methods I'd rather have them named better than the original JS one. (e.g. maybe `append_html`, `prepend_html`, `append_html_outside`, `prepend_html_outside`, or something like that?)

src/webapi/element.rs
kevincox: 1. Done
2. Done.
3. Done.

src/webapi/element.rs
kevincox: They aren't very necessary but they save a lot of typing and show up in autocomplete lists which I find very useful. I would prefer to have them.

As for the names I'm not the biggest fan either. I like `prepend_html` and `append_html`. However I don't really like `append_html_outside` and `append_html_outside`. The words append/prepend don't seem to fit when the content is likely going somewhere in the middle of the parent element.

koute: > However I don't really like `append_html_outside` and `prepend_html_outside`. The words append/prepend don't seem to fit when the content is likely going somewhere in the middle of the parent element.

Hmmm... then how about `insert_html_before` and `insert_html_after`?

kevincox: I like those. I will update to `prepend_html`, `append_html`, `insert_html_before` and `insert_html_after`.

src/webapi/element.rs
```diff
@@ -163,8 +164,47 @@ pub trait IElement: INode + IParentNode + IChildNode {
     fn has_pointer_capture( &self, pointer_id: i32 ) -> bool {
         js!( return @{self.as_ref()}.hasPointerCapture( @{pointer_id} ); ).try_into().unwrap()
     }
+
+    /// Insert nodes from HTML fragment into specified position.
+    ///
+    /// [(JavaScript docs)](https://developer.mozilla.org/en-US/docs/Web/API/Element/insertAdjacentHTML)
+    // https://dom.spec.whatwg.org/#dom-element-insertadjacentelement
```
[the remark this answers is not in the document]
kevincox: Fixed.

src/webapi/dom_exception.rs
```diff
@@ -44,6 +44,17 @@ impl IDomException for InvalidAccessError {}
 
 error_boilerplate! { InvalidAccessError, name = "InvalidAccessError" }
 
+/// Occurs when the object can not be modified.
+// NoModificationAllowedError
```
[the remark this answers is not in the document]
kevincox: Added.

src/webapi/element.rs
```diff
@@ -181,3 +221,73 @@ impl IElement for Element {}
...
+impl InsertPosition {
+	fn keyword(&self) -> &str {
+		match *self {
+			InsertPosition::BeforeBegin => "beforebegin",
+			InsertPosition::AfterBegin => "afterbegin",
+			InsertPosition::BeforeEnd => "beforeend",
+			InsertPosition::AfterEnd => "afterend",
+		}
+	}
+}
```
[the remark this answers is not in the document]
kevincox: Done.

src/webapi/element.rs
```diff
@@ -181,3 +221,73 @@ impl IElement for Element {}
...
+
+/// Errors thrown by `Element::insert_adjacent_html`.
+error_enum_boilerplate! {
+    InsertAdjacentError,
+    NoModificationAllowedError, SyntaxError
+}
+
+impl InsertPosition {
+	fn keyword(&self) -> &str {
```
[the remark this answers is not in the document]
kevincox: Done.

src/webapi/element.rs
koute: Sounds good to me!

src/webapi/element.rs
```diff
@@ -163,8 +164,47 @@ pub trait IElement: INode + IParentNode + IChildNode {
     fn has_pointer_capture( &self, pointer_id: i32 ) -> bool {
         js!( return @{self.as_ref()}.hasPointerCapture( @{pointer_id} ); ).try_into().unwrap()
     }
+
+    /// Insert nodes from HTML fragment into specified position.
+    ///
+    /// [(JavaScript docs)](https://developer.mozilla.org/en-US/docs/Web/API/Element/insertAdjacentHTML)
+    // https://dom.spec.whatwg.org/#ref-for-dom-element-insertadjacentelement
+    fn insert_adjacent_html( &self, position: &InsertPosition, html: &str ) -> Result<(), InsertAdjacentError> {
```
koute: Since the enum is `Copy` this should take `InsertPosition`. (In general the `Copy` types should be passed by value while `Clone` types should be passed by reference.)

kevincox: Good point, fixed.

src/webapi/element.rs
koute: Ping? (:

kevincox: Done.

koute:
LGTM! Thanks!

PR #271, opened 2018-08-31
Add createElementNS support

siler:
Add bindings for createElementNS. This is necessary if we would like to use elements other than those in the default namespace, such as the SVG elements.

### DIFF
--- a/src/webapi/document.rs
+++ b/src/webapi/document.rs
@@ -22,11 +22,6 @@ use webapi::dom_exception::{InvalidCharacterError, NamespaceError};
 pub struct Document( Reference );
 
 error_enum_boilerplate! {
-    CreateElementError,
-    InvalidCharacterError
-}
-
-error_enum_boilerplate! {
     CreateElementNsError,
     InvalidCharacterError,
     NamespaceError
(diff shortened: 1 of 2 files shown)


src/webapi/document.rs
```diff
@@ -58,6 +58,19 @@ impl Document {
         }
     }
 
+    /// Creates an element with the specified namespace URI and qualified name.
+    /// To create an element without specifying a namespace URI, use the `createElement` method.
+    ///
+    /// [(JavaScript docs)](https://developer.mozilla.org/en-US/docs/Web/API/Document/createElementNS)
+    // https://dom.spec.whatwg.org/#ref-for-dom-document-createelementns
+    pub fn create_element_ns( &self, namespace_uri: &str, tag: &str ) -> Result< Element, TODO > {
```
koute: Thanks for the PR!

Since you're adding this - could you replace the `TODO` with an actual error type? (The `TODO`s were added in batch to mark a bunch of already existing methods as throwing, but we want to get rid of them eventually.)

I can see from the specs that this can throw at least `NotSupportedError`, `NamespaceError` and `InvalidCharacterError`. So:
1) please create an aggregate error with the `error_enum_boilerplate!`,
2) if possible (and it should be possible from what I can see) please add tests for these errors (the specs say in what situations they get thrown; just pick one per error to cover every possible variant),

siler: I didn't see a NotSupportedError for this file. Also added tests for createElement.

src/webapi/document.rs
```diff
@@ -21,6 +21,17 @@ use private::TODO;
 #[reference(subclass_of(EventTarget, Node))]
 pub struct Document( Reference );
 
+error_enum_boilerplate! {
+    CreateElementError,
+    InvalidCharacterError
+}
```
koute: So I guess these can't actually throw `NotSupportedError`? (Reading the specs more closely it seems like it might be the case.) If so then we don't need this enum - we can just return `InvalidCharacterError` from `create_element` directly. (In general those enums are when we have 2 or more possible exceptions.)

siler: Updated!

koute:
LGTM; thanks!

PR #272, opened 2018-09-04
Added a conversion from Number to f64.

Migi:
The existing conversions from Number to Rust primitives all used `TryFrom`, but to `f64` a direct conversion (with `From` or `Into`) is possible and preferred.


koute:
LGTM! Thanks!

Issue #273, opened 2018-09-06
Mutable closures (FnMut) in js! macros are unsound


Migi:
The stdweb runtime handles regular closures (that only implement `Fn`) and mutable closures (that implement `FnMut`) the same way.  The wrong way, for `FnMut`. It doesn't check that you're not calling a `FnMut` closure while a call to that same `FnMut` closure is already going on. Essentially, it allows multiple mutable references to the state of the closure (and all variables in it) to exist at the same time.

Here's an example demonstrating the bug:
```rust
#[macro_use]
extern crate stdweb;

use std::sync::atomic::{fence, Ordering};

fn main() {
	// this int gets moved into the closure f below
	let mut int = 0;

	let f = move |operation: String, callback: ::stdweb::Value| {
		if operation == "print_and_call_back" {
			// This &mut i32 is valid for the whole scope, nothing re-borrows it.
			// If we don't modify the i32 it points to, its value should
			// stay the same as long as we borrow it.
			let mut_ref : &mut i32 = &mut int;
			
			console!(log, "Before callback: ", *mut_ref); // prints 0 on my machine
			
			// Without this line, the compiler actually relies on this fact that
			// the i32 pointed to shouldn't be able to change
			// and it prints 0 both times.
			// With this line, it could still do that (it could do whatever it wants,
			// it's undefined behavior after all), but at least on my machine
			// it makes it print 1 after the callback.
			fence(Ordering::SeqCst);

			// callback is just a Value, and doesn't borrow anything in this scope.
			js! {
				@{callback}();
			}
			
			console!(log, "After callback: ", *mut_ref); // prints 1 on my machine
		} else if operation == "increment" {
			int += 1;
		}
	};
	
	js! {
		let f = @{f};
		
		f("print_and_call_back", function () {
			f("increment", null);
		});
	
		f.drop();
	}
}
```

Since JavaScript is mostly single-threaded and you can only use closures that are valid for the `'static` lifetime inside the `js!` macro, this is unlikely to cause any real-world bugs, but still it's unsound and should be fixed.

If I'm not mistaken, the `js!` macro can't actually distinguish between `Fn`s and `FnMut`s right now, at least not without a modifier like `Once`. And I don't think we want to prohibit all closures from being called recursively like in the code above (for `Fn`s, such recursion is fine). So one possible solution would be to make a `Mut` modifier similar to the `Once` one, and if you want to use a `FnMut` in a `js!` macro you need to wrap it in that `Mut` modifier.

[...]

PR #274, opened 2018-09-06
Basic fix for #273, simply forbidding `FnMut`s in `js!`

Migi:
This change simply forbids `FnMut`s from being used in `js!` macros, without implementing any `Mut` modifier or any other solution that would allow `FnMut`s to be used in the `js!` macro safely.

**This change is NOT backwards compatible**, but any fixes to #273 will have to break backwards compatibility in some way.

Most notably, MutationObserver::new and IEventTarget::add_event_listener accepted FnMut+'static closures as arguments. I changed these to accept only Fn+'static closures, as handling FnMuts correctly incurs unnecessary overhead for very little benefit (due to the 'static bound), as discussed in #273.

I'll wait on your opinion on whether you think a Mut modifier is worth it or not (and what should happen if illegal recursion is detected, a JavaScript exception or a Rust error) before trying to implement that.


Migi:
This PR was superseded by #279.

Issue #278, opened 2018-09-10
Closures can be dropped while they are being called


Migi:
This issue is related to #273 and #277.

The runtime currently doesn't check that when you call `.drop()` in JavaScript on a Rust closure, that a call to that closure isn't actually going on at that very moment. This is possible through callbacks.

Example code:

```rust
#[macro_use]
extern crate stdweb;

use std::sync::atomic::{fence, Ordering};

struct MyStruct {
	was_dropped: bool
}

impl Drop for MyStruct {
	fn drop(&mut self) {
		self.was_dropped = true;
	}
}

fn main() {
	let s = MyStruct { was_dropped: false };

	let f = move |callback: ::stdweb::Value| {
		js! {
			@{callback}();
		}
		fence(Ordering::SeqCst);
		console!(log, "s.was_dropped: ", s.was_dropped); // prints true
	};
	
	js! {
		let f = @{f};

		f(function() {
			f.drop();
		});
	}
}
```

I'm also working on a fix for this.

Issue #277, opened 2018-09-10
The runtime checks that prevent a closure from being dropped twice are not very thorough


Migi:
If you use a closure in a `js!` macro, then you have to call `.drop()` manually or else you're leaking memory. There is a check to make sure that if you call `f.drop()` twice, the Rust closure is only dropped once, but if you call `let drop = f.drop;` first, then calling `drop()` twice actually does drop the Rust closure twice.

Example:

```rust
#[macro_use]
extern crate stdweb;

struct MyStruct {
	val: i32
}

impl Drop for MyStruct {
	fn drop(&mut self) {
		console!(log, "Dropping");
	}
}

fn main() {
	let s = MyStruct { val: 7 };

	let f = move || {
		console!(log, s.val);
	};
	
	js! {
		let f = @{f};

		// this works fine (only drops MyStruct once):
		//f.drop();
		//f.drop();

		// but this does not:
		let drop = f.drop;
		drop();
		drop();
	}
}
```

I'm working on a fix for this.

PR #279, opened 2018-09-10
Add Mut modifier

Migi:
This PR supersedes PR #274, and fixes issues #273, #277 and #278.

This PR implements the `Mut` modifier discussed in issue #273. The `js!` macro now only accepts `Fn` closures, and to use `FnMut` closures you need to wrap them in `Mut(...)`, very similar to the situation with `FnOnce` and `Once(...)`.

Interestingly, a closure with a `Mut` modifier is not slower than a regular closure, because even regular closures need to keep track of ongoing calls due to issue #278.

When the runtime detects any illegal recursion or premature drops, a `ReferenceError ` is thrown to the JavaScript world (as it already did when a dropped closure was called).

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -182,6 +182,7 @@ pub use webcore::array::Array;
 pub use webcore::symbol::Symbol;
 
 pub use webcore::unsafe_typed_array::UnsafeTypedArray;
+pub use webcore::mutfn::Mut;
 pub use webcore::once::Once;
 pub use webcore::instance_of::InstanceOf;
 pub use webcore::reference_type::ReferenceType;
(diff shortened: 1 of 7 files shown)


koute:
This look great! Thanks a lot!

I'd like to request one extra thing though - could you split this into two PRs - one disallowing `FnMut`s in `js!` and one having everything else?

I'd like to merge the most of this right away, however as disallowing `FnMut`s is a breaking change it'd need to go into `0.5`. I still want to release a version or two in the `0.4` series, and then we can make the `js!` more strict and bump it to `0.5`.

Migi:
Alright, I've added a commit that allows `FnMut` again in the `js!` macro. That commit can simply be reverted for version 0.5.

koute:
Awesome! Thank you!

PR #284, opened 2018-09-21
Add `IElement::namespace_uri`

sapir:


src/webapi/element.rs
```diff
@@ -15,6 +15,17 @@ use webcore::try_from::TryFrom;
...
 pub trait IElement: INode + IParentNode + IChildNode {
+    /// The Element.namespaceURI read-only property returns the namespace URI
+    /// of the element, or null if the element is not in a namespace.
+    ///
+    /// [(JavaScript docs)](https://developer.mozilla.org/en-US/docs/Web/API/Element/namespaceURI)
+    // https://dom.spec.whatwg.org/#ref-for-dom-element-namespaceuri
+    fn namespace_uri( &self ) -> Option< String > {
+        js!(
+            return @{self.as_ref()}.namespaceURI;
+        ).try_into().ok()
```
koute: The `ok` here should be an `unwrap` - that is, we want to enforce that the snippet will either return nothing or a `String`, and have it panic otherwise.

With an `ok` *if* this snippet would return something other than a `String` we would silently return `None` here; by using an `unwrap` we assert that such a situation should *never* happen.

koute:
LGTM! Thanks!

PR #285, opened 2018-09-26
[WIP] js! proc macro

ForsakenHarmony:
This is just the first version, looking for feedback

included 2 examples which should obviously be removed later

in
```rs
js_new!(
    return new UIEvent(
        @{ClickEvent::EVENT_TYPE},
        {
            detail: 1,
        }
    )
)

js_new! (
    return @{self}.width;
)
```

out
```rs

__js_raw_asm ! ( "return new UIEvent(($1),{detail:1,})" , { ClickEvent :: EVENT_TYPE } )

__js_raw_asm ! ( "return($1).width;" , { self } )
```


koute:
Thanks!

So far this looks simpler than I though it will. (:

A few comments:
 - The macro should be in the `stdweb-internal-macros` crate instead of in `stdweb-derive`, and should be reexported like the `js_export` and `async_test` macros.
  - Can you add a test which check that the stringification (so only the first argument of the `__js_raw_asm!`) works as it currently does for the `js!` macro? (There are a few corner cases with Rust's tokenizations which produce invalid JS when translated naively, e.g. `++` which will be tokenized as `+ +` with a naive `macro_rules!` macro.) You can just copy-paste the `tests::stringify` test from `src/webcore/macros.rs`.
  - We should add a new feature to `stdweb`'s `Cargo.toml` (e.g. "procedural-js-macro" or something like that, doesn't matter what it's named) which would turn make the old `js!` macro proxy into this new one - basically gate the current `js!` with a `#[cfg(not(feature = "procedural-js-macro"))]`, then add another `macro_rules! js` with `#[cfg(feature = "procedural-js-macro")]` and just pass all of the tokens into our new macro. (Or I guess maybe we don't need the passthrough `js!` macro? Do procedural macros automatically appear when you `#[macro_use]` the create containing them? I don't know.)

ForsakenHarmony:
1 and 2 done

can do 3 when the new version works

the old `js!` macro does more than just the `__js_raw_asm!` part, it also adds the result around and some other stuff, what would be the next thing to add?

koute:
We're finally switching to a procedural macro based `js!` here: https://github.com/koute/stdweb/pull/306

@ForsakenHarmony I've based the code on that PR on yours, so thanks a lot for this!

ForsakenHarmony:
Sorry, I should have continued here

But in haply if you could use it in some way

koute:
It's no problem! Thank again. (:

PR #289, opened 2018-10-05
Add a macro for creating js translation functions

shelvacu:
I noticed there's a lot of repeated code for javascript FFI. This macro means that instead of writing

```rust
/// Docs docs docs
fn focus( &self ) {
    js! { @(no_return)
        @{self.as_ref()}.focus()
    }
}

/// Other docs
fn dataset( &self ) -> StringMap {
    unsafe {
        js!(
            return @{self.as_ref()}.dataset;
        ).into_reference_unchecked().unwrap()
    }
}

/// More docs
fn get_bounding_client_rect( &self ) -> Rect {
    js! (
        return @{self.as_ref()}.getBoundingClientRect();
    ).try_into().unwrap()
}

```

you can just write

```rust
js_translate_function!{
    /// Docs docs docs
    focus:focus() => noreturn:()
}

js_translate_function!{
    /// Other docs
    dataset:dataset none => reference:StringMap
}

js_translate_function!{
    /// More docs
    get_bounding_client_rect:getBoundingClientRect() => into:Rect
}
```

### DIFF
--- a/src/webcore/macros.rs
+++ b/src/webcore/macros.rs
@@ -753,6 +753,66 @@ macro_rules! error_enum_boilerplate {
     }
 }
 
+macro_rules! js_translate_function {
+    ( $(#[$attr:meta])* $rust_name:ident : $js_name:ident none => $translation:ident : $return:ty ) => {
+        $( #[ $attr ] )*
+        fn $rust_name( &self ) -> $return {
+            __js_translate_function_inner!{ self, $translation, $js_name none }
+        }
+    };
+    (
+        $(#[$attr:meta])*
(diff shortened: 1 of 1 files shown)


koute:
Sorry for the delay. Thanks for the PR!

Hmm.... I'd definitely like something like this, but I'm not sure if we want it in the current form. In general I'd like to eventually make `stdweb` compatible with the `wasm-bindgen`-based ecosystem, so that would mean that at very least we'd probably want a procedural macro which has exactly the same syntax as `wasm-bindgen`'s, but instead of generating `wasm-bindgen`-compatible code would generate `cargo-web`-compatible bindings.

PR #298, opened 2018-11-04
query_selector and query_selector_all error type

o0Ignition0o:
Return a SyntaxError instead of a TODO for ParentNode query_selector and query_selector_all, as described [in the MDN docs](https://developer.mozilla.org/en-US/docs/Web/API/Element/querySelector#Exceptions)

### DIFF
--- a/src/webapi/parent_node.rs
+++ b/src/webapi/parent_node.rs
@@ -1,7 +1,7 @@
 use webcore::reference_type::ReferenceType;
 use webapi::node_list::NodeList;
 use webapi::element::Element;
-use private::TODO;
+use webapi::dom_exception::SyntaxError;
 
 /// The `ParentNode` mixin contains methods and properties
 /// that are common to all types of `Node` objects that can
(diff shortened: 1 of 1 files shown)


koute:
Thanks for the PR! (:

Since we're at it could we perhaps add a test? This should be very easy to test for as it's just a `SyntaxError`. You can look at e.g. `src/webapi/element.rs` on how to add one and just do a similar thing in this file.

You can run all of the tests locally like this: `cargo web test --features web_test` (without the `--features web_test` it won't run the DOM related tests)

o0Ignition0o:
Sure, I'll try to add one, thanks for the way you explain how to proceed by the way! :)

o0Ignition0o:
Here you go :)

Please let me know if I can improve the tests or if you would rather avoid using a macro during the tests for compile time / readability reasons :)

koute:
Yeah, the macro should be unnecessary (just out of curiosity - why did you decide to use it anyway?); you should be able to just turn it into a function. (: Or, actually, you don't have to define anything - you can just use [`Document::create_element`](https://docs.rs/stdweb/*/stdweb/web/struct.Document.html#method.create_element), e.g.: `document().create_element( "foobar" ).unwrap()`, (You can import ['document()` from here](https://docs.rs/stdweb/*/stdweb/web/fn.document.html).)

Besides this looks good to me!

o0Ignition0o:
I think I looked at the tests in example.rs a bit too much :p
I saw a div() function and an h1() function and I didn't want to create a p function etc... I didn't know about the create_element function though (Haven't taken a proper look at the crate yet ^^)

I ll switch to a regular method !

o0Ignition0o:
Ok this looks better now :)

PR #299, opened 2018-11-07
Custom Elements: CustomElementRegistry

prasannavl:
[WIP] Initial impl of `CustomElementRegistry`

I do have a few questions before completing this:

1. How exactly do I represent a Constructor here? (https://github.com/koute/stdweb/compare/master...prasannavl:patch-1#diff-55e0c9ca42d6126fc52bcae5a9a2bf95R38)
1. Is it okay to unwrap, this? Initially I did a `no_return`, but then added the error type, and as such the `Result<(), Err>`. Does this fit the project conventions? (https://github.com/koute/stdweb/compare/master...prasannavl:patch-1#diff-55e0c9ca42d6126fc52bcae5a9a2bf95R41)
1. Again, constructor type. But also, can `Option` be returned. Does this fit the project conventions, or should this be a result with empty error/not found error? What's the best way to represent `undefined` here. (https://github.com/koute/stdweb/compare/master...prasannavl:patch-1#diff-55e0c9ca42d6126fc52bcae5a9a2bf95R48)
1. I followed this convention by looking at other implementations, but it is correct to unwrap this? (https://github.com/koute/stdweb/compare/master...prasannavl:patch-1#diff-55e0c9ca42d6126fc52bcae5a9a2bf95R48)

### DIFF
--- a/src/webapi/custom_elements.rs
+++ b/src/webapi/custom_elements.rs
@@ -1,0 +1,62 @@
+use webcore::value::Reference;
+webapi::error::TypeError;
+
+use webapi::dom_exception::{
+    NotSupportedError,
+    SyntaxError
+};
+
+#[cfg(feature = "experimental_features_which_may_break_on_minor_version_bumps")]
+use webcore::promise::{Promise, TypedPromise};
+
+/// The CustomElementRegistry interface provides methods for registering custom elements 
+/// and querying registered elements. To get an instance of it, use the window.customElements property. 
+///
+/// [(JavaScript docs)](hhttps://developer.mozilla.org/en-US/docs/Web/API/CustomElementRegistry)
+/// https://html.spec.whatwg.org/multipage/custom-elements.html#customelementregistry
+#[derive(Clone, Debug, PartialEq, Eq, ReferenceType)]
+#[reference(instance_of = "CustomElementRegistry")]
+pub struct CustomElementRegistry(Reference);
+
+
+error_enum_boilerplate! {
+    /// A enum of the exceptions that CustomElementRegistry.define() may throw
+    DefineError,
+    /// A TypeError
+    TypeError
+    /// A NotSupportedError
+    NotSupportedError,
+    /// A SyntaxError
+    SyntaxError,
+}
+
+impl CustomElementRegistry {
+    /// Defines a new custom element
+    ///
+    /// [(JavaScript docs)](https://developer.mozilla.org/en-US/docs/Web/API/CustomElementRegistry/define)
+    /// https://html.spec.whatwg.org/multipage/custom-elements.html#dom-customelementregistry-define
+    pub fn define( &self, name: &str, constructor: Any) -> Result<(), DefineError> {
+        js!(
+            return @{self}.define(name, constructor); 
+        ).try_into().unwrap()
+    }
+
+    /// Returns the constuctor for the named custom element, or undefined if the custom element is not defined.
+    ///
+    /// [(JavaScript docs)](https://developer.mozilla.org/en-US/docs/Web/API/CustomElementRegistry/get)
+    /// https://html.spec.whatwg.org/multipage/custom-elements.html#dom-customelementregistry-get
+    pub fn get( &self, name: &str ) -> Option<Constructor> {
+        js!( return @{self}.get(name); ).try_into().unwrap()
+    }
+
+    /// Returns a promise that will be fulfilled when a custom element becomes defined with the given name.
+    /// (If such a custom element is already defined, the returned promise is immediately fulfilled.)
+    ///
+    /// [(JavaScript docs)](https://developer.mozilla.org/en-US/docs/Web/API/CustomElementRegistry/whenDefined)
+    /// https://html.spec.whatwg.org/multipage/custom-elements.html#dom-customelementregistry-whendefined
+    #[cfg(feature = "experimental_features_which_may_break_on_minor_version_bumps")]
+    pub fn whenDefined( &self, name: &str ) -> TypedPromise<(), SyntaxError > {
+        let promise: Promise = js!( return @{self}.whenDefined(name); ).try_into().unwrap();
+        TypedPromise::new(promise)
+    })
+}


src/webapi/custom_elements.rs
```diff
@@ -0,0 +1,62 @@
...
+};
+
+#[cfg(feature = "experimental_features_which_may_break_on_minor_version_bumps")]
+use webcore::promise::{Promise, TypedPromise};
+
+/// The CustomElementRegistry interface provides methods for registering custom elements 
+/// and querying registered elements. To get an instance of it, use the window.customElements property. 
+///
+/// [(JavaScript docs)](hhttps://developer.mozilla.org/en-US/docs/Web/API/CustomElementRegistry)
+/// https://html.spec.whatwg.org/multipage/custom-elements.html#customelementregistry
```
koute: Links to the specs should be in a normal comment, that is `//` instead of `///`. (:

src/webapi/custom_elements.rs
```diff
@@ -0,0 +1,62 @@
...
+#[derive(Clone, Debug, PartialEq, Eq, ReferenceType)]
+#[reference(instance_of = "CustomElementRegistry")]
+pub struct CustomElementRegistry(Reference);
+
+
+error_enum_boilerplate! {
+    /// A enum of the exceptions that CustomElementRegistry.define() may throw
+    DefineError,
```
koute: A little more verbose name would be better here, I think. Maybe `CustomElementDefineError`? It's a little long, but simply `DefineError` seems to generic.

src/webapi/custom_elements.rs
```diff
@@ -0,0 +1,62 @@
...
+    /// A SyntaxError
+    SyntaxError,
+}
+
+impl CustomElementRegistry {
+    /// Defines a new custom element
+    ///
+    /// [(JavaScript docs)](https://developer.mozilla.org/en-US/docs/Web/API/CustomElementRegistry/define)
+    /// https://html.spec.whatwg.org/multipage/custom-elements.html#dom-customelementregistry-define
+    pub fn define( &self, name: &str, constructor: Any) -> Result<(), DefineError> {
```
koute: Hmmm.... we currently don't have a type which can represent a constructor. [According to this a constructor it can be detected with some tricks](https://stackoverflow.com/questions/40922531/how-to-check-if-a-javascript-function-is-a-constructor) so I guess we technically could add one?

You'd define such a type using the usual `#[derive(ReferenceType)]`, however you would omit the `#[reference(instance_of = "...")]` line and instead implement the `InstanceOf` trait yourself, and put the detection code inside of it. (You can take a look at e.g. the definition of the `error_boilerplate!` for an example on how to impl the `InstanceOf`) Then once we have this type you could use it here and in the return type of `get`.

src/webapi/custom_elements.rs
```diff
@@ -0,0 +1,62 @@
...
+        js!(
```
koute: You need to use `js_try!` here.

src/webapi/custom_elements.rs
```diff
@@ -0,0 +1,62 @@
...
+        js!( return @{self}.get(name); ).try_into().unwrap()
+    }
+
+    /// Returns a promise that will be fulfilled when a custom element becomes defined with the given name.
+    /// (If such a custom element is already defined, the returned promise is immediately fulfilled.)
+    ///
+    /// [(JavaScript docs)](https://developer.mozilla.org/en-US/docs/Web/API/CustomElementRegistry/whenDefined)
+    /// https://html.spec.whatwg.org/multipage/custom-elements.html#dom-customelementregistry-whendefined
+    #[cfg(feature = "experimental_features_which_may_break_on_minor_version_bumps")]
+    pub fn whenDefined( &self, name: &str ) -> TypedPromise<(), SyntaxError > {
```
koute: This should be named in snake case, that is `when_defined`

src/webapi/custom_elements.rs
```diff
@@ -0,0 +1,62 @@
...
+        let promise: Promise = js!( return @{self}.whenDefined(name); ).try_into().unwrap();
```
koute: Yes, this `unwrap` is fine. (As long as we have a guarantee that `whenDefined` will always return a `Promise`.)

src/webapi/custom_elements.rs
```diff
@@ -0,0 +1,62 @@
+use webcore::value::Reference;
+webapi::error::TypeError;
```
koute: You're missing `use` here. (:

src/webapi/custom_elements.rs
Pauan: I think the check for constructors is more involved, because custom elements require actual ES6 classes (and they must have the appropriate `extends`).

I think this also ties into the ability to define JS classes in stdweb, and we don't currently have a thorough plan for that.

I suppose we could create a simple `struct Constructor(Reference)` type for now (without any runtime checking).

koute: > I think the check for constructors is more involved, because custom elements require actual ES6 classes (and they must have the appropriate extends).

Isn't the ES6 class syntax *mostly* a syntax sugar? I grabbed one of the examples off MDN of custom elements and I've modified the element definition to use the old-style of defining classes, and it looks like the example still works...? (At least in Firefox.) I wonder if that's just an accident? (Granted, I'm getting an `InvalidStateError` exception from somewhere once the control returns to the main loop, but the example still works, and my manual class-defining-voodoo might be slightly not what it expects, so, uh, I'm confused.)

prasannavl: Sorry, I haven't had the time to get back on this yet. 

But, AFAIK, ES6 is not just sugar though. It can be done without ES6 syntax - using `Object.setPrototypeOf`, or `Reflect.setPrototypeOf` to be more compliant - however if I'm correct, these are reflection API in the JS engine, and are less performant than es6 syntax.

koute: > Could you post your code? When I tried using ES5 classes in Chrome it gave me errors when I called `customElements.define`.

I don't have in on hand anymore, however it really was just one of the examples picked at random with the class definition replaced with your garden variety old-style prototype-based JS class definition.

So this begs the question - is it actually possible to detect that something is a genuine ES6 class?

PR #300, opened 2018-11-09
Huge string performance boost

Pauan:
I noticed that when passing large strings from Rust to JS, it takes a long time.

This PR fixes that, by using the browser's built-in `TextDecoder` and `TextEncoder` features, if they are available (if they're not available it continues to use the old behavior). This is what wasm-bindgen does (which is how I found out about it).

When passing a large string to JS, the old code took ~19,115ms, but with this PR it now takes ~364ms!

As usual, it's best to review with [whitespace ignored](https://github.com/koute/stdweb/pull/300/files?utf8=%E2%9C%93&diff=split&w=1).

### DIFF
--- a/stdweb-internal-runtime/src/runtime.js
+++ b/stdweb-internal-runtime/src/runtime.js
@@ -210,7 +210,7 @@ Module.STDWEB_PRIVATE.serialize_array = function serialize_array( address, value
 // New browsers and recent Node
 var cachedEncoder = ( typeof TextEncoder === "function"
     ? new TextEncoder( "utf-8" )
-    // Old Node
+    // Old Node (before v11)
     : ( util && typeof util === "object" && typeof util.TextEncoder === "function"
         ? new util.TextEncoder( "utf-8" )
         // Old browsers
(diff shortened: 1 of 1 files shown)


stdweb-internal-runtime/src/runtime.js
```diff
@@ -212,18 +207,43 @@ Module.STDWEB_PRIVATE.serialize_array = function serialize_array( address, value
     }
 };
 
-Module.STDWEB_PRIVATE.from_js = function from_js( address, value ) {
-    var kind = Object.prototype.toString.call( value );
-    if( kind === "[object String]" ) {
+if ( typeof TextEncoder === "function" ) {
```
rickycodes: This is just personal preference, but I find `typeof` operator a bit verbose. Using the `in` operator to check the window object feels cleaner
```suggestion
if ( 'TextEncoder' in window ) {
```

koute: Under Node.js you don't have a `window` object and we want to stay compatible with it, so in this case the `typeof` version is better. (:

Although unfortunately under Node.js you need to import `TextEncoder` from `util` to use it so it will still use the old encoder/decoder, which reminds me - @Pauan I'm fine with merging this PR as-is, however would you be interesting in adding Node.js support? (Which should be relatively simple all things considered.)

rickycodes: ah, didn't realize y'all were targeting both browser and Node.js

Pauan: @koute Since it wasn't much code, I went ahead and added in support for older Nodes (before v11).

koute: Oh, I had no idea; that's good to know!

koute:
Looks good to me! Thanks a lot @Pauan!

Issue #304, opened 2018-12-02
Support Web Touch API


rickyhan:
The quicksilver game engine has stdweb as a dependency. In order to support mobile browsers, it requires WebTouch API: https://github.com/ryanisaacg/quicksilver/issues/366

Relevant:
https://developer.mozilla.org/en-US/docs/Web/API/Touch_events

PR #305, opened 2018-12-03
Add support for TouchEvent

rickyhan:
#304

### DIFF
--- a/src/webapi/events/touch.rs
+++ b/src/webapi/events/touch.rs
@@ -191,3 +191,36 @@ impl IEvent for TouchStart {}
 impl IUiEvent for TouchStart {}
 impl ITouchEvent for TouchStart {}
 
+
+#[cfg(all(test, feature = "web_test"))]
+mod tests {
+    use super::*;
+    use webapi::event::ConcreteEvent;
+
+    #[test]
+    fn test_touch_event() {
+        let event: Event = js!(
+            return new TouchEvent(
+                @{TouchMove::EVENT_TYPE},
+                {
+                    touches: [],
+                    targetTouches: []
+                    changedTouches: [],
+                    ctrlKey: true,
+                    shiftKey: true,
+                    altKey: true,
+                    metaKey: true
+                }
+            );
+        ).try_into().unwrap();
+        assert_eq!( event.event_type(), TouchMove::EVENT_TYPE );
+        assert!( event.ctrl_key() );
+        assert!( event.alt_key() );
+        assert!( event.shift_key() );
+        assert!( event.meta_key() );
+        assert_eq!( event.touches() );
+        assert_eq!( event.target_touches(), vec![] );
+        assert_eq!( event.changed_touches(), vec![] );
+    }
+
+}
--- a/src/webapi/events/touch.rs
+++ b/src/webapi/events/touch.rs
@@ -7,12 +7,12 @@ use webapi::touch::Touch;
 /// interacting with a touch device (such as a phone).
 ///
 /// [(JavaScript docs)](https://developer.mozilla.org/en-US/docs/Web/API/TouchEvent)
-// https://w3c.github.io/uievents/#touchevent
+// https://w3c.github.io/touch-events/#idl-def-touchevent
 pub trait ITouchEvent: IUiEvent {
     /// Returns whether the Alt key was down when this event was fired.
     ///
     /// [(JavaScript docs)](https://developer.mozilla.org/en-US/docs/Web/API/TouchEvent/altKey)
-    // https://w3c.github.io/uievents/#ref-for-dom-touchevent-altkey-1
+    // https://w3c.github.io/touch-events/#touchevent-interface
     #[inline]
     fn alt_key( &self ) -> bool {
         js!(
@@ -23,7 +23,7 @@ pub trait ITouchEvent: IUiEvent {
     /// Indicates whether the Ctrl key was down when this event fired.
     ///
     /// [(JavaScript docs)](https://developer.mozilla.org/en-US/docs/Web/API/TouchEvent/ctrlKey)
-    // https://w3c.github.io/uievents/#ref-for-dom-mouseevent-ctrlkey-1
+    // https://w3c.github.io/touch-events/#touchevent-interface
     #[inline]
     fn ctrl_key( &self ) -> bool {
         js!(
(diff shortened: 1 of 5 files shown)


koute:
Hmm... the failure on Travis is interesting; it might be related to the new Emscripten which the newest `cargo-web` is using. You can ignore it.

src/webapi/events/touch.rs
```diff
@@ -0,0 +1,193 @@
+use webcore::value::Reference;
+use webcore::try_from::TryInto;
+use webapi::event::{IEvent, IUiEvent, UiEvent, Event};
+use webapi::touch::Touch;
+
+/// The `ITouchEvent` interface represents events that occur due to the user
+/// interacting with a touch device (such as a phone).
+///
+/// [(JavaScript docs)](https://developer.mozilla.org/en-US/docs/Web/API/TouchEvent)
+// https://w3c.github.io/uievents/#touchevent
```
koute: These links are not valid. (:

In this case this link should be:

https://w3c.github.io/touch-events/#idl-def-touchevent

Please fix the other links too.

If it's not possible to link to something (e.g. I think the fields in the `TouchEvent` interface cannot be linked to?) then you can skip those.

src/webapi/touch.rs
```diff
@@ -0,0 +1,138 @@
...
+impl Touch {
+
+    /// Returns a unique identifier for this Touch object. A given touch point
+    /// (say, by a finger) will have the same identifier for the duration of
+    /// its movement around the surface. This lets you ensure that you're
+    /// tracking the same touch all the time.
+    ///
+    /// [(JavaScript docs)](https://developer.mozilla.org/en-US/docs/Web/API/Touch/identifier)
+    #[inline]
+    pub fn identifier(&self) -> f64 {
```
koute: According to the specs this is `long` (so `i32`), not `f64`.

src/webapi/touch.rs
```diff
@@ -0,0 +1,138 @@
...
+        js!(
+            return @{self.as_ref()}.pageY;
+        ).try_into().unwrap()
+    }
+
+    /// Returns the Element on which the touch point started when it was first placed on the surface, even if the touch point has since moved outside the interactive area of that element or even been removed from the document.
+    ///
+    /// [(JavaScript docs)](https://developer.mozilla.org/en-US/docs/Web/API/Touch/target)
+    #[inline]
+    pub fn target(&self) -> Element {
```
koute: According to specs this returns `EventTarget`, not `Element`.

(If you really want to add a convenance method which returns an `Element` you can add `fn target_element( &self ) -> Option< Element >` if you want.)

src/webapi/touch.rs
```diff
@@ -0,0 +1,138 @@
...
+        js!(
+            return @{self.as_ref()}.target;
+        ).try_into().unwrap()
+    }
+
+    /// Returns the X radius of the ellipse that most closely circumscribes the area of contact with the screen. The value is in pixels of the same scale as screenX.
+    ///
+    /// [(JavaScript docs)](https://developer.mozilla.org/en-US/docs/Web/API/Touch/radiusX)
+    #[inline]
+    pub fn radius_x(&self) -> f64 {
```
koute: Please leave a `TODO` here that this (and some of the others) should probably be changed to `f32` at a later date.

(Currently it doesn't matter as everything goes through the JS so it's ~turtles~ `f64` all the way down, but the specs say that those should be `f32` which I think will make a difference when we get native DOM bindings for WASM.)

koute:
Did you close this PR by a mistake? (:

src/webapi/touch.rs
```diff
@@ -0,0 +1,215 @@
...
+        js!(
+            return @{self.as_ref()}.target;
+        ).try_into().unwrap()
+    }
+
+
+
+    /// Returns the X radius of the ellipse that most closely circumscribes the area of contact with the screen. The value is in pixels of the same scale as screenX.
+    ///
+    /// TODO: f32 is not yet supported
```
koute: Please move this TODO out of the doc comment; it's a TODO for us, not necessarily for the users. (:

src/webapi/touch.rs
```diff
@@ -0,0 +1,215 @@
...
+impl Touch {
+
+    /// Returns a unique identifier for this Touch object. A given touch point
+    /// (say, by a finger) will have the same identifier for the duration of
+    /// its movement around the surface. This lets you ensure that you're
+    /// tracking the same touch all the time.
+    ///
+    /// [(JavaScript docs)](https://developer.mozilla.org/en-US/docs/Web/API/Touch/identifier)
+    #[inline]
+    pub fn identifier(&self) -> i64 {
```
koute: Why `i64`? (: The specs say that it's `long`, which is supposed to be `i32`, so is the `i64` here by mistake or is there an actual reason for it?

src/webapi/touch.rs
```diff
@@ -0,0 +1,215 @@
...
+    Reference,
+    Value,
+};
+
+/// The Touch interface represents a single contact point on a touch-sensitive
+/// device. The contact point is commonly a finger or stylus and the device may
+/// be a touchscreen or trackpad.
+///
+/// [(JavaScript docs)](https://developer.mozilla.org/en-US/docs/Web/API/Touch)
+#[derive(Clone, Debug, Eq, PartialEq, ReferenceType)]
```
koute: Please put this specs link here:

https://w3c.github.io/touch-events/#touch-interface

You don't have to add links to the methods as it's not possible to link to the fields anyway.

koute:
@rickyhan Can you confirm that you don't intend to continue these PRs? (It's perfectly fine if you don't. I just need to know so that I can reuse the work you've already done.)

KappaDistributive:
What's stopping this PR from getting pulled? 

The only concerns left seem to be the return type of identifier() and a missing reference to https://w3c.github.io/touch-events/#touch-interface.

If there's more work to do, I'd be happy to finish it.

koute:
@KappaDistributive Yes, you're right that it's almost finished. I was planning on pulling it in very soon myself; I just had some other stuff to finish. I guess I can do this sooner rather than layer. (:

koute:
Pushed to `master`. I'll probably release it tomorrow.

KappaDistributive:
Awesome. Thanks!

PR #311, opened 2018-12-19
Add support for setting the XHR response type

ryanisaacg:


src/webapi/xml_http_request.rs
```diff
@@ -99,6 +99,23 @@ impl XmlHttpRequest {
         }
     }
 
+    /// Set the type that the XmlHttpRequest should return
+    ///
+    /// [(JavaScript docs)](https://developer.mozilla.org/en-US/docs/Web/API/XMLHttpRequest/responseType)
+    // https://xhr.spec.whatwg.org/#ref-for-dom-xmlhttprequest-responsetype
+    pub fn set_response_type(&self, type: XhrResponseType) {
```
koute: This can throw `InvalidStateError` and `InvalidAccessError` according to the specs, so it should `Return` with an appropriate error type.

ryanisaacg: Possibly dumb question: what is the idiomatic way to return more than one error type in stdweb?

koute: Apologies for the late reply.

You'd probably want to use the `error_enum_boilerplate!` macro to make a sum type of every error the methon can return. (Just grep the sources for an example and do exactly the same thing.)

ryanisaacg: I pushed a commit, let me know if it's correct

koute:
Thanks!

PR #324, opened 2019-02-22
Implements FormData

sanpii:
https://developer.mozilla.org/en-US/docs/Web/API/FormData
https://xhr.spec.whatwg.org/#formdata

### DIFF
--- a/src/webapi/form_data.rs
+++ b/src/webapi/form_data.rs
@@ -1,0 +1,287 @@
+use webapi::blob::IBlob;
+use webapi::dom_exception::InvalidStateError;
+use webapi::element::IElement;
+use webapi::error::TypeError;
+use webapi::file::File;
+use webcore::try_from::TryFrom;
+use webcore::try_from::TryInto;
+use webcore::value::ConversionError;
+use webcore::value::Reference;
+use webcore::value::Value;
+
+/// The `FormData` interface provides a way to easily construct a set of key/value pairs
+/// representing form fields and their values, which can then be easily sent using the
+/// `XMLHttpRequest.send()` method. It uses the same format a form would use if the encoding type
+/// were set to `"multipart/form-data"`.
+///
+/// [(JavaScript docs)](https://developer.mozilla.org/en-US/docs/Web/API/FormData)
+// https://xhr.spec.whatwg.org/#formdata
+#[derive(Clone, Debug, PartialEq, Eq, ReferenceType)]
+#[reference(instance_of = "FormData")]
+pub struct FormData( Reference );
+
+/// Represents a type of data stores in FormData.
+#[derive(Clone, Debug, PartialEq)]
+pub enum FormDataEntry {
+    /// File data
+    File( File ),
+    /// Text data
+    String( String )
+}
+
+error_enum_boilerplate! {
+    FormDataFromElementError,
+    InvalidStateError, TypeError
+}
+
+impl TryFrom< Value > for FormDataEntry {
+    type Error = ConversionError;
+
+    fn try_from(value: Value) -> Result< Self, Self::Error > {
+        let entry = match value {
+            Value::String(s) => FormDataEntry::String(s),
+            Value::Reference(r) => FormDataEntry::File(File(r)),
+            _ => return Err(ConversionError::type_mismatch(&value, "string or reference".into())),
+        };
+
+        Ok(entry)
+    }
+}
+
+impl TryFrom< Value > for Option< FormDataEntry > {
+    type Error = ConversionError;
+
+    fn try_from(value: Value) -> Result< Self, Self::Error > {
+        let entry = match value {
+            Value::Null|Value::Undefined => None,
+            Value::String(s) => Some(FormDataEntry::String(s)),
+            Value::Reference(r) => Some(FormDataEntry::File(File(r))),
+            _ => return Err(ConversionError::type_mismatch(&value, "null, string or reference".into())),
+        };
+
+        Ok(entry)
+    }
+}
+
+impl FormData {
+    /// Creates a new `FormData`.
+    ///
+    /// [(JavaScript docs)](https://developer.mozilla.org/en-US/docs/Web/API/FormData/FormData)
+    pub fn new() -> Self {
+        js! (
+            return new FormData();
+        ).try_into().unwrap()
+    }
+
+    /// Creates a new `FormData` from a form element.
+    ///
+    /// [(JavaScript docs)](https://developer.mozilla.org/en-US/docs/Web/API/FormData/FormData)
+    pub fn from_element<T>( form: &T ) -> Result< Self, FormDataFromElementError > where T: IElement {
+        js_try! (
+            let form = @{form.as_ref()};
+
+            if ( ! (form instanceof HTMLFormElement) ) {
+                throw new TypeError("Argument 1 of FormData::from_element does not implement interface HTMLFormElement.");
+            }
+
+            return new FormData(form);
+        ).unwrap()
+    }
+
+    /// Appends a new value onto an existing key, or adds the key if it does not already exist.
+    ///
+    /// [(JavaScript docs)](https://developer.mozilla.org/en-US/docs/Web/API/FormData/append)
+    // https://xhr.spec.whatwg.org/#dom-formdata-append
+    pub fn append_string( &self, name: &str, value: &str ) {
+        js! { @(no_return)
+            @{self}.append(@{name}, @{value});
+        }
+    }
+
+    /// Appends a new blob onto an existing key, or adds the key if it does not already exist.
+    ///
+    /// [(JavaScript docs)](https://developer.mozilla.org/en-US/docs/Web/API/FormData/append)
+    // https://xhr.spec.whatwg.org/#dom-formdata-append-blob
+    pub fn append_blob<T>( &self, name: &str, value: &T, filename: Option< &str > ) where T: IBlob {
+        js! { @(no_return)
+            @{self}.append(@{name}, @{value.as_ref()}, @{filename});
+        }
+    }
+
+    /// [(JavaScript docs)](https://developer.mozilla.org/en-US/docs/Web/API/FormData/delete)
+    // https://xhr.spec.whatwg.org/#dom-formdata-delete
+    pub fn delete( &self, name: &str ) {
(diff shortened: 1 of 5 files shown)


src/webapi/formdata.rs
```diff
@@ -0,0 +1,207 @@
...
+// https://xhr.spec.whatwg.org/#formdata
+#[derive(Clone, Debug, PartialEq, Eq, ReferenceType)]
+#[reference(instance_of = "FormData")]
+pub struct FormData( Reference );
+
+impl FormData {
+    /// Creates a new `FormData`.
+    ///
+    /// [(JavaScript docs)](https://developer.mozilla.org/en-US/docs/Web/API/FormData/FormData)
+    pub fn new( form: &Element ) -> Self {
```
koute: From what I can see in the specs this doesn't take any `HTMLElement`, it requires `HTMLFormElement`, and the contructor can also be called without any parameters, and it can throw `InvalidStateError`.

But you don't have to add all of that if you don't need it; we can keep this function *if* we change its prototype to something like this:

```rust
pub fn from_element<T>( form: &T ) -> Result< Self, FormDataFromElementError > where T: IElement
```

where the `FormDataFromElementError` is an enum (use the `error_enum_boilerplate!` macro) with `InvalidStateError` and `TypeError` variants.

So, take the argument by its trait (since we currently don't have any `Deref`s; that might change in the future, but not yet.), and actually handle the cases where it either throws an `InvalidStateError`, or the passed `form` isn't actually an `HTMLFormElement`.

src/webapi/formdata.rs
```diff
@@ -0,0 +1,207 @@
...
+        js! (
+            return new FormData(@{form});
+        ).try_into().unwrap()
+    }
+
+    /// Appends a new value onto an existing key, or adds the key if it does not already exist.
+    ///
+    /// [(JavaScript docs)](https://developer.mozilla.org/en-US/docs/Web/API/FormData/append)
+    // https://xhr.spec.whatwg.org/#dom-formdata-append
+    pub fn append( &self, name: &str, value: &str ) {
```
koute: I'd probably prefer to call it `append_string` since none of the overloads of `append`s really feel like the default one.

src/webapi/formdata.rs
```diff
@@ -0,0 +1,207 @@
...
+        js! (
+            return @{self}.has(@{name});
+        ).try_into().unwrap()
+    }
+
+    /// Sets a new value for an existing key, or adds the key/value if it does not already exist.
+    ///
+    /// [(JavaScript docs)](https://developer.mozilla.org/en-US/docs/Web/API/FormData/set)
+    // https://xhr.spec.whatwg.org/#dom-formdata-set
+    pub fn set( &self, name: &str, value: &str ) {
```
koute: Same here as with the `append` overloads - `set_string`

src/webapi/formdata.rs
```diff
@@ -0,0 +1,207 @@
...
+        js! { @(no_return)
+            @{self}.delete(@{name});
+        }
+    }
+
+    /// Deletes a key and its value(s).
+    ///
+    /// [(JavaScript docs)](https://developer.mozilla.org/en-US/docs/Web/API/FormData/get)
+    // https://xhr.spec.whatwg.org/#dom-formdata-get
+    pub fn get( &self, name: &str ) -> Value {
```
koute: This should return `Option< FormDataEntry >` where `FormDataEntry` is:

```
#[derive(Clone, Debug)]
enum FormDataEntry {
    File( File ),
    String( String )
}
```

src/webapi/formdata.rs
```diff
@@ -0,0 +1,207 @@
...
+        js! (
+            return @{self}.get(@{name});
+        ).try_into().unwrap()
+    }
+
+    /// Returns all the values associated with a given key.
+    ///
+    /// [(JavaScript docs)](https://developer.mozilla.org/en-US/docs/Web/API/FormData/getAll)
+    // https://xhr.spec.whatwg.org/#dom-formdata-getall
+    pub fn get_all( &self, name: &str ) -> Vec< Value > {
```
koute: This should return `Vec< FormDataEntry >`.

src/webapi/mod.rs
```diff
@@ -44,6 +44,7 @@ pub mod selection;
 pub mod midi;
 pub mod slotable;
 pub mod shadow_root;
+pub mod formdata;
```
koute: Rename to `form_data`

src/webapi/formdata.rs
```diff
@@ -0,0 +1,207 @@
...
+        js! { @(no_return)
+            @{self}.append(@{name}, @{value});
+        }
+    }
+
+    /// Appends a new blob onto an existing key, or adds the key if it does not already exist.
+    ///
+    /// [(JavaScript docs)](https://developer.mozilla.org/en-US/docs/Web/API/FormData/append)
+    // https://xhr.spec.whatwg.org/#dom-formdata-append-blob
+    pub fn append_blob( &self, name: &str, value: Blob, filename: &str ) {
```
koute: Change it to:

```rust
pub fn append_blob<T>( &self, name: &str, value: &T, filename: Option< &str > ) where T: IBlob {
```

src/webapi/formdata.rs
```diff
@@ -0,0 +1,207 @@
...
+        js! { @(no_return)
+            @{self}.set(@{name}, @{value});
+        }
+    }
+
+    /// Sets a new blob for an existing key, or adds the key/value if it does not already exist.
+    ///
+    /// [(JavaScript docs)](https://developer.mozilla.org/en-US/docs/Web/API/FormData/set)
+    // https://xhr.spec.whatwg.org/#dom-formdata-set-blob
+    pub fn set_blob( &self, name: &str, value: Blob, filename: &str ) {
```
koute: Change it to:

```rust
pub fn set_blob<T>( &self, name: &str, value: &T, filename: Option< &str > ) where T: IBlob {
```

sanpii:
@koute Thank you for your comments.

I have just one problem with the `from_element` function: it doesn’t fail with a non-HTMLFormElement. It seems the behavior of chromium, but I also try:
```rust
        js_try! (
            if ( ! @{form} instanceof HTMLFormElement ) {
                throw new InvalidStateError();
            }

            return new FormData(@{form});
        ).unwrap()
```

Without success.

koute:
Yep. It seems that for this:

```
new FormData(document.createElement("div"))
```

Firefox throws this:

```
TypeError: Argument 1 of FormData.constructor does not implement interface HTMLFormElement.
```

while Chromium ignores it. (Which, I'd argue, is a violation of the specs since the specs clearly say that the constructor expects an `HTMLFormElement`.)

We should pick the more strict behavior here and just generate the `TypeError` ourselves. The code you've pasted in your comment should work, however:

1. Throw `TypeError` instead of `InvalidStateError`, and be sure to construct a `TypeError` with the same error message as is generated by Firefox.
2. Only pass `@{form}` once, e.g. do `var form = @{form};`.
3. The precedence of the `instanceof` is not what you expect; `!` is being applied *first*, so what happens here is that the `form` gets converted to a boolean and negated (the joys of implicit type conversion), which makes the whole condition always false.

sanpii:
Thank you to fixing my lack in javascript :smiley:

src/webapi/form_data.rs
```diff
@@ -0,0 +1,291 @@
...
+        js! (
+            return new FormData();
+        ).try_into().unwrap()
+    }
+
+    /// Creates a new `FormData` from a form element.
+    ///
+    /// [(JavaScript docs)](https://developer.mozilla.org/en-US/docs/Web/API/FormData/FormData)
+    pub fn from_element<T>( form: &T ) -> Result< Self, FormDataFromElementError >
+        where T: IElement + JsSerialize {
```
koute: To be consistent with the rest of the library remove the `JsSerialize` bound and use `@{form.as_ref()}` instead of `@{form}`. (Sorry, it slipped my mind the first time I looked through this.) Same for other methods here.

src/webapi/form_data.rs
```diff
@@ -0,0 +1,291 @@
...
+        js! { @(no_return)
+            @{self}.append(@{name}, @{value});
+        }
+    }
+
+    /// Appends a new blob onto an existing key, or adds the key if it does not already exist.
+    ///
+    /// [(JavaScript docs)](https://developer.mozilla.org/en-US/docs/Web/API/FormData/append)
+    // https://xhr.spec.whatwg.org/#dom-formdata-append-blob
+    pub fn append_blob<T>( &self, name: &str, value: &T, filename: &str )
```
koute: You didn't change the type of `filename` argument. (:

src/webapi/form_data.rs
```diff
@@ -0,0 +1,291 @@
...
+        js! { @(no_return)
+            @{self}.set(@{name}, @{value});
+        }
+    }
+
+    /// Sets a new blob for an existing key, or adds the key/value if it does not already exist.
+    ///
+    /// [(JavaScript docs)](https://developer.mozilla.org/en-US/docs/Web/API/FormData/set)
+    // https://xhr.spec.whatwg.org/#dom-formdata-set-blob
+    pub fn set_blob<T>( &self, name: &str, value: &T, filename: &str )
```
koute: You didn't change the type of `filename` here either.

src/webapi/form_data.rs
```diff
@@ -0,0 +1,291 @@
...
+        Ok(entry)
+    }
+}
+
+impl TryFrom< Value > for Option< FormDataEntry > {
+    type Error = ConversionError;
+
+    fn try_from(value: Value) -> Result< Self, Self::Error > {
+        let entry = match value {
+            Value::Null => None,
```
koute: Since this is user-visible handling `Value::Undefined` makes sense too.

sanpii:
Done, I hope I missed nothing.

Is it intentional to call `FormDataEntry` instead of `FormDataEntryValue` as in spec?

koute:
> Is it intentional to call `FormDataEntry` instead of `FormDataEntryValue` as in spec?

Yes. We tend to keep the names inline with what is in the specs, however in some cases we do use a different name. Here the name from the specs is somewhat long, and the suffix of `Value` seems somewhat superfluous, so it makes sense to shorten it a little. 

Anyway, looks good to me! Thanks!

PR #327, opened 2019-03-05
Expose `MouseButtonsState`

ForsakenHarmony:


koute:
Thanks for the PR!

Please add the relevant docs to the `MousteButtonState::is_down` so that [Travis passes](https://travis-ci.org/koute/stdweb/jobs/502292721#L1076). (The method *is* mostly self explanatory, but we enforce that every public item has at least *some* docs, which means even in cases like this we need to add *something*.)

ForsakenHarmony:
Done

koute:
LGTM! Thanks!

PR #335, opened 2019-04-28
Updating to the stable Futures API

Pauan:
The Future trait was recently stabilized. There were some last-minute changes, so this PR fixes stdweb to work with the latest Rust compiler.

Hopefully there shouldn't be any more changes (and Future will soon be available in Stable).


koute:
Thanks!

PR #337, opened 2019-05-03
Add the clear_timeout method to IWindowOrWorker

ryanisaacg:
I'm not too sure about the change I made to set_timeout, but I know that in JS set_timeout returns an ID that can be passed to clear_timeout.

### DIFF
--- a/src/webapi/window_or_worker.rs
+++ b/src/webapi/window_or_worker.rs
@@ -21,12 +21,22 @@ pub trait IWindowOrWorker: ReferenceType {
     ///
     /// [(JavaScript docs)](https://developer.mozilla.org/en-US/docs/Web/API/WindowOrWorkerGlobalScope/setTimeout)
     // https://html.spec.whatwg.org/#windoworworkerglobalscope-mixin:dom-settimeout
-    fn set_timeout< F: FnOnce() + 'static >( &self, callback: F, timeout: u32 ) {
+    fn set_timeout< F: FnOnce() + 'static >( &self, callback: F, timeout: u32 ) -> i32 {
         let callback = Box::into_raw( Box::new( callback ) );
         __js_raw_asm!( "\
             Module.STDWEB_PRIVATE.acquire_js_reference( $0 ).setTimeout( function() {\
                 Module.STDWEB_PRIVATE.dyncall( 'vi', $1, [$2] );\
             }, $3 );\
-        ", self.as_ref().as_raw(), funcall_adapter::< F > as extern fn( *mut F ), callback, timeout );
+        ", self.as_ref().as_raw(), funcall_adapter::< F > as extern fn( *mut F ), callback, timeout )
+    }
+
+    /// Clears a timer previously established by set_timeout
+    ///
+    /// [(JavaScript docs)](https://developer.mozilla.org/en-US/docs/Web/API/WindowOrWorkerGlobalScope/clearTimeout)
+    // https://html.spec.whatwg.org/#windoworworkerglobalscope-mixin:dom-clear-timeout
+    fn clear_timeout( &self, id: i32 ) {
+        js! {
+            clearTimeout(@{id})
+        }
     }
 }
--- a/src/webapi/window_or_worker.rs
+++ b/src/webapi/window_or_worker.rs
@@ -1,4 +1,6 @@
+use webcore::value::Reference;
 use webcore::reference_type::ReferenceType;
+use webcore::try_from::TryInto;
 
 extern fn funcall_adapter< F: FnOnce() >( callback: *mut F ) {
     let callback = unsafe {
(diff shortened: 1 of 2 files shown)


koute:
Thanks for the PR!

1. The JS snippet inside of `set_timeout` is missing a `return` so it will always return zero.
2. The snippet in `clear_timeout` is missing a semicolon.
3. We can' change the prototype of this function without breaking backwards compatibility. If we want to add this in 0.4 we should add another method, e.g. `set_clearable_timeout`.
4. This will leak memory. The clear function needs to `.drop()` the closure on JS side.
5. Instead of returning a raw `i32` I'd prefer to return an opaque handle, e.g. `TimeoutHandle`, and define `TimeoutHandle::clear()` instead.

ryanisaacg:
I'm unsure how to fix 4, but I think I've addressed the other points?

ryanisaacg:
@koute ping?

koute:
Oh, sorry for the delay!

1. It'd be better to leave the current `set_timeout` as is was (instead of defining it in terms of `set_clearable_timeout`) since that way it's going to be a little faster.
2. Please take a look at the `src/webapi/mutation_observer.rs` as to how to drop the callback when it's canceled, e.g. `MutationObserver::new` on how to set it up, and the `Drop` impl on how to do when the user wants to `clear` the timeout.

ryanisaacg:
Sorry for disappearing for two weeks!

I think I have the changes, but I'm not confident in them.

src/webapi/window_or_worker.rs
```diff
@@ -29,4 +31,32 @@ pub trait IWindowOrWorker: ReferenceType {
...
         ", self.as_ref().as_raw(), funcall_adapter::< F > as extern fn( *mut F ), callback, timeout );
     }
+
+    /// Sets a timer which executes a function once after the timer expires and can be cleared
+    ///
+    /// [(JavaScript docs)](https://developer.mozilla.org/en-US/docs/Web/API/WindowOrWorkerGlobalScope/setTimeout)
+    // https://html.spec.whatwg.org/#windoworworkerglobalscope-mixin:dom-settimeout
+    fn set_clearable_timeout< F: FnOnce() + 'static >( &self, callback: F, timeout: u32 ) -> TimeoutHandle {
+        //let callback = Box::into_raw( Box::new( callback ) );
+        let callback_reference: Reference = js! ( return @{callback}; ).try_into().unwrap();
```
koute: You need to use the wrapper from `webcore/once.rs` to pass an `FnOnce` into `js!`.

src/webapi/window_or_worker.rs
```diff
@@ -29,4 +31,32 @@ pub trait IWindowOrWorker: ReferenceType {
...
+        let id = js! {
+            setTimeout(@{callback_reference}, @{timeout});
```
koute: You're missing a `return`, and IIRC you should to use `js!()` instead of `js! {}`.

src/webapi/window_or_worker.rs
```diff
@@ -29,4 +31,32 @@ pub trait IWindowOrWorker: ReferenceType {
...
+        }.try_into().unwrap();
```
koute: It would probably be simpler to just have a single `js!` and return a JS `Object` from it, instead of having two of them.

ryanisaacg:
@koute The 'Once' docs seem to imply that the 'drop' stuff is unnecessary?

ryanisaacg:
@koute I think the CI failure is due to a flaky test, any other issues with the PR?

src/webapi/window_or_worker.rs
```diff
@@ -29,4 +32,35 @@ pub trait IWindowOrWorker: ReferenceType {
...
+
+impl TimeoutHandle {
+    /// Clears a timer previously established by set_clearable_timeout
+    ///
+    /// [(JavaScript docs)](https://developer.mozilla.org/en-US/docs/Web/API/WindowOrWorkerGlobalScope/clearTimeout)
+    // https://html.spec.whatwg.org/#windoworworkerglobalscope-mixin:dom-clear-timeout
+     pub fn clear( & self ) {
+        js! { @(no_return)
+            clearTimeout(@{&self.0}.id);
+            @{&self.0}.callback.drop();
```
koute: It'll be more efficient to do something like this instead of serializing `self.0` twice (the `js!` macro is not smart enough to collapse duplicate interpolations):

```js
var val = @{&self.0};
clearTimeout(val.id);
val.callback.drop();
```

src/webapi/window_or_worker.rs
```diff
@@ -29,4 +32,35 @@ pub trait IWindowOrWorker: ReferenceType {
...
+#[derive(Debug)]
+/// A reference to a timeout object created by set_clearable_timeout
+pub struct TimeoutHandle(Object);
+
+impl TimeoutHandle {
+    /// Clears a timer previously established by set_clearable_timeout
+    ///
+    /// [(JavaScript docs)](https://developer.mozilla.org/en-US/docs/Web/API/WindowOrWorkerGlobalScope/clearTimeout)
+    // https://html.spec.whatwg.org/#windoworworkerglobalscope-mixin:dom-clear-timeout
+     pub fn clear( & self ) {
```
koute: Hm... I'd probably prefer for this to take `self` instead. (E.g. just like with `RequestAnimationFrameHandle` or `EventListenerHandle`.)

koute:
Yep, the failure is totally spurious. Feel free to ignore it for now.

ryanisaacg:
@koute I think those issues should all be fixed

koute:
Looks good to me! Thanks!

PR #348, opened 2019-06-18
Use "$crate::" when referencing helper macros

### LINKED ISSUE #347: Import js macro using `use` requires importing _js_impl as well
As part of transition to rust 2018 idioms in one of my projects, I ran into an issue with macro imports in `stdweb`. Particularly, the following code fails to compile:

```rust
use stdweb::js;

fn main() {
    js! {
        console.log("Hello, world!");
    }
}
```

With the following error:

```
error: cannot find macro `_js_impl!` in this scope
 --> src/main.rs:4:5
  |
4 | /     js! {
5 | |         console.log("Hello, world!");
6 | |     }
  | |_____^
  |
  = note: this error originates in a macro outside of the current crate (in Nightly builds, run with -Z external-macro-backtrace for more info)
```

Importing the macro using `#[macro_use]` still works, as does 

```rust
use stdweb::{_js_impl, js};

fn main() {
    js! {
        console.log("Hello, world!");
    }
}
```

The [edition guide's macro changes section](https://doc.rust-lang.org/edition-guide/rust-2018/macros/macro-changes.html#local-helper-macros) documents this error, as well as two ways to fix it. Since the minimum rust version for this crate is 1.33.0, I think using `$crate::` to reference local macros should be reasonable?

Edit: realized the minimum rust version of stdweb is already 1.33.0

daboross:
Fixes #347.

This fix allows users to import `js_serializable`, `console` and `js` macros, as well as the `stdweb_derive::ReferenceType` proc macro, using a single `use` statement each, without requiring importing any of the helper macros. The fix implement is described in [the edition guide's macro changes section](https://doc.rust-lang.org/stable/edition-guide/rust-2018/macros/macro-changes.html#macros-with-crate-prefix), and as I understand it this is the preferably way to handle helper macros in crates with a minimum rust version of 1.30.0+.

The actual change is just to use "$crate::" when referencing any helper macros, and to change one reference of `__js_raw_asm!` to `::stdweb::__js_raw_asm!` in `stdweb-derive`.

This change should be compatible with any usage of the macros, including importing using `#[macro_use]` and usage in 2015-edition crates.

### DIFF
--- a/src/ecosystem/serde.rs
+++ b/src/ecosystem/serde.rs
@@ -1160,15 +1160,15 @@ macro_rules! __js_deserializable_serde_boilerplate {
 #[macro_export]
 macro_rules! js_serializable {
     ($kind:tt) => {
-        __js_serializable_serde_boilerplate!( () ($kind) () );
+        $crate::__js_serializable_serde_boilerplate!( () ($kind) () );
     };
 
     (impl< $($impl_arg:tt),* > for $kind:ty where $($bounds:tt)*) => {
-        __js_serializable_serde_boilerplate!( ($($impl_arg),*) ($kind) ($($bounds)*) );
+        $crate::__js_serializable_serde_boilerplate!( ($($impl_arg),*) ($kind) ($($bounds)*) );
     };
 
     (impl< $($impl_arg:tt),* > for $kind:ty) => {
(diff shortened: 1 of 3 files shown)


koute:
LGTM! Thanks!

PR #349, opened 2019-06-19
Implement slice for IBlob

Gowee:
Hi.

The `slice` method on `Blob` object is very useful when handling big blobs (esp. files).

I implement it following the same pattern in the file. But I am not sure whether the function signature is appropriate or not. For example, I do not know what is the convention to handle optional parameters and I just use triple `Option` which seems a bit clumsy. So suggestions are appreciated. 

Also not sure how to write proper tests, but I have tested it locally.


src/webapi/blob.rs
```diff
@@ -31,6 +31,22 @@ pub trait IBlob: ReferenceType {
             Some( mime )
         }
     }
+
+    /// Create a new `Blob` object containing the data in the specified range of bytes of the
+    /// source `Blob`.
+    /// 
+    /// [(JavaScript docs)](https://developer.mozilla.org/en-US/docs/Web/API/Blob/slice)
+    /// https://w3c.github.io/FileAPI/#ref-for-dfn-slice
```
koute: The link to the specs should not be part of the docs. (Change `///` into `//` for this line.)

src/webapi/blob.rs
```diff
@@ -31,6 +31,22 @@ pub trait IBlob: ReferenceType {
...
+    fn slice( &self, start: Option<u64>, end: Option<u64>, content_type: Option<&str> ) -> Blob {
```
koute: Hm... while this works it's not very Rusty.

Something like this using [RangeBounds](https://doc.rust-lang.org/std/ops/trait.RangeBounds.html) would be better I think:

```rust
fn slice< T >( &self, range: T ) -> Blob where T: RangeBounds { ... }
fn slice_with_content_type( &self, range: T, content_type: &str ) -> Blob where T: RangeBounds { ... }
```

then you could use it:

```
blob.slice( 0..12 );
blob.slice( ..12 );
blob.slice( 10.. );
// etc.
```

src/webapi/blob.rs
```diff
@@ -31,6 +31,22 @@ pub trait IBlob: ReferenceType {
...
+        let reference = self.as_ref();
+        js! (
+            return @{reference}.slice(
+                @{start.map(|val| val as f64)},
```
koute: Convert the `u64` to a `Number` with `try_into().unwrap()` and pass that instead. (That will prevent silent truncation.)

koute:
Thanks for the PR!

As far as the tests go - you could create a `Blob` with some dummy data in it (just use the currently unexposed constructor through the `js!` macro to construct such a `Blob`), and then just add a normal test (grep for `web_test` for examples).

Gowee:
Thanks for your advice!

I have rewritten some of the relevant codes accordingly except that tests are still left undone because I have trouble setting up the test environment (in my WSL which has no Chrome/Chromium).

src/webapi/blob.rs
```diff
@@ -31,6 +34,46 @@ pub trait IBlob: ReferenceType {
...
+    /// [(JavaScript docs)](https://developer.mozilla.org/en-US/docs/Web/API/Blob/slice)
+    // https://w3c.github.io/FileAPI/#ref-for-dfn-slice
+    fn slice< T >( &self, range: T ) -> Blob
+        where T: RangeBounds<u64>
+    {
+        self._slice(range, None)
+    }
+
+    /// [slice](IBlob::slice) `Blob` with the provided `content_type`.
+    fn slice_with_content_type< T >( &self, range: T, content_type: &str ) -> Blob
```
Gowee: `IBlob` also has a function named `mime` of which the meaning is the same as `content_type` here. Maybe I need to change either so that the two to be consistent? Any idea?

src/webapi/blob.rs
```diff
@@ -31,6 +34,46 @@ pub trait IBlob: ReferenceType {
...
+        where T: RangeBounds<u64>
+    {
+        self._slice(range, Some(content_type))
+    }
+
+    #[doc(hidden)]
+    fn _slice< T >( &self, range: T, content_type: Option<&str> ) -> Blob
```
koute: It's actually not necessary to make this a trait method. You can just make it into a freestanding function, e.g. something like this:

```rust
fn slice_blob< T, R >( blob: &T, range: R, content_type: Option< &str > ) -> Blob
    where T: IBlob, R: RangeBounds< u64 >
{
...
}
```

and then just call it from the actual trait methods.

src/webcore/try_from.rs
```diff
@@ -27,3 +27,15 @@ impl< T, U > TryInto< U > for T where U: TryFrom< T > {
         U::try_from( self )
     }
 }
+
+impl< T, U > TryFrom< Option<T> > for Option<U>
```
koute: Please get rid of this impl. In the future I'd like to get rid of our `TryFrom` and just use one from the stdlib, and this impl will make that harder.

src/webapi/blob.rs
koute: It's just a name so it doesn't really matter that much. We can't change the name of the `mime` method without breaking backwards compatibility, so the only thing we can do is to change the name of this parameter. You can rename it to `mime` if you want.

src/webapi/blob.rs
```diff
@@ -31,6 +34,46 @@ pub trait IBlob: ReferenceType {
...
+        where T: RangeBounds<u64>
+    {
+        let start: Option<Number> = match range.start_bound() {
+            Bound::Included(&n) => Some(n),
+            Bound::Excluded(&n) => Some(n + 1),
+            _ => None
```
koute: Please explicitly match `Bound::Unbounded` instead of `_`, make it resolve to `0` (since that's the default), and get rid of the `Option`.

src/webapi/blob.rs
```diff
@@ -31,6 +34,46 @@ pub trait IBlob: ReferenceType {
...
+        }.try_into().unwrap();
+        let end: Option<Number> = match range.end_bound() {
+            Bound::Included(&n) => Some(n + 1),
+            Bound::Excluded(&n) => Some(n),
+            _ => None
```
koute: Please explicitly match `Bound::Unbounded`. You should be able to use something like `.map( |value| value.try_into().unwrap() )` on this `Option` so that the extra `TryFrom` impl you've added will be unnecessary.

Gowee: Need I also replace `None` with something like `self.len()` here? I was assuming it be converted to `undefined` so that the default parameter is used. Was I right?

src/webapi/blob.rs
Gowee: I notice that you are using some formatting conventions in codes which are different from those enforced by the default rustfmt. Do you have a formatter / configurations so that I can adhere easily, if necessary?

koute:
Thanks!

PR #351, opened 2019-07-01
Don't assume libstd is in the root.

josephlr:
Some of the code in `stdweb` assumes that `std` is always at the root, an uses things like `::std::foo::bar`, while some other code just uses things like `std::foo::bar`. The second usage pattern is correct. `libstd` is not always at the root.

Specifically, [`getrandom`](github.com/rust-random/getrandom) is mostly a `no_std` library, but when using `stdweb` it uses `libstd`. It does this by putting `extern crate std` in the `wasm32` submodule. In this case, `std` is present but not in the root.

This PR changes all uses of `::std::` to just be `std::`.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -107,8 +107,7 @@
     missing_docs,
     missing_debug_implementations,
     trivial_numeric_casts,
-    unused_import_braces,
-    unused_qualifications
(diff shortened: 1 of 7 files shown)


koute:
Thanks!

PR #354, opened 2019-07-09
Re-export the unload and beforeunload events via the web::events API

ryanisaacg:
Also make the events concrete so they can be listened for


koute:
Thanks!

PR #362, opened 2019-08-14
Allow converting &Array to Vec<V>

daboross:
I implemented this by replacing the original `impl TryFrom<Array> for Vec<V>` impl with one for `&'a Array`, then adding back the original implementation referencing this reference-based one.

Fixes #228.


koute:
Looks good to me! Thanks!

daboross:
Thanks!

PR #363, opened 2019-08-15
Add Object::to_iter for retrieving an object's keys and values without allocating a rust collection

### LINKED ISSUE #359: Allow converting Object into HashMap with key-type which implements FromStr?
Right now, there are several ways to convert objects to rust `HashMap`s, but none are exactly ideal for custom key types (which implement `FromStr`).

There's an impl `TryFrom<Value> for HashMap<String, V>`, but this forces the key type to be a string.

There's using `stdweb::serde::Serde` to wrap the `HashMap<MyKey, V>`, but this doesn't allow the value to be a reference-type.

Then there's the option of using retrieving a `HashMap<String, V>`, using `into_iter()`, converting keys, and building a second `HashMap` of the correct types. This works, but is inefficient in that it requires allocating two full `HashMap`s to do the conversion.

Would it be reasonable to have a method on `Object` for converting to `HashMap<K, V>`, using either `FromStr` or a custom conversion function for the key conversion?

The code implementing the impl `TryFrom<Object> for HashMap<String, V>` uses an internal function `deserialize_object` which I think could be used to implement this, but I'm not sure.

daboross:
This could be a bit overzealous, but here it is.

This allows converting an `Object` into a container like `HashMap<i32, V>` rather than only `HashMap<String, V>`, as well as any other map-like thing. It uses a user-provided parsing function to parse the string keys into the other type.

I can make this less permissive if needed - especially since I was only originally thinking about HashMaps.

Fixes #359.

### DIFF
--- a/src/webcore/serialization.rs
+++ b/src/webcore/serialization.rs
@@ -4,6 +4,7 @@ use std::i32;
 use std::collections::{BTreeMap, HashMap};
 use std::marker::PhantomData;
 use std::hash::Hash;
+use std::ops::Deref;
 
 use webcore::ffi;
 use webcore::callfn::{CallOnce, CallMut};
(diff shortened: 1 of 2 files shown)


koute:
Sorry for the delay and thanks for the PR!

Wow, this is indeed very generic and somewhat complex, and I probably don't really want to expose a `serde`-specific error type here. Hmm... perhaps it would be better to just make a method which simply returns an iterator? E.g. something like this: (`to_iter` and not `iter` because creating such iterator is not zero cost)

```
    fn to_iter( &self ) -> impl ExactSizeIterator< Item = (String, Value) > { ... }
```

then the user would just be able to efficiently convert the keys, convert the values, and collect the collection into whatever they want (a `HashMap`, a `BTreeMap`, or something else).

Such an iterator could be adapted from the existing private `ObjectDeserializer` - basically, 1) convert the slices to raw pointers, and 2) add a `Drop` which will consume the whole iterator (to prevent leaks) and deallocate the arrays for keys and values. That should, *I think*, work. What would you say to this?

daboross:
This is great!

I didn't really feel satisfied with forcing reliance on `ConversionError` - and the iterator feels like a cleaner solution overall too.

Since `deserialize_object` has the comment `// TODO: Panic-safety.`, I'm assuming it could benefit from an iterator with a drop implementation too. Would it be alright if I modified `ObjectDeserializer` to implement `Drop` like you described, and then exposed a method like `serialization::deserialize_object` which just returns an `ObjectDeserializer`?

With your first comment - all delays are fine with me! Nothing I'm doing with this is mission-critical, or anything - and I can't claim to be consistently punctual myself.

daboross:
> Since `deserialize_object` has the comment `// TODO: Panic-safety.`, I'm assuming it could benefit from an iterator with a drop implementation too. Would it be alright if I modified `ObjectDeserializer` to implement `Drop` like you described, and then exposed a method like `serialization::deserialize_object` which just returns an `ObjectDeserializer`?

I implemented this. Let me know if you'd rather have a separate type and not have it affect the existing deserialization code?

I'm not 100% sure whether panicking in WASM will unwind the stack, but if it does, this change should make `deserialize_object` panic-safe.

src/webcore/serialization.rs
```diff
@@ -214,13 +215,54 @@ impl SerializedUntaggedArray {
...
+        assert_ne!( mem::size_of::< T >(), 0 );
+        unsafe {
+            ffi::dealloc( self.ptr as *mut u8, self.length * mem::size_of::< T >() );
+        }
+    }
+}
+
+pub struct ObjectDeserializer {
+    key_slice: OwnedFfiSlice < SerializedUntaggedString >,
+    value_slice: OwnedFfiSlice < SerializedValue< 'static > >,
```
daboross: I couldn't find any reason why using `'static` here would cause a safety risk, but it seems like it could be wrong in some way since the lifetime parameter does exist there. Since this makes `ObjectDeserializer` own its contents and drop it on drop, it didn't feel right to use any external lifetime parameter here.

koute:
The `'static` lifetime looks fine to me in this case.

And I don't think unwinding works in WASM yet (at least it didn't the last time I checked), so on panic the code might leak anyway, but there isn't really anything we can do about it now.

Anyway, looks good to me! Thanks!

daboross:
> And I don't think unwinding works in WA SM yet (at least it didn't the last time I checked), so on panic the code might leak anyway, but there isn't really anything we can do about it now.

Makes sense. It'll work once we get real unwinding, at least.

In any case, I'm glad to have this - and with a nicer interface than my original pr! Thanks!

PR #364, opened 2019-08-19
fix futures example to work with current nightly (2019-08-18)

fkohlgrueber:
Update the futures example to work with current nightlies (tested with 2019-08-18). I also added the futures example to the examples workspace.


koute:
Thanks!

PR #365, opened 2019-08-29
Add selectionStart and selectionEnd attributes for input elements

Wodann:
Added selectionStart and selectionEnd attributes for input elements. These attribute only apply when the input is a text, search, url, telephone or password.

https://html.spec.whatwg.org/#textFieldSelection

As far as I can tell, similar functionality does not have any unit tests. If there is a need and format for these kind of unit tests, could you point me to it?

### DIFF
--- a/src/webapi/html_elements/input.rs
+++ b/src/webapi/html_elements/input.rs
@@ -1,5 +1,6 @@
 use webcore::value::Reference;
 use webcore::try_from::TryInto;
+use webapi::dom_exception::InvalidStateError;
 use webapi::event_target::{IEventTarget, EventTarget};
 use webapi::node::{INode, Node};
 use webapi::element::{IElement, Element};
@@ -22,6 +23,7 @@ impl IHtmlElement for InputElement {}
 
 impl InputElement {
     /// The value of the control. This attribute is optional except when the input is a radio button or a checkbox.
+    /// 
     // https://html.spec.whatwg.org/#the-input-element:dom-input-value
     #[inline]
     pub fn raw_value( &self ) -> String {
@@ -31,6 +33,7 @@ impl InputElement {
     }
 
     /// Sets the value of the control.
+    /// 
     // https://html.spec.whatwg.org/#dom-input-value
     #[inline]
     pub fn set_raw_value( &self, value: &str ) {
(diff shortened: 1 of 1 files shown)


Wodann:
Build seems to have failed during build setup. Can you trigger a retry?

src/webapi/html_elements/input.rs
```diff
@@ -38,4 +38,44 @@ impl InputElement {
             @{self}.value = @{value};
         }
     }
+
+    /// The offset to the start of the selection.
+    /// This attribute only applies when the input is a text, search, url, telephone or password.
+    /// https://html.spec.whatwg.org/#textFieldSelection
```
koute: Wrong link, and shouldn't be a doc comment. (This also applies to other methods you've added.)

The correct one for `selection_start` is:

```
// https://html.spec.whatwg.org/#the-input-element:dom-textarea/input-selectionstart
```

(Always directly link to the piece of IDL.)

src/webapi/html_elements/input.rs
```diff
@@ -38,4 +38,44 @@ impl InputElement {
             @{self}.value = @{value};
         }
     }
+
+    /// The offset to the start of the selection.
```
koute: Add an extra newline (a line with only `///`) after summary. (This also applies to other methods you've added.)

src/webapi/html_elements/input.rs
```diff
@@ -38,4 +38,44 @@ impl InputElement {
...
+    /// This attribute only applies when the input is a text, search, url, telephone or password.
+    /// https://html.spec.whatwg.org/#textFieldSelection
+    #[inline]
+    pub fn selection_start( &self ) -> Option<u32> {
+        js! (
+            return @{self}.selectionStart
```
koute: Missing semicolon. (This also applies to other methods you've added.)

src/webapi/html_elements/input.rs
```diff
@@ -38,4 +38,44 @@ impl InputElement {
...
+        ).try_into().ok()
+    }
+
+    /// Sets the offset to the start of the selection.
+    /// This attribute only applies when the input is a text, search, url, telephone or password.or password.
+    /// https://html.spec.whatwg.org/#textFieldSelection
+    #[inline]
+    pub fn set_selection_start( &self, value: u32 ) {
```
koute: From what I can see from the specs this should return `Result< (), InvalidStateError >`. (This also applies to `set_selection_end`.)

Wodann:
Thank you for your feedback. I updated the PR with your comments. If you are happy with the changes, I can still squash the two commits into one to create a cleaner history, if you'd like.

koute:
Thanks!

PR #377, opened 2019-11-10
Expose data() on ImageData

tversteeg:


Pauan:
`set_data` shouldn't exist, since `data` is read-only.

It also might be a good idea to rename `get_data` to just `data`.

tversteeg:
How can the data be changed if it's read-only?

Pauan:
The `Uint8ClampedArray` which is returned by `data` is mutable, but the `data` property itself is read-only.

tversteeg:
I've rebased my comment and just exposed `data`, is this what you had in mind?

Pauan:
It looks good to me, just need to wait for @koute to review it.

Just a note that this isn't quite enough to support what you want, because `TypedArray` doesn't currently have any methods to mutate it, so those would have to be added as well.

mathieucaroff:
I've had a look at the [documentation of TypedArray](https://docs.rs/stdweb/0.4.20/stdweb/web/struct.TypedArray.html), and it's not clear to me how a TypedArray allows being written to.

Pauan:
@mathieucaroff See my comment right above yours.

PR #385, opened 2020-01-06
Replace sha1 dependency with sha-1

linkmauve:
This other crate is being maintained, and its API is simpler and more generic when using other [RustCrypto crates](https://github.com/RustCrypto/hashes) as well.


linkmauve:
With sha1 being the new upstream of sha-1, this change shouldn’t get merged any more, instead we should update sha1 to its 0.10 version.

linkmauve:
This project seems abandoned, closing.

PR #412, opened 2020-11-23
use MaybeUninit instead of uninitialized

arturoc:
Latest versions of rust will panic when trying to leave an uninitialized SerializedValue from mem::uninitialized which was breaking the js! macro

### DIFF
--- a/src/webcore/serialization.rs
+++ b/src/webcore/serialization.rs
@@ -412,8 +412,9 @@ macro_rules! untagged_boilerplate {
             #[inline]
             fn from( untagged: $untagged_type ) -> Self {
                 unsafe {
-                    let mut value: SerializedValue = mem::uninitialized();
-                    *(&mut value as *mut SerializedValue as *mut $untagged_type) = untagged;
+                    let mut value = mem::MaybeUninit::<SerializedValue>::uninit();
+                    *(value.as_mut_ptr() as *mut $untagged_type) = untagged;
+                    let mut value = value.assume_init();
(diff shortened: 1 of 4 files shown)


src/webcore/serialization.rs
```diff
@@ -412,8 +412,9 @@ macro_rules! untagged_boilerplate {
             #[inline]
             fn from( untagged: $untagged_type ) -> Self {
                 unsafe {
-                    let mut value: SerializedValue = mem::uninitialized();
-                    *(&mut value as *mut SerializedValue as *mut $untagged_type) = untagged;
+                    let mut value = mem::MaybeUninit::<SerializedValue>::uninit();
+                    *(value.as_mut_ptr() as *mut $untagged_type) = untagged;
+                    let mut value = value.assume_init();
```
RalfJung, line 417: `assume_init` should only be called after the `tag` has also been initialized.

RalfJung: Also this is still wrong for many cases -- the assignment of `untagged` might fail to completely initialize both `data_1` and `data_2`.

And finally, if `untagged` is a pointer type, then this would transmute a pointer to an integer, which is [not something one should do](https://github.com/rust-lang/rust/pull/95547). `SerializedValue` seems to be intended as a type that can hold *arbitrary* data, so `u64` and `u32` are the wrong types -- those types are specifically for integers, not for arbitrary data. Use e.g. `MaybeUninit<u64>` for arbitrary data of size 64 bits.

